PR #151, opened 2026-01-19
Work2

RasmusTho:


RasmusTho:
Enforced CI gates hard-fail when ok=false, and always print gates status + reasons before exiting. Added regression test covering ok=true/false. Tests: PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q -m 'not pg'.

PR #215, opened 2026-03-28
Fix Colima Ollama runtime endpoint startup

RasmusTho:
## Summary
- add `host.docker.internal:host-gateway` to the runtime containers so Colima-hosted Ollama is reachable from Docker
- increase the Ollama endpoint probe timeout during startup so the initial embedding call can complete before the probe fails

## Verification
- `DEBUG=0 VAULT_ROOT=/Users/rasmus/workspace/tmp/runtime-test-vault START_MODE=runtime scripts/start_full_system.sh`
- verified runtime endpoint repair from `http://ollama:11434` to `http://host.docker.internal:11434`
- verified startup completed with `runtime verified: true`

### DIFF
--- a/app/config/llm.py
+++ b/app/config/llm.py
@@ -38,9 +38,6 @@ def _log_provider_info(provider: str) -> None:
 
 def ensure_provider() -> str:
     global _ACTIVE_PROVIDER
-    if _ACTIVE_PROVIDER is not None:
-        return _ACTIVE_PROVIDER
-
     candidate = _normalize_provider(os.getenv("LLM_PROVIDER"))
     enforce = _env_flag("LLM_PROVIDER_ENFORCE", default=False)
 
@@ -51,6 +48,9 @@ def ensure_provider() -> str:
                 "export LLM_PROVIDER=ollama (plus any model/endpoint vars) or LLM_PROVIDER=mock for deterministic runs."
             )
         candidate = "mock"
+
+    if _ACTIVE_PROVIDER == candidate:
+        return _ACTIVE_PROVIDER
 
     _ACTIVE_PROVIDER = candidate
     _log_provider_info(candidate)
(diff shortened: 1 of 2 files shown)


RasmusTho:
Update on latest branch changes:

## Additional fix in this branch
- repair the legacy compat Postgres store shim so it works with the current dual-table schema
- stop the ingest PG test from dropping `decisions` directly now that `view_objects_ready_for_projection` depends on it

## Why
- `app/stores/postgres.py` was still writing `source_ref` into `objects`, but the live schema now keeps `source_ref` on `store_objects`
- `tests/ingest/test_vault_root_ingest_pg.py` failed with `DependentObjectsStillExist` because `DROP TABLE decisions` now conflicts with the dependent view

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q tests/int/test_pg_backend.py::test_pg_roundtrip`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q tests/ingest/test_vault_root_ingest_pg.py::test_ingest_vault_root_persists_objects_before_classification`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 STORE_BACKEND=memory LLM_PROVIDER=mock pytest -q tests/api/test_ask_alias.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 STORE_BACKEND=memory LLM_PROVIDER=mock pytest -q tests/test_hybrid_search.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 STORE_BACKEND=memory LLM_PROVIDER=mock pytest -q tests/cli/test_embed_probe.py`

Note: the mock embedding regression described in the initial report was not reproducible in this checkout; the targeted mock embedding/API tests were already passing locally.

PR #470, opened 2026-04-17
fix: normalize whitespace-only domain values to 'unscoped' at ingest boundary

### LINKED ISSUE #466: bug: whitespace-only domain values not normalized to 'unscoped' at ingest boundary
## Summary

`_ingest_single` in `app/ingest/vault_alpha.py` uses:

```python
domain = str(frontmatter.get("domain") or "unscoped")
```

A frontmatter value like `domain: "   "` (whitespace-only string) is truthy, so it is stored verbatim instead of being normalized to `"unscoped"`. Downstream, `_extract_domain` strips whitespace and returns `None`, causing the document to be excluded from every scoped query — including `ASK_DOMAIN_SCOPE=unscoped` — which violates the ingest boundary normalization contract established in Issue #437.

## Impact

Documents with whitespace-only domain values are silently un-queryable: they appear ingested but can never be retrieved in any scoped query, including the unscoped catch-all.

## Expected Behavior

Whitespace-only domain values should be treated as missing and normalized to `"unscoped"` at the ingest/write boundary, consistent with the intent of the #437 fix.

## Fix

```python
domain_raw = frontmatter.get("domain") or ""
domain = str(domain_raw).strip() or "unscoped"
```

## Source

Identified by Codex review on PRs #461 and #463. Code was merged to main via PR #460 (Issue #437). Out of scope for SEPSURF PRs; tracked here as a follow-up.

RasmusTho:
Fixes #466

## Summary
Whitespace-only `domain` frontmatter values (e.g. `"   "`) were truthy and bypassed the `or "unscoped"` fallback, leaving those documents silently un-queryable. The fix strips the value before falling back so any whitespace-only value is treated as absent.

## Validation
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3.12 -m pytest tests/test_domain_write_boundary.py -v -m "not pg"` — 8/8 passed, including new `test_ingest_with_whitespace_only_domain_normalizes_to_unscoped`.

---

### DIFF
--- a/app/ingest/vault_alpha.py
+++ b/app/ingest/vault_alpha.py
@@ -402,7 +402,7 @@ def _ingest_single(path: Path, *, vault_root: Path, trace_id: str, raw_text: str
     title = _frontmatter_title(frontmatter) or _derive_title(body, path)
     review_state = str(frontmatter.get("review_state") or "provisional")
     maturity = str(frontmatter.get("maturity") or "note")
-    domain = str(frontmatter.get("domain") or "unscoped")
+    domain = str(frontmatter.get("domain") or "").strip() or "unscoped"
     stripped_body = strip_ai_panels(body)
     stripped_text = stripped_body.strip()
     ingest_fingerprint = _compute_ingest_fingerprint(stripped_text, path)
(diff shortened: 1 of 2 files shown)


RasmusTho:
post-merge owner-doc check: no owner-doc change implied. (STATUS.md already documents this fix as shipped at "whitespace-only domain normalization shipped via PR #470".)

PR #748, opened 2026-05-03
docs: fix dispatcher SSH wrapper example

RasmusTho:
- [x] Docs authoring lane

## Summary
- update the dispatcher wrapper snippet to set `PYTHONPATH=.` so `app.dispatcher` resolves on Demerzel
- use safe argument forwarding with remote positional args (`"$@"`) and avoid the broken literal `--` prefix before dispatcher subcommands

## Validation
- verified `dispatcher status --json` returns `{"ok": true, "db_exists": true, ...}` after applying the wrapper locally


RasmusTho:
post-merge owner-doc check: no owner-doc change implied.

PR #791, opened 2026-05-06
fix: redact password query params in settings_explain DSN

RasmusTho:
Fixes #772

## Summary
- redact password-like DSN query parameters (for example password, pass, pwd) in settings_explain output
- preserve existing masking for userinfo credentials and DSN structure
- add regression test covering query-param credential leakage case from Codex review

## Validation
- ./.venv/bin/pytest -q tests/cli/test_settings_explain_cli.py

## Dispatcher
- Claimed via dispatcher task github-issue-772 before implementation.

### DIFF
--- a/app/cli/settings_explain.py
+++ b/app/cli/settings_explain.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import os
 from typing import Any, Sequence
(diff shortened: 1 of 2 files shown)


RasmusTho:
@codex review

PR #3508, opened 2026-07-11
Add two-stream five-dimension segmentation core emitting proposed episodes (ERE-04)

RasmusTho:
Fixes #3179

## Summary

ERE-04: the two-stream five-dimension segmentation core. Consumes `heimdal.observations` (existing per-consumer cursor via `app/heimdal/publish.py`, consumer id `mimer.episode_resolution_engine`) and `vault.activity` (`ingest.vault.changed` / `ingest.object.created` / `ingest.object.deleted` outbox topics via a new topic-scoped, `delivered_at`-independent durable cursor) strictly through the ERE-01 stream registry, detects boundaries with five-dimension shift detection (named, single-sourced provisional constants in `app/episodes/segmenter.py`; over-segmentation preferred), and emits `segmentation: proposed` Episode notes through the ERE-02 guarded store seam with full `derived_from` and bitemporal bounds from `observed_at` (never emission time). Heimdal's per-session `episode_id` is a boundary hint checked before every shift dimension — one session never spans two proposals (ADR-0054 §3). Deterministic tick: `python -m app.cli episodes tick --json`. Fold-by-key idempotent under at-least-once redelivery (cursor advance strictly after note emission + deterministic per-segment episode ids). Segments partitioned per-scope by default (ERE-08 owns the full gate). Heimdal untouched.

New substrate: `episode_engine_state` table (migration `a1b2c3d4e5f6`, forward-only) holding the vault-activity cursor + open-segment state — pure rebuildable tick-runtime bookkeeping, never authoritative.

## Claim receipt

```
pickup-claim-complete issue=3179 branch=codex/issue-3179-segmentation-core worktree=/Users/rasmus/worktrees/issue-3179 coordination_mode=dispatcher-backed fallback_reason=none task_id=github-RasmusTho--agentic-pkm-mvp-issue-3179 lease_id=lease-f7f8bfe6 holder=claude-sonnet-ere-3179 evidence=verified-dispatcher-lease
```

## AC -> Verify

| AC | Verify | Result |
| --- | --- | --- |
| AC1 fixture segments into two episodes, correct bounds + derived_from | `tests/episodes/test_segmentation_core.py::test_two_stream_fixture_segments_into_expected_episodes` | pass |
| AC2 idempotent under redelivery | `::test_segmentation_idempotent_under_redelivery` (+ `::test_emission_idempotent_under_redelivery`) | pass |
| AC3 session hint respected | `::test_heimdal_session_hint_respected` | pass |
| AC4 registry-only consumption at entrypoint | `tests/episodes/test_stream_registry.py::test_engine_consumes_only_registered_streams` (real segmenter body) | pass |
| AC5 schema-valid proposal-class notes | `::test_proposals_are_schema_valid_proposal_class` | pass |
| AC6 named single-sourced provisional constants | doc writeback `docs/EPISODE_RESOLUTION_ENGINE/README.md :: Provisional thresholds (RQ-E1)` | done |
| AC7 per-scope partitioning default | `::test_segments_keyed_per_scope_by_default` | pass |

## Validation

- `ruff check app tests` — All checks passed.
- `mypy app` — 5 errors, ALL pre-existing on clean origin/main @ 115922f2 (`app/builderops/model_inquiry_adapters.py` types-requests, `app/builderops/ckm/store.py:429` from #3505, `app/heimdal/screen_capture.py:51,94` from #3485); 0 NEW errors (my tree checks 694 files vs main's 690, identical error set).
- `pytest -q tests/episodes/` — 24 passed, 2 skipped (pg-marked skip without Postgres, by design).
- `pytest -q -m "not pg"` — full suite on the rebased tree: see checks; pre-rebase run was 7747 passed / 7 failed, of which 1 was mine (fixed: db-import guard allowlist entries for the two new bounded-DB modules), 3 fixed by upstream (#3503 + docs-index), and 3 reproduce identically on clean origin/main @ 115922f2 (pre-existing, not this PR): `tests/properties/test_guard_at_seam.py::test_every_write_note_relative_seam_has_port_coverage` (offender `app/chat/session_log.py:68`), `tests/api/test_capture_inbox_api.py::test_capture_appends_to_inbox_through_governed_pipeline`, `tests/heimdal/test_published_control_surface_schema.py::test_published_schema_matches_settings_note_registry`.
- `RUN_INTEGRATED_RUNTIME_UAT=1 pytest -q -m "not pg" tests/uat` — 10 passed.
- Alembic: single head `a1b2c3d4e5f6` (revises `e0f2a9c4b7d1`), forward-only with reversibility classification.

## Doc writebacks

- `docs/EPISODE_RESOLUTION_ENGINE/README.md :: Provisional thresholds (RQ-E1)` — delivered-constants table (AC6).
- `docs/EVENTS.md :: Outbox consumer contract` — new subsection documenting the secondary topic-scoped per-consumer cursor pattern introduced for `vault.activity`.

## BuilderOps Routing

- Reason: bounded slice delivered as specified; the one divergence (upstream test-suite drift classified against clean origin/main, resolved by rebase) is described in Validation and names no upstream artifact needing a LearningSignal.

---


app/episodes/vault_activity_stream.py
```diff
@@ -0,0 +1,167 @@
...
+                    payload = _json.loads(payload)
+                rows.append(
+                    VaultActivityRow(
+                        id=str(rid), topic=str(topic), created_at=created, payload=dict(payload or {})
+                    )
+                )
+    return rows
+
+
+def _note_rel_path_from_payload(payload: Mapping[str, Any]) -> str | None:
```
RasmusTho: **Review gate (coordinator, round 1) — CONFIRMED:** `_note_rel_path_from_payload` falls back to payload['path'], which for ingest.object.created/deleted is an ABSOLUTE filesystem path (vault_sync.py resolves it) — `Path(vault_root) / rel_path` then discards vault_root entirely, so a relocated vault or explicit --vault-root override silently reads the ingest-time path (empty dims or wrong vault in shared-path deployments). Requested: detect absolute fallback paths and re-relativize against the ingest-time root or reject them (fail-loud), + a vault_activity_stream test covering created/deleted payload shapes with a differing vault_root.

app/episodes/engine_state.py
```diff
@@ -0,0 +1,100 @@
...
+
+import json
+from typing import Any
+
+from app.db.db import conn_rw
+
+STATE_TABLE = "episode_engine_state"
+
+
+def get_state(key: str) -> dict[str, Any] | None:
```
RasmusTho, line 74: **Review gate (coordinator, round 1) — CONFIRMED (invariant→producers):** episode_engine_state is queried with no fail-loud schema preflight, unlike every sibling state module (app/heimdal/cursor_store.py `_assert_pg_schema` → SchemaMissingError with migration hint). A pre-migration environment gets a raw UndefinedTable traceback from deep inside get_state. Requested: EngineStateSchemaMissingError preflight via to_regclass, mirroring cursor_store, + test.

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+    dt = value if value.tzinfo else value.replace(tzinfo=timezone.utc)
+    return dt.astimezone(timezone.utc).isoformat()
+
+
+# ---------------------------------------------------------------------------
+# Five-dimension shift detection (ADR-0051 commitment 2)
+# ---------------------------------------------------------------------------
+
+
+def _goal_shift(open_segment: OpenSegment, signal: SegmentationSignal) -> bool:
```
RasmusTho: **Review gate (coordinator, round 1) — CONFIRMED (altitude):** `_goal_shift` and `_protagonist_shift` are line-for-line identical apart from the attribute read. Requested: one `_disjoint_set_shift(enabled, open_set, signal_set)` helper called twice (the next dimension in this family reuses it instead of a third copy).

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+
+#: Causal-break dimension (v1): an explicit Heimdal `supersedes` marker on
+#: the observation payload is treated as a discontinuity.
+CAUSAL_BREAK_DETECTION_ENABLED: Final[bool] = True
+
+#: Place-shift dimension: absent in v1 by design (spec) -- unfed until a
+#: calendar/location stream lands (ERE-09/ERE-10). Never contributes a
+#: shift; kept named here so its absence is a documented decision, not a
+#: silent omission.
+PLACE_SHIFT_DETECTION_ENABLED: Final[bool] = False
```
RasmusTho, line 191: **Review gate (coordinator, round 1) — PLAUSIBLE (low severity, author's call):** the four *_DETECTION_ENABLED Final[bool] constants have no runtime toggle path and PLACE_SHIFT_DETECTION_ENABLED has zero read sites. Acceptable as documented intent; if kept, note in the constants table that they are compile-time documentation, not runtime flags.

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+    dims = resolve_activity_dimensions(row, vault_root=vault_root)
+    scope_raw = dims.get("scope")
+    scope = scope_raw.strip() if isinstance(scope_raw, str) and scope_raw.strip() else _DEFAULT_SCOPE
+    spheres = dims.get("sphere_memberships") or []
+    # Provisional goal-dimension proxy for vault activity (no explicit
+    # project/area field exists on `extract_context_dimensions_for_note`
+    # yet): sphere membership is the closest available goal/context binding.
+    goal = tuple(sorted({str(s) for s in spheres if str(s).strip()}))
+
+    created_at = row.created_at if row.created_at.tzinfo else row.created_at.replace(tzinfo=timezone.utc)
```
RasmusTho: **Review gate (coordinator, round 1) — CONFIRMED:** vault.activity signals take observed_at from the outbox row's created_at (emission/insert time) while the real change time sits unused in payload['mtime'] — ingestion lag shifts episode bounds, violating the spec's 'bounds from observed_at, never emission time' rule and docs/EVENTS.md's own envelope-vs-payload convention. Requested: read the payload's observation time (mtime) with a fail-loud/skip posture when absent.

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+                continue
+            if detect_shift(current, signal, now=now):
+                closed.append(_close(current))
+                current = _new_open_segment(signal)
+            else:
+                current = _extend_open_segment(current, signal)
+        if current is not None:
+            working[scope] = current
+
+    for scope, current in list(working.items()):
```
RasmusTho, line 457: **Review gate (coordinator, round 1) — CONFIRMED:** the staleness closure loop closes open segments on WALL-CLOCK now - last_signal_at with no heimdal_session_id continuity check — a delivery/tick lag >TIME_GAP splits one Heimdal session across two proposals, contradicting AC3's 'one session never spans two proposals' and session-hint precedence. Requested: closure decision based on observed-time gaps consistent with the spec, with the session hint honored at the closure seam too (consult the spec's quiescence semantics; closure of a segment whose session may still be live must not fire on delivery lag).

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+        # Idempotent fold (AC2): this exact signal already contributed to
+        # this open segment (at-least-once redelivery before the cursor
+        # advanced) -- a no-op, never a duplicate contribution.
+        return open_segment
+    derived_from = open_segment.derived_from
+    if signal.provenance_ref not in derived_from:
+        derived_from = derived_from + (signal.provenance_ref,)
+    return replace(
+        open_segment,
+        last_signal_at=max(open_segment.last_signal_at, signal.observed_at),
```
RasmusTho: **Review gate (coordinator, round 1) — CONFIRMED:** _extend_open_segment never lowers start (no start=min(...) term) — a late-delivered signal with earlier observed_at (independent second-stream cursor) folds in but leaves time.start postdating a signal in the episode's own derived_from, breaking AC1's bitemporal bounds. Requested: start=min(open.start, signal.observed_at) + out-of-order regression test.

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+    return row.id
+
+
+def _signal_from_heimdal_row(row: ObservationRow) -> SegmentationSignal:
+    payload = row.envelope.get("payload")
+    payload = payload if isinstance(payload, Mapping) else {}
+    observation_id = _observation_id_of(row, payload)
+
+    observed_at_raw = payload.get("observed_at_start")
+    observed_at = _parse_dt(observed_at_raw) if observed_at_raw else row.created_at
```
RasmusTho: **Review gate (coordinator, round 1) — PLAUSIBLE:** _signal_from_heimdal_row falls back to row.created_at when observed_at_start is missing — leaks emission time defensively. The Heimdal publish contract makes it unreachable in practice; prefer fail-loud skip (count + log) over silent emission-time substitution.

app/alembic/versions/a1b2c3d4e5f6_ere04_segmentation_engine_state.py
```diff
@@ -0,0 +1,71 @@
...
+  `delivered_at` would race the worker's own dispatch. `heimdal.observations`
+  reuses the EXISTING `heimdal_observation_cursor` table via
+  `app.heimdal.publish` and needs no new row family here.
+- `open_segment:<scope>` -- the accumulated situation-model state of one
+  scope's currently-open (not yet proposed) segment, so a restart re-derives
+  in-flight segmentation instead of losing it.
+
+Never authoritative: this table is pure tick-runtime bookkeeping, fully
+replayable from the underlying streams (Heimdal observation log; DB outbox) --
+losing it only means the engine re-derives open segments from event zero,
```
RasmusTho: **Review gate (coordinator, round 1) — PLAUSIBLE (design honesty):** the migration claims 'same posture as heimdal_observation_cursor' but co-locates the vault.activity cursor with open-segment state; wiping episode_engine_state alone resets ONE stream's cursor to event zero while the heimdal cursor stays advanced — a skewed replay, not the claimed symmetric rebuild. Requested: at minimum correct the rationale comment and document the recovery posture (both cursors wiped together, or replay is idempotent-by-design because proposals fold by deterministic key — if the latter is true, say so and add the assertion/test).

app/episodes/vault_activity_stream.py
RasmusTho: Fixed in b8846c28: _resolve_note_path now containment-checks every candidate (absolute ingest-time or relative) against the current vault_root via is_relative_to and rejects anything not underneath it (also closes .. traversal). Regression: test_vault_activity_absolute_paths_contained_to_vault_root.

app/episodes/engine_state.py
RasmusTho: Fixed in b8846c28: EngineStateSchemaMissingError + _assert_schema (to_regclass preflight) on every public function, mirroring cursor_store._assert_pg_schema, with a migration hint. Regression: test_engine_state_schema_preflight_fails_loud_with_migration_hint.

app/episodes/segmenter.py
RasmusTho: Fixed in b8846c28: _goal_shift/_protagonist_shift collapsed into one _disjoint_set_shift(enabled, open_set, signal_set) helper called twice.

app/episodes/segmenter.py
RasmusTho: Kept as documented compile-time constants; added a comment stating they are compile-time documentation of active dimensions, not runtime flags (no env/settings toggle path by design).

app/episodes/segmenter.py
RasmusTho: Fixed in b8846c28: vault.activity observed_at now comes from payload mtime (_parse_observation_time); the outbox row created_at is never substituted; rows without an observation time are skipped fail-loud (skipped_no_observation_time count + log). Regression: test_vault_activity_signal_uses_payload_mtime_not_row_created_at.

app/episodes/segmenter.py
RasmusTho: Fixed in b8846c28: quiescence closure now measures against durable observed-time stream watermarks (_closure_frontier); a session-bound segment gates on the Heimdal frontier specifically, so delivery/tick lag can never split one session across two proposals. Wall-clock removed from the fold entirely (now param dropped). Regression: test_session_never_split_by_delivery_lag.

app/episodes/segmenter.py
RasmusTho: Fixed in b8846c28: _extend_open_segment now sets start=min(open.start, signal.observed_at); observed_at_end also extends last_signal_at over the full observation window. Regression: test_out_of_order_signal_widens_segment_start + test_observed_at_end_extends_segment_bounds.

app/episodes/segmenter.py
RasmusTho: Fixed in b8846c28: _signal_from_heimdal_row returns None (skip fail-loud, count + log) when observed_at_start is absent instead of falling back to row.created_at. Regression: test_heimdal_signal_skips_row_without_observed_at.

app/alembic/versions/a1b2c3d4e5f6_ere04_segmentation_engine_state.py
RasmusTho: Fixed in b8846c28: migration + engine_state docstrings corrected to state the asymmetric recovery posture honestly (single-stream reset is a skewed replay, not supported; operator recovery resets both cursor families together — full both-stream replay is deterministic and emission-deduped, asserted by test_emission_idempotent_under_redelivery). DB_SCHEMA.md documents the table.

app/episodes/segmenter.py
```diff
@@ -71,7 +120,668 @@ def run_segmenter_stub(
...
+            if current is None:
+                current = _new_open_segment(signal)
+                continue
+            if detect_shift(current, signal):
+                closed.append(_close(current))
+                current = _new_open_segment(signal)
+            else:
+                current = _extend_open_segment(current, signal)
+        if current is not None:
+            working[scope] = current
```
RasmusTho, line 455: **Review gate round 2 — CONFIRMED (regression from the round-1 fix):** the quiescence frontier is per-STREAM (tick-wide max across ALL scopes), so a sibling scope's later Heimdal signal in the same batch closes THIS scope's just-created segment. Worse: it reintroduces the AC3 session-split — a session-bound segment closed on another session's frontier progress, whose own continuation arrives next tick, becomes two proposals. Fix must make closure per-scope and defer session-bound segments to their own session/ERE-06 rather than a global stream frontier. Grounding the exact rule in TWO_STREAM_SEGMENTATION_CORE.md before fixing.

app/episodes/segmenter.py
```diff
@@ -71,7 +120,668 @@ def run_segmenter_stub(
...
+
+
+def _parse_observation_time(value: Any) -> datetime | None:
+    """Parse a payload observation-time value: epoch seconds (the watcher's
+    `mtime` float) or an ISO-8601 string. ``None`` when absent/unparseable --
+    the caller skips the signal fail-loud (count + log), never substitutes
+    emission time."""
+    if isinstance(value, bool):
+        return None
+    if isinstance(value, (int, float)):
```
RasmusTho, line 505: **Review gate round 2 — CONFIRMED (crash):** _parse_observation_time's numeric branch calls datetime.fromtimestamp(float(value)) with no try/except (unlike the string branch), so mtime=NaN/inf/out-of-range raises straight out and crashes the whole tick — every other row in the batch is lost, contradicting the 'skip fail-loud (count+log)' contract. Wrap in try/except (ValueError, OverflowError, OSError) → return None. Also treat epoch<=0 as absent (mtime=0 sentinel must not date a signal to 1970 and drag start back via the min()-widening).

app/episodes/segmenter.py
```diff
@@ -71,7 +120,668 @@ def run_segmenter_stub(
...
+
+def _extend_open_segment(open_segment: OpenSegment, signal: SegmentationSignal) -> OpenSegment:
+    if signal.signal_id in open_segment.signal_ids:
+        # Idempotent fold (AC2): this exact signal already contributed to
+        # this open segment (at-least-once redelivery before the cursor
+        # advanced) -- a no-op, never a duplicate contribution.
+        return open_segment
+    derived_from = open_segment.derived_from
+    if signal.provenance_ref not in derived_from:
+        derived_from = derived_from + (signal.provenance_ref,)
```
RasmusTho, line 373: **Review gate round 2 — PLAUSIBLE:** the crash-residual 'never a duplicate' claim is now false because fix #3 makes start mutable: if the stale (undeleted) open-segment resurrects on a later out-of-order signal with observed_at < original start, the recomputed start changes, _deterministic_episode_id mints a NEW id, the file-exists check misses the original note, and a second overlapping note is written. Either key the deterministic id on something stable (not start) or make re-emission genuinely idempotent against the already-written note.

app/episodes/segmenter.py
RasmusTho: Fixed in bc8192e6. The closure frontier is now keyed PER-SCOPE, not per-stream. `fold_signals_into_segments`'s `frontiers` param is `scope -> max observed instant for that scope`, and a segment in scope X closes on quiescence iff scope X's OWN observed head has moved >45min past its last signal. A sibling scope's later Heimdal signal in the same batch can no longer close another scope's just-created segment (the confirmed AC3 violation). `_closure_frontier` and its session-bound-vs-unbound branching are removed — per-scope keying subsumes it: a sibling session in a different scope can never close you; a different session in the SAME scope closes via `detect_shift` (unchanged); the same session continuing always extends. The durable per-stream watermark machinery (`_STREAM_WATERMARK_KEY_PREFIX`, `_load_stream_watermarks`, the persistence loop) is gone; the frontier is now built per-scope from each tick's own consumed signals. A scope with no signal this tick stays open, deferred to ERE-06. New regression test: `test_sibling_scope_progress_never_closes_segment`.

app/episodes/segmenter.py
RasmusTho: Fixed in bc8192e6. The numeric branch of `_parse_observation_time` is now guarded via a shared `_epoch_seconds_to_dt` helper that catches ValueError/OverflowError/OSError -> None, so a NaN/inf/out-of-range mtime skips fail-loud (counted + logged) instead of crashing the whole tick. Numeric strings route through the same helper, so "nan"/"inf" strings are also handled. An epoch <= 0 (e.g. an mtime=0 sentinel) is treated as ABSENT (return None) so it never dates a signal to 1970 and drags `start` backward through the min()-widening. The docstring now documents that a purely-numeric string is epoch seconds by design. New tests: test_parse_observation_time_guards_malformed_numeric and test_malformed_mtime_skips_row_but_batch_still_folds.

app/episodes/segmenter.py
RasmusTho: Fixed in bc8192e6. `_deterministic_episode_id` no longer depends on `start`. It is now keyed on the segment's STABLE identity — `scope | derived_from[0]` (the first-folded signal's provenance_ref), which append-only extension and out-of-order start-widening never mutate. So a resurrected stale segment (the documented crash-residual) whose `start` widens on replay still mints the SAME id, the existence-check finds the original note, and no duplicate proposal is written. derived_from is non-empty in practice (a segment is always created from a signal with a non-empty provenance_ref); the empty fallback is defensive only, and two distinct same-scope segments always differ in derived_from[0] so they cannot collide. Docstring + the crash-residual comment in run_segmentation_tick updated to state the id is start-independent. New test: test_episode_id_stable_across_start_widening (asserts start-widened and end-extended variants share an id, and distinct first refs differ).

app/episodes/segmenter.py
```diff
@@ -71,7 +131,677 @@ def run_segmenter_stub(
...
+            else:
+                current = _extend_open_segment(current, signal)
+        if current is not None:
+            working[scope] = current
+
+    for scope, current in list(working.items()):
+        # Quiescence is judged ONLY against this segment's own scope frontier
+        # (per-scope isolation): a sibling scope moving on says nothing about
+        # whether this scope's own later signals are still in flight.
+        frontier = frontier_map.get(scope)
```
RasmusTho, line 461: **Review gate round 3 — CONFIRMED (frontier must use observed_at, not observed_until):** the per-scope frontier is built from max(observed_until) this tick, so a long observation A (09:00-10:30) inflates scope 'work's frontier to 10:30 and closes a sibling session-bound segment B (created later same tick via a goal shift, last_signal_at=09:10) — 10:30-09:10=80min>gap — reopening the AC3 session-split within-scope. Root cause: the coordinator's round-2 design guidance said observed_until; that was wrong. The closure frontier is the scope's READ POSITION (when the next signal STARTED), so build it from max(observed_at). Keep observed_until only for last_signal_at/bounds. Also: no test drives run_segmentation_tick's real frontier construction (all round-2 tests hand-build the frontiers dict) — add an integration-level test that exercises the production path with an observed_at_end-bearing signal + a nested sibling session.

docs/DB_SCHEMA.md
```diff
@@ -309,6 +309,34 @@ Interpretation:
...
+`docs/EVENTS.md :: Secondary per-consumer cursor readers` for the consumer contract.
+
+- `episode_engine_state` — generic key/value state for the segmentation tick
+  (`app/episodes/segmenter.py::run_segmentation_tick`).
+  - `key` (`text`, PK) — namespaced row families:
+    - `cursor:vault.activity:<consumer_id>` — the engine's own durable read position over the
+      `outbox` table's vault-activity topics (independent of `outbox.delivered_at`, which the
+      worker dispatcher owns);
+    - `open_segment:<scope>` — one scope's currently-open (not yet proposed) segment state;
+    - `stream_watermark:<stream_id>` — max observed instant consumed per stream (observed-time
```
RasmusTho: **Review gate round 3 — CONFIRMED (doc/code drift):** this still lists `stream_watermark:<stream_id>` as a live row family of episode_engine_state, but the round-2 redesign deleted the watermark persistence entirely (per-tick per-scope frontier now). engine_state.py and the migration docstring were updated; this doc was not. Remove/rewrite the stream_watermark bullet so DB_SCHEMA.md matches the shipped schema.

app/episodes/segmenter.py
RasmusTho: Fixed in 63874745. The per-scope quiescence-closure frontier in `run_segmentation_tick` is now built from `signal.observed_at` (the read position — where the next signal STARTED), not `signal.observed_until` (a signal's content-span end). This closes the confirmed within-scope AC3 split: previously a signal A with a long `observed_at_end` (e.g. 09:00–10:30) pushed the scope frontier to 10:30, so when a protagonist/goal shift closed A and opened a nested session-bound segment B (last_signal_at=09:10), the quiescence loop closed B in the same tick (10:30−09:10=80min>45) and B's session reopened next tick as a second proposal. Now the frontier is max(observed_at)=max(09:00,09:10)=09:10, so B (last 09:10) is not closed and stays open. `observed_until` remains authoritative for `last_signal_at` and emitted bounds — only the closure read-position keys on `observed_at`. New enforcement test `test_tick_long_observed_window_never_closes_nested_session_segment` drives the REAL run_segmentation_tick frontier construction (stubbed stream readers + engine_state, real fold + emission) and I verified it fails with two proposals if the frontier keys on observed_until. Three cases confirmed: (a) round-2 sibling-scope {work:09:00, other:09:50} still doesn't close work; (b) this nested case B not closed; (c) AC1 pure-fold still closes the trailing segment against its explicit frontier dict.

docs/DB_SCHEMA.md
RasmusTho: Fixed in 63874745. Removed the stale `stream_watermark:<stream_id>` row-family bullet from docs/DB_SCHEMA.md under `episode_engine_state`. The watermark persistence was deleted in round 2; the quiescence-closure frontier is now a per-scope read position computed fresh from each tick's own consumed signals, not carried durably. The doc now matches the shipped schema (engine_state.py and the migration docstring were already correct).

app/episodes/segmenter.py
```diff
@@ -71,7 +110,508 @@ def run_segmenter_stub(
...
+        engine_state.delete_state(f"{_OPEN_SEGMENT_KEY_PREFIX}{scope}")
+
+    # Advance cursors LAST (INV-ERE-F): a crash before this point means the
+    # next tick re-reads the same batch, deduped by fold-by-key
+    # (signal_ids already recorded on the open segment) plus the
+    # deterministic episode id (an already-written note is never re-emitted).
+    if heimdal_rows:
+        advance_cursor_for_consumer(consumer_id, heimdal_rows)
+    if vault_rows:
+        advance_vault_activity_cursor(consumer_id, vault_rows)
```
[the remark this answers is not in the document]
RasmusTho: Captured by the daily review-comment closure audit as residual terminal-PR feedback: #3521. Leaving this thread unresolved until #3521 either lands a fix, proves this is already addressed, or records a supersession/discard receipt.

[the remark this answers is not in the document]
RasmusTho: Audit #3521 classification: ADDRESSED. Current crash-safe ordering retains closed-segment `signal_ids` through both cursor advances; a one-stream replay folds to no-op and the start-independent episode id prevents a duplicate partial proposal. Evidence: `bc8192e6`, current `app/episodes/segmenter.py:1191-1197,1372-1378`.

PR #3517, opened 2026-07-12
feat(settings): wire vault settings ingestion at startup + reload/degrade (SETTINGS-01, #3159)

RasmusTho:
## Change Lane
- [x] Implementation lane
- [ ] Docs authoring lane
- [ ] Governance lane

## Linked Issue
Closes #3159

## Summary
Vault-authored `@Settings` sources now load at API, worker, and production registry-watcher startup, including active dev/test vault bindings and `@Settings/agents/*.md`.

The production registry path detects source changes even when its note scope is narrowed. It recompiles against its bound vault and publishes an atomic generation signal on the existing channel-scoped shared runtime volume. API and worker observe that signal on their next settings read and rebuild their own local projection from vault markdown; invalid edits retain the last-valid bundle and propagate `settings.state=degraded_last_valid` to health.

## SBS Impact
Product/Runtime System change (WSP hot path; DRI derived projection; OEF health signal).
- Markdown remains the vault-authoritative source; `runtime/settings` and the reload signal are derived runtime state.
- No new authority or approval path; no-vault boot remains typed defaults with no CWD fallback.
- The signal uses the already shared `/app/tmp` compose runtime surface, so channel isolation remains provided by the existing per-project volume.

## Owner-Doc Writeback
- [x] `docs/OBSERVABILITY.md` documents cross-process invalidation and the health-state semantics.

## Review fixes
- Registry watcher source-delta ingestion is wired into `python -m app.cli watcher run`'s real scan path.
- API/worker receive cross-process invalidation and recompile their process-local projection; no module-local event-bus assumption remains.
- Source discovery resolves using `active_environment()` and supports `VAULT_ROOT_DEV` / `VAULT_ROOT_TEST`.
- Source discovery includes `@Settings/agents/*.md`.

## Validation
- `pytest -q tests/settings/test_ingestion_startup.py tests/watcher/test_registry_tick_sleep.py` — 14 passed.
- `ruff check app tests` — clean.
- `mypy app` — 4 pre-existing unrelated errors in `app/builderops/ckm/{models,store}.py` and `app/heimdal/screen_capture.py`; none in changed modules.
- `pytest -q -m "not pg"` — running locally in the shared test queue when this handoff was published; CI will provide the authoritative fresh full-suite result.

## BuilderOps Routing
- Records/projections/receipts: none.
- Reason: bounded runtime wiring and documentation change; no BuilderOps authority material was produced.

## Remaining pre-merge UAT
Run on the mac-mini test channel:
- `RUN_INTEGRATED_RUNTIME_UAT=1 pytest -q tests/integration -k settings`
- Dev-channel receipt: edit a setting in Obsidian, verify a live API/worker effect and `/api/health` reports `settings.state=ok, source=vault`.


app/settings/reload_signal.py
```diff
@@ -0,0 +1,80 @@
...
+import json
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from time import time_ns
+from typing import Any
+
+
+_SIGNAL_ENV = "SETTINGS_RELOAD_SIGNAL_PATH"
+_DEFAULT_SIGNAL_PATH = Path("/app/tmp/settings-reload.json")
```
RasmusTho: **[P1] Put the test-channel signal on its shared runtime volume**

The default always targets `/app/tmp`, but `docker-compose.test.yml` mounts the channel-scoped `runtime-tmp` volume at `/app/tmp-test` for API, worker, and watcher and does not set `SETTINGS_RELOAD_SIGNAL_PATH`. Thus each test container reads/writes its own unshared `/app/tmp/settings-reload.json`, so the required watcher→API/worker invalidation cannot work in the mac-mini test channel. Resolve this path through the same environment-scoped runtime-artifact policy (or set the override consistently for every service) and cover the actual test compose path.

app/watcher/registry.py
```diff
@@ -986,6 +996,20 @@ def _collect_changed_entries(
...
+            if source_delta.reloaded:
+                summary["settings_source_reloads_in_tick"] = (
+                    int(summary.get("settings_source_reloads_in_tick", 0)) + 1
+                )
+            if source_delta.errors:
+                state.errors += len(source_delta.errors)
+                summary["settings_source_errors_in_tick"] = int(
+                    summary.get("settings_source_errors_in_tick", 0)
+                ) + len(source_delta.errors)
         changed_entries.append(ChangedEntry(rel_path=rel, mtime=mtime, digest=digest))
```
RasmusTho, line 1023: **[P1] Stop settings control files before ordinary watcher emission**

After handling an `@Settings` source, this still appends it to `changed_entries`; every registry spec then processes it as normal content. With the shipped `configs/watchers.yaml`, an edit is recompiled once per spec and the ingest spec emits `ingest.vault.changed` for `@Settings/global.md` (reproduced with the production tick), leaking runtime-control markdown into the knowledge ingestion channel. Update per-spec file state after the settings reload, but do not enqueue/emit the settings source through panel/ingest business flows; ideally reload once per registry tick rather than once per spec.

app/settings/ingestion.py
```diff
@@ -0,0 +1,218 @@
...
+        return None
+    if vault_root is None:
+        return None
+    return vault_root / SETTINGS_SOURCE_DIR_NAME
+
+
+def get_settings_ingestion_state() -> SettingsIngestionState:
+    # A watcher reload runs in a different container.  Mirror its published
+    # result into this process so API health reports the same fail-loud state.
+    signal = read_reload_signal()
```
RasmusTho, line 96: **[P1] Do not claim last-valid state in a process that has none**

This blindly replaces the process-local health state with the watcher's signal. If the watcher had a last-valid bundle, then publishes `degraded_last_valid`, a newly restarted API first correctly records `invalid_sources/defaults` for the same broken source, but its next health read is overwritten to `degraded_last_valid/vault` even though that API never loaded a valid bundle and is serving defaults. I reproduced exactly that transition. Treat the shared record as invalidation/error evidence and derive degradation from the receiving process's local prior state; a fresh process must remain `invalid_sources` until it successfully loads vault settings.

app/settings/compiler.py
```diff
@@ -307,19 +307,25 @@ def _update_reference(path: Path, title: str, model: Any, auto_heal: bool, *, va
...
 
 
-def compile_all(*, auto_heal: bool | None = None) -> SettingsBundle:
+def compile_all(
+    *, auto_heal: bool | None = None, vault_dir: Path | None = None
+) -> SettingsBundle:
+    # ``vault_dir`` overrides the settings-source location so ingestion can compile
+    # the *selected* vault's ``@Settings`` (SETTINGS-01) rather than the packaged
+    # ``vault/@Settings`` convention; callers that omit it keep the existing default.
+    settings_dir = vault_dir if vault_dir is not None else VAULT
```
RasmusTho, line 387: **[P1] Carry the selected vault into the specialized settings loaders**

`vault_dir` now scopes the generic and agent markdown reads, but later `load_watcher_settings()` is still called without `vault_root`, so it independently resolves only the unscoped vault binding. Under `PKM_ENVIRONMENT=dev` with only `VAULT_ROOT_DEV`, I compiled a selected vault whose `watchers.md` sets `auto_exec_default: false`; the emitted `watchers.yaml` contained `true` and an empty source path. Pass `vault_root` (and the active environment where needed) to specialized loaders so all compiled settings come from the same selected channel/vault—otherwise a safety-relevant watcher false policy is silently ignored.

tests/api/test_capture_inbox_api.py
```diff
@@ -79,6 +79,10 @@ def test_capture_appends_to_inbox_through_governed_pipeline(
 
     def _spy(action: str) -> None:
         guard_actions.append(action)
```
RasmusTho, line 81: **[P3] Remove the temporary #3518 traceback instrumentation**

This branch still commits explicitly temporary `DIAG-3518` stack/identity prints even though #3518 is closed as superseded and the assertion is already aligned with main. Leaving this in the permanent full-suite test produces noisy tracebacks on every normal run and keeps unrelated diagnostic scope in #3159; remove both temporary print blocks before closure.

RasmusTho (commented):
Fresh independent review of #3159 at 6b6c0d37. The previously posted registry wiring, process-local bus, active-channel root, and agents-subtree findings are resolved, but the current cross-process/channel and degradation paths still have the inline blockers below. CI/UAT gates remain separate.

app/settings/compiler.py
```diff
@@ -434,7 +440,11 @@ def compile_all(*, auto_heal: bool | None = None) -> SettingsBundle:
 
     panel_actions_settings = load_panel_actions_settings()
```
RasmusTho, line 490: [P1] Preserve last-valid/defaults by publishing the compiled projection atomically

`compile_all()` writes `global.yaml`, providers, routing, instance, and agent files before the specialized panel/watcher loaders have validated. If a later loader fails, `ingest_settings()` reports `degraded_last_valid`, but the on-disk projection is already partially replaced. Reproduction on this SHA: compile `global=WARNING` with valid `watchers.md`; edit `global=DEBUG` and set watcher frontmatter to `auto_run: broken`; the invalid ingest correctly leaves the cached bundle at WARNING, but after clearing the process cache (fresh process/restart) `get_settings_bundle()` reads DEBUG from the failed compile. This violates “never substitute/replace while last-valid exists” and makes `invalid_sources/source=defaults` untruthful on cold boot. Compile into staging and atomically publish only after every generic and specialized source validates (or restore the prior projection on failure), with a restart/fresh-process test for a late validation failure.

app/watcher/watcher.py
```diff
@@ -317,6 +324,19 @@ def run_tick(
             hashed = _hash_file(cfg.vault_path / rel)
             if hashed is not None:
                 digest = hashed[0]
+        # A settings source edit (@Settings/*.md) re-ingests the effective bundle
+        # so the running services honor it within one tick (SETTINGS-01 / F1).
+        source_delta = handle_settings_source_delta(rel_path=rel)
```
RasmusTho: [P1] Keep @Settings out of the legacy watcher event flow and bind its configured vault

The registry path now handles this correctly, but the still-exposed `vault-watcher-run`/`vault-watcher-daemon` path calls `handle_settings_source_delta()` without `cfg.vault_path / "@Settings"` and then falls through to `_emit_scan_event`. Reproduction with `WatcherConfig.vault_path=<tmp>/bound`, no `VAULT_ROOT`, and `@Settings/global.md` under the configured vault yields `settings_source_reloads_in_tick=None`, ingestion state `no_vault/defaults`, and `emitted_in_tick=1`. Thus this watcher neither applies its bound vault setting nor treats control markdown as control-only; it publishes the source as ordinary vault content. Pass the configured source directory and `continue` after recording source state, matching the registry implementation; add coverage through this production CLI path.

app/settings/reload_signal.py
RasmusTho: Addressed by 143314e7: the signal path now derives from WATCHER_HEARTBEAT_PATH, so the compose test channel uses the shared /app/tmp-test volume; focused topology coverage passes.

app/watcher/registry.py
RasmusTho: Addressed by 143314e7: registry ticks deduplicate @Settings reloads and record control files as seen without adding them to ordinary changed entries or business-flow emissions; regression coverage passes.

app/settings/ingestion.py
RasmusTho: Addressed by 143314e7: receiving processes derive degradation from their own local valid-bundle history; the shared signal cannot make a fresh process claim degraded_last_valid. Regression coverage passes.

app/settings/compiler.py
RasmusTho: Addressed by 143314e7: compile_all carries the selected vault root into load_watcher_settings; active-channel watcher policy coverage passes.

tests/api/test_capture_inbox_api.py
RasmusTho: Addressed on the current branch: the temporary #3518 traceback instrumentation is absent from the PR diff.

app/settings/compiler.py
RasmusTho: Fixed on PR #3517 head 7aa14d8406dd0a793110041248b9fec9e808192a (commit 7aa14d84): all generic and specialized settings are validated before staging, then the complete projection is published; tests/settings/test_ingestion_startup.py covers a cache-cleared fresh-process late-failure regression. CI settings and full local settings suite are green.

app/watcher/watcher.py
RasmusTho: Fixed on PR #3517 head 7aa14d8406dd0a793110041248b9fec9e808192a (commit cd59c8f1): the legacy watcher passes cfg.vault_path / @Settings and continues before normal event emission. tests/watcher/test_watcher_controls.py covers bound-vault source handling and no outbox event.

PR #3519, opened 2026-07-12
Add calendar as a registered fusion source (ERE-09)

RasmusTho:
Fixes #3184

## Summary
Read-only CalDAV/ICS calendar adapter registered as the third ERE fusion source (ERE-09). Normalizes calendar entries to the ERE-01 signal contract (bitemporal, per-dimension confidence, provenance UID+etag); HEIM-6-honest three-state attendee resolution (read-only, no register mutation); private-bindings credential resolution that fails loud when the registry marks calendar `live` but credentials are absent; soft degradation surfaced via a new `degraded` key on the tick summary. Registry flipped `calendar: planned -> live` in both `docs/EPISODE_RESOLUTION_ENGINE/stream_registry.md` and the README inventory (stream_ids kept backticked for the parity test).

## Architecture fork resolution (see issue comment)
The issue frames ERE-09 as "a new source is a registry entry + adapter, not an engine change." `run_segmentation_tick` (merged via #3179) is not a generic dispatcher — it has per-stream ingestion blocks, and vault.activity was itself added additively there. Per the coordinator's fork resolution on the issue, the enforcement AC means the **shift-detection core is untouched** and the stream is registry-driven, NOT literally zero segmenter bytes. `segmenter.py`: **108 insertions, 0 deletions** — one `_signal_from_calendar_row` normalizer + one additive calendar ingestion block + the `degraded` summary key; `detect_shift`/`_disjoint_set_shift`/`fold_signals_into_segments`/named constants are byte-for-byte untouched (`test_calendar_joins_fusion_via_registry_only` enforces this).

## Validation
- `ruff check app tests` — clean
- `mypy app` — 0 new errors (5 pre-existing in app/builderops/*, app/heimdal/screen_capture.py)
- `pytest -q tests/episodes/test_calendar_adapter.py` — 7 passed
- `pytest -q tests/episodes/test_stream_registry.py` — 9 passed (README-parity AC4 with calendar live)
- `pytest -q -m "not pg"` — passes except verified pre-existing-on-main failures (capture_inbox_api, guard_at_seam session_log:68, model_inquiry_adapters timing flake) and one load-induced flake of tests/ops/test_start_full_runtime_health.py::test_dev_start_full_returns_zero_with_deferred_index_rebuild (passes in isolation in 18s; the calendar-live flip degrades softly and does not break start-full — confirmed by the coordinator). 0 real failures introduced.
- **Live CalDAV receipt deferred to the mac mini test channel** — the laptop is not the runtime env; tests use recorded ICS fixtures / a fake CalDAV response, no live network.

## BuilderOps Routing
- Reason: bounded external-read adapter slice; the architecture fork was resolved as a coordinator contract clarification on the issue (run-state carries the receipt); an owner-optional generic-ingestion-dispatch refactor is filed as a separate follow-up.

---


app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,707 @@
...
+    local datetime, or an all-day ``VALUE=DATE``. No TZID offset table in v1
+    (a local/TZID datetime is treated as UTC-naive-to-UTC, a documented
+    best-effort posture, never a crash) -- ``None`` on anything unparseable,
+    the caller skips the item fail-loud rather than guessing a time."""
+    value = value.strip()
+    match = _DT_UTC_RE.match(value)
+    if match:
+        y, mo, d, h, mi, s = (int(g) for g in match.groups())
+        return datetime(y, mo, d, h, mi, s, tzinfo=timezone.utc)
+    match = _DT_LOCAL_RE.match(value)
```
RasmusTho, line 600: **Review gate round 1 — CONFIRMED (correctness, will bite on real iCloud data):** _parse_ics_datetime treats any non-'Z' DTSTART/DTEND as UTC — but iCloud writes TZID-qualified local times (`DTSTART;TZID=Europe/Stockholm:20260711T090000`), not literal Z. A 9am local meeting becomes 9am UTC = multi-hour skew on the `time` dimension your own confidence table scores 0.95/by_construction (highest trust), silently pushing signals outside their true window and mis-segmenting. Fixtures only exercise the 'Z' branch. Fix: resolve TZID to an offset (zoneinfo) for DTSTART/DTEND; a floating (no-Z, no-TZID) value should degrade confidence or be flagged, not assumed UTC. Add a TZID fixture test.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,707 @@
...
+_CALENDAR_QUERY_BODY: Final[str] = (
+    '<?xml version="1.0" encoding="utf-8" ?>\n'
+    f'<C:calendar-query xmlns:D="{_DAV_NS}" xmlns:C="{_CALDAV_NS}">\n'
+    "  <D:prop>\n"
+    "    <D:getetag/>\n"
+    "    <C:calendar-data/>\n"
+    "  </D:prop>\n"
+    "  <C:filter>\n"
+    '    <C:comp-filter name="VCALENDAR">\n'
+    '      <C:comp-filter name="VEVENT"/>\n'
```
RasmusTho: **Review gate round 1 — CONFIRMED (coverage gap):** the CalDAV REPORT has no <C:expand> and parse_vevent has no RRULE handling — a recurring master VEVENT is returned once and, because signal_id (uid:etag) is constant, contributes exactly once ever; subsequent occurrences never emit. Most real calendars are recurring-dominated, so the 'strongest cheap signal' goes largely blind. Fix: request server-side recurrence expansion (<C:expand start end>) in the query so each occurrence is a distinct instance with its own RECURRENCE-ID/DTSTART (and per-occurrence signal_id). Add an expanded-recurrence fixture.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,707 @@
...
+#: Production uses :func:`_default_transport`; tests inject a fixture/fake
+#: transport (recorded ICS text or a simulated `CalendarUnreachableError`) --
+#: the live-CalDAV receipt is a mac-mini test-channel deliverable, deferred.
+CalendarTransport = Callable[[CalendarBinding], Sequence[CalendarRawItem]]
+
+_DAV_NS: Final[str] = "DAV:"
+_CALDAV_NS: Final[str] = "urn:ietf:params:xml:ns:caldav"
+_CALDAV_TIMEOUT_SECONDS: Final[float] = 15.0
+
+_CALENDAR_QUERY_BODY: Final[str] = (
```
RasmusTho: **Review gate round 1 — PLAUSIBLE (perf, live-endpoint):** the calendar-query filter has no <C:time-range>, so every tick refetches the entire calendar history unbounded. Pairs with the expand fix — add a bounded time-range window (recent + near-future) which both caps the fetch and is required for <C:expand>. Masked in fixtures; surfaces on the mac-mini live channel against an aged calendar.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,707 @@
...
+    dtstart: datetime | None
+    dtend: datetime | None
+    location: str | None
+    attendees: tuple[CalendarAttendee, ...]
+
+
+_DT_UTC_RE = re.compile(r"^(\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})(\d{2})Z$")
+_DT_LOCAL_RE = re.compile(r"^(\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})(\d{2})$")
+_DATE_ONLY_RE = re.compile(r"^(\d{4})(\d{2})(\d{2})$")
+_ATTENDEE_CN_RE = re.compile(r"CN=([^;:]+)", re.IGNORECASE)
```
RasmusTho: **Review gate round 1 — PLAUSIBLE (attendee correctness):** _ATTENDEE_CN_RE `CN=([^;:]+)` doesn't strip RFC 5545 quoted-string quotes, so `CN="Doe, John"` (required when the name contains a comma) captures the literal quotes → exact-match against register labels fails → a known attendee is misclassified AttendeeUnresolved every tick. Fix: strip surrounding double-quotes from the captured CN. Add a quoted-CN fixture.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 504852fa: _parse_ics_datetime resolves TZID via zoneinfo to a correct UTC instant; a floating value (no Z, no resolvable TZID) degrades the time dimension to (0.6, heuristic) via ParsedCalendarEvent.time_is_floating instead of claiming by_construction UTC. TZID + floating fixture tests added.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 504852fa: the REPORT now requests server-side <C:expand> over a bounded window; each occurrence is fanned into a per-occurrence CalendarRawItem with an occurrence_key (RECURRENCE-ID / DTSTART) folded into signal_id + provenance, so occurrences no longer collide on the master uid:etag. Expanded-recurrence fixture added.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 504852fa: bounded <C:time-range> via named constants _CALDAV_TIME_RANGE_PAST (7d) / _CALDAV_TIME_RANGE_FUTURE (60d), shared with the expand window (RFC 4791 9.6.5). Caps the per-tick fetch and enables expansion; documented in code + CALENDAR_STREAM_ADAPTER.md.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 504852fa: _ATTENDEE_CN_RE tries a quoted-string alternative and _strip_cn_quotes removes surrounding quotes before register lookup, so CN="Doe, John" resolves. Quoted-CN fixture added.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,903 @@
...
+                continue
+            etag_el = prop.find(f"{{{_DAV_NS}}}getetag")
+            if etag_el is not None and etag_el.text:
+                etag = etag_el.text.strip().strip('"')
+            data_el = prop.find(f"{{{_CALDAV_NS}}}calendar-data")
+            if data_el is not None and data_el.text:
+                ics_text = data_el.text
+        if not ics_text:
+            continue
+        event_blocks = _split_vevent_blocks(ics_text) or [ics_text]
```
RasmusTho, line 391: **Review gate round 2 — CONFIRMED (idempotency regression):** occurrence_key inclusion in signal_id is gated on `is_expanded = len(event_blocks) > 1` — i.e. on how many siblings of a series happen to fall in THIS tick's window, not a stable property. A series with one occurrence in-window gets `uid:etag` (no key); when a sibling later enters the window, that SAME occurrence becomes `uid:etag:occurrence_key` — a different identity for the identical instance → the segmenter folds it as a new append-only signal → duplicate evidence, breaking the same-occurrence→same-signal_id idempotency the round-1 fix established. Fix: canonical per-occurrence signal_id = `uid` + (RECURRENCE-ID when present), independent of window contents / block count; a truly non-recurring event has no recurrence-id → `uid`. Add a single↔multiple-in-window transition test.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,903 @@
...
+    "goal": (0.5, "heuristic"),
+}
+
+
+def _iso(value: datetime) -> str:
+    dt = value if value.tzinfo else value.replace(tzinfo=timezone.utc)
+    return dt.astimezone(timezone.utc).isoformat()
+
+
+def _calendar_signal_id(uid: str, etag: str, occurrence_key: str | None = None) -> str:
```
RasmusTho: **Review gate round 2 — PLAUSIBLE (edit-noise):** signal_id embeds the resource `etag`, but CalDAV stores a recurring series as ONE resource with ONE etag — editing any one occurrence bumps the shared etag, so every unmodified sibling occurrence in-window gets a new signal_id next tick and is re-emitted as 'new' (noisy duplicate segmentation for stable meetings). Fix: base per-occurrence change detection on the occurrence's OWN content (DTSTART/DTEND/SUMMARY hash, or per-occurrence LAST-MODIFIED/SEQUENCE) rather than the shared resource etag, so editing occurrence N doesn't re-emit M. Add an etag-bump fixture.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 669a2c3e.

`occurrence_key` (`_parse_multistatus` in `app/episodes/calendar_stream.py`) is now derived from the occurrence's own RECURRENCE-ID whenever the parsed event carries one — regardless of `len(event_blocks)`. Production always sends `<C:expand>`, so an RFC-4791-compliant server stamps every expanded instance (including the lone one a narrow tick window returns) with its own RECURRENCE-ID; that field no longer depends on how many siblings happen to also land in this tick's expand/time-range window. The old `is_expanded` gate is kept only as a narrow defensive fallback (DTSTART-based) for a non-compliant server that returns >1 VEVENT under one href/etag but omits RECURRENCE-ID on some instance.

Identity and change-detection are now split via a new `calendar_signal_id` helper (also addresses the sibling finding below): identity is `uid:occurrence_key` (window-independent), and a per-occurrence change-token is appended separately.

New test: `test_occurrence_signal_id_stable_single_vs_multiple_in_window` (`tests/episodes/test_calendar_adapter.py`) — parses the same occurrence once alone in a multistatus response and once alongside a sibling, and asserts the resulting `signal_id`/`provenance_ref` are identical in both cases.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 669a2c3e.

Added `calendar_signal_id` (`app/episodes/calendar_stream.py`), the single canonical identity+change-token builder used by both `calendar_event_to_signal_contract` and `segmenter._signal_from_calendar_row`. For one occurrence of a recurring series, `signal_id` is now `uid:occurrence_key:content_token`, where `content_token` (`_occurrence_content_token`) is a hash of the occurrence's OWN `DTSTART`/`DTEND`/`SUMMARY`, plus `SEQUENCE`/`LAST-MODIFIED` when the server provides them (both newly parsed in `parse_vevent`) — never the shared resource etag. A non-recurring event is unchanged: `uid:etag`.

Identity (`uid:occurrence_key`, finding 1 above) and change-detection (`content_token`) are now properly decoupled: editing occurrence N changes only N's content_token, never M's.

New test: `test_sibling_edit_bumps_shared_etag_but_not_unedited_occurrence_signal_id` (`tests/episodes/test_calendar_adapter.py`) — two occurrences of one recurring series, tick 2 edits only occurrence B's SUMMARY (bumping the shared resource etag for the whole series). Asserts occurrence A's `signal_id` is unchanged across the etag bump, occurrence B's `signal_id` does change (the "edited event gets a new signal" invariant preserved at the occurrence level), and B's identity portion (`uid:occurrence_key`) stays the same across the edit — only its trailing content_token differs.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,1001 @@
...
+            uid = event.uid if event is not None else href
+            if not uid:
+                continue
+            occurrence_key: str | None = None
+            if event is not None:
+                if event.recurrence_id is not None:
+                    # Window-independent (finding 1): this occurrence's OWN
+                    # RECURRENCE-ID, never gated on sibling count.
+                    occurrence_key = _ics_utc_stamp(event.recurrence_id)
+                elif multiple_blocks_this_response and event.dtstart is not None:
```
RasmusTho: **Review gate round 3 — CONFIRMED (round-2 fix incomplete — fallback still window-dependent):** the DTSTART-fallback branch gates occurrence_key on `multiple_blocks_this_response` — the same window-dependence round 2 eliminated on the main path, still present on the no-RECURRENCE-ID path. Same occurrence gets uid:etag alone vs uid:dtstart-key:token with a sibling in-window → identity flip → duplicate fold. Fix: make occurrence identity WINDOW-INDEPENDENT — never gate on block count. Use RECURRENCE-ID when present; when absent, a stable sibling-independent rule (single-event uid-based, or DTSTART unconditionally) + documented residual. Add a no-RECURRENCE-ID single↔multiple-in-window test.

app/episodes/calendar_stream.py
```diff
@@ -0,0 +1,1001 @@
...
+
+    A CalDAV recurring series is ONE resource with ONE ETag: editing any
+    single occurrence bumps that shared ETag for every sibling occurrence.
+    Keying the per-occurrence signal_id's change-token on ETag would
+    therefore give every UNMODIFIED sibling occurrence a new signal_id next
+    tick too, and the segmenter (which treats any new signal_id as fresh
+    evidence) would wrongly re-emit them all as "new". This hash only
+    changes when THIS occurrence's own fields change, so editing occurrence
+    N never re-emits sibling occurrence M."""
+    parts = [
```
RasmusTho: **Review gate round 3 — PLAUSIBLE (hash-input ambiguity):** _occurrence_content_token joins fields with an unescaped '|'; free-text SUMMARY can contain '|', so different structured field sets serialize identically → same token → a real content change missed. Fix: hash a structured encoding (json.dumps of an ordered dict, or length-prefixed fields) so free text can't shift field boundaries.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 9331fe28.

Removed ALL block-count gating from occurrence identity in `_parse_multistatus` (`app/episodes/calendar_stream.py`). The `elif multiple_blocks_this_response and event.dtstart is not None` branch is gone; `occurrence_key` is now derived SOLELY from the occurrence's own RECURRENCE-ID:

```python
occurrence_key: str | None = None
if event is not None and event.recurrence_id is not None:
    occurrence_key = _ics_utc_stamp(event.recurrence_id)
```

Chosen rule (a) — conservative and idempotent: when RECURRENCE-ID is absent, the block is treated as a single (non-recurring) event, so `occurrence_key` stays `None` and its identity is the plain `uid:etag`, identically whether or not a sibling shares the response. **Documented residual** (in the `_parse_multistatus` docstring): a non-compliant server that expands a recurring master into multiple VEVENT blocks WITHOUT stamping RECURRENCE-ID would collapse those blocks onto one `uid:etag` signal (an under-count of that series' occurrences) — deliberately preferred over a sibling-dependent identity because it is stable and idempotent (never a duplicate-fold under at-least-once redelivery), whereas a block-count-gated key is not. iCloud (the sole configured backend) stamps RECURRENCE-ID on every expanded instance, so the residual is hypothetical.

New test: `test_no_recurrence_id_occurrence_signal_id_stable_single_vs_multiple_in_window` — a no-RECURRENCE-ID block parsed alone vs. paired with a sibling in one response yields `occurrence_key is None` and the identical `signal_id` (`no-rid-series:etag-master`) in both cases.

app/episodes/calendar_stream.py
RasmusTho: Fixed in 9331fe28.

`_occurrence_content_token` (`app/episodes/calendar_stream.py`) now hashes a structured, unambiguous JSON encoding instead of a `|`-joined string:

```python
payload = json.dumps(
    {
        "dtstart": _iso(event.dtstart) if event.dtstart is not None else None,
        "dtend": _iso(event.dtend) if event.dtend is not None else None,
        "summary": event.summary,
        "sequence": event.sequence,
        "last_modified": _iso(event.last_modified) if event.last_modified is not None else None,
    },
    sort_keys=True,
    separators=(",", ":"),
    ensure_ascii=False,
)
return hashlib.sha1(payload.encode("utf-8")).hexdigest()[:16]
```

JSON string escaping makes every field self-delimiting, so no free-text value can shift a field boundary. `json` is now a module-level import (removed the redundant lazy import in `resolve_calendar_bindings`).

New test: `test_content_token_summary_with_pipe_does_not_collide` — constructs the concrete forgery that collides under the OLD join specifically: SUMMARY `Team Sync` + SEQUENCE=3 vs. SUMMARY `Team Sync|seq=3` + no SEQUENCE, identical DTSTART/DTEND. I verified the old `"|".join(...)` produces byte-identical strings for these two (→ same SHA), while the new JSON encoding keeps their tokens (and full signal_ids) distinct — so a real edit involving a `|`-containing SUMMARY is never silently dropped as unchanged. The shared identity portion (`pipe-event:20260706T090000Z`) still marks them as the same occurrence.

app/episodes/segmenter.py
```diff
@@ -733,6 +803,40 @@ def run_segmentation_tick(
...
+        # fetch, which the helper already folds into `calendar_degraded`
+        # (AC5) instead of raising. No durable read-position cursor: unlike
+        # the append-only heimdal/outbox logs, a CalDAV poll re-reads the
+        # calendar's current item set each tick and relies on the
+        # uid+etag-keyed `signal_id` for at-least-once idempotency (an
+        # unchanged event redelivers the same signal_id -> no-op fold; an
+        # edited event gets a new etag -> a new signal, itself append-only
+        # evidence of the change) -- the same INV-ERE-F guarantee, a
+        # different mechanism.
+        calendar_items, calendar_degraded = read_calendar_raw_items_for_tick()
```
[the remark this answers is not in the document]
RasmusTho: Captured by the daily review-comment closure audit as residual terminal-PR feedback: #3521. Leaving this thread unresolved until #3521 either lands a fix, proves this is already addressed, or records a supersession/discard receipt.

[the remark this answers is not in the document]
RasmusTho: Audit #3521 classification: STILL REAL. `_CalendarStreamAdapter.advance_cursor` remains intentionally empty while the adapter rereads a fixed window, so closed calendar events lack a durable consumption boundary. Bounded high-priority ready follow-up: #4318.

PR #3565, opened 2026-07-13
Add deterministic CKM repository ingestion

RasmusTho:
Fixes #3141

## Summary
- Add deterministic, local-only docs, tests, source, and git artifact adapters.
- Persist per-source watermarks for idempotent incremental ingestion and expose `ckm ingest --source repo`.

## Validation
- `python -m pytest tests/builderops/ckm -q`
- `ruff check app tests`
- Cold and warm CLI runs against the repository (warm: zero changes)

## BuilderOps Routing
- Records/projections/receipts: AgentWorklog `awl_20260713060937_746d7e73`; BuilderOpsReceipt `receipt_20260713060937_d57add0f`.
- Reason: implementation handoff and publication evidence for CKM-03.


app/builderops/ckm/ingest_repo.py
```diff
@@ -0,0 +1,190 @@
...
+    previous_head = store.get_watermark("git")
+    commits = list(iter_git(root, previous_head, limit=git_limit))
+    changed = 0
+    for artifact in commits:
+        if store.get_artifact_by_source_ref(artifact.natural_key) is None:
+            store.upsert_artifact(source_ref=artifact.natural_key, artifact_kind=artifact.artifact_kind, source=artifact.source, watermark=artifact.source_watermark, provenance=artifact.provenance)
+            changed += 1
+    head = _git(root, "rev-parse", "HEAD").strip()
+    if previous_head != head:
+        store.set_watermark("git", head)
```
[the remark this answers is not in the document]
RasmusTho: Fixed in b010aae9: `--git-limit` now bounds each local `git log` page while ingestion traverses every page before persisting the `HEAD` watermark. The regression test verifies pages 0, 2, and 4 for a five-commit fixture and confirms the watermark is only `HEAD` after all commits are stored.

app/builderops/ckm/ingest_repo.py
```diff
@@ -0,0 +1,214 @@
...
+        "source": _ingest_tree(store, "source", iter_source(root)),
+    }
+    previous_head = store.get_watermark("git")
+    commits = list(iter_git(root, previous_head, limit=git_limit))
+    changed = 0
+    for artifact in commits:
+        if store.get_artifact_by_source_ref(artifact.natural_key) is None:
+            store.upsert_artifact(source_ref=artifact.natural_key, artifact_kind=artifact.artifact_kind, source=artifact.source, watermark=artifact.source_watermark, provenance=artifact.provenance)
+            changed += 1
+    head = _git(root, "rev-parse", "HEAD").strip()
```
RasmusTho: [P1] Advance the git watermark only to the paged snapshot\n\n`iter_git()` snapshots `current_head` before paging, but this second `rev-parse` can observe a newer commit that arrived while the cold scan was running. The code then stores that newer SHA even though `commits` only contains history through the earlier snapshot, so the next incremental run starts after the unseen commit and it is permanently absent from `ckm_artifact`. I reproduced this by committing between the two `rev-parse` calls: the stored watermark equaled the late HEAD while `get_artifact_by_source_ref("git:" + HEAD)` was `None`. Return the original snapshot HEAD from the paging operation (or capture it once in `ingest_repo`) and persist exactly that value; add a regression test that changes HEAD between paging and watermark persistence.

RasmusTho: Fixed in d7c60289 (current head 8cd08c18 after main merges): ingestion now captures HEAD once, passes that immutable snapshot into paged traversal, and persists the same SHA as the git watermark. A regression creates a late commit during traversal, verifies the watermark stays on the traversed snapshot, then confirms the next run ingests the pending commit and advances honestly.

app/builderops/ckm/ingest_repo.py
```diff
@@ -0,0 +1,225 @@
...
+    removed = store.delete_artifacts_not_in(artifact_source, {artifact.natural_key for artifact in materialized})
+    if store.get_watermark(source) != watermark:
+        store.set_watermark(source, watermark)
+    return {"artifacts": len(materialized), "changed": changed, "removed": removed, "watermark": watermark}
+
+
+def ingest_repo(store: CkmStore, root: Path, *, git_limit: int = 500) -> dict[str, dict[str, int | str]]:
+    """Ingest local repository artifacts, preserving idempotent source watermarks."""
+    root = root.resolve()
+    store.ensure_schema()
```
RasmusTho, line 240: [P2] Keep unchanged ingestion row-idempotent

`ingest_repo()` calls `ensure_schema()` on every run, but that method always appends a `ckm_schema_ensured` BuilderOps receipt with a fresh UUID idempotency key. I reproduced an unchanged cold/warm pair where `ckm_artifact` stayed stable but both `builderops_records` and `builderops_idempotency_keys` grew from 1 to 2. That contradicts this issue’s “inserts zero new rows” AC and INV-CKM-7’s requirement that unchanged ingestion produce no new rows; the current regression only checks `result["docs"]["changed"]`. Please make the no-op schema ensure receipt-idempotent (or move it outside the repeatable ingest path) and assert persisted row counts across an unchanged rerun.

RasmusTho: Fixed in f8e25b50: `ensure_schema()` now reuses the first `ckm_schema_ensured` receipt instead of appending a fresh receipt/idempotency key on every warm ingestion. The incremental regression now snapshots row counts for ckm_artifact, ckm_watermark, builderops_records, and builderops_idempotency_keys and requires an unchanged rerun to leave all four identical.

app/builderops/ckm/ingest_repo.py
```diff
@@ -0,0 +1,225 @@
...
+    for line in text.splitlines():
+        if line.startswith("State:"):
+            return line.strip()
+    return "State: (not declared)"
+
+
+def _doc_kind(relative: str, text: str) -> str:
+    if relative.startswith("docs/adr/"):
+        return "adr"
+    if relative.startswith("docs/CAPABILITY_") and "task_id:" in text:
```
RasmusTho: [P1] Classify all task-spec directories, not only CAPABILITY_*

The spec says task files in spec directories must be `spec`, but this prefix gate misclassifies the repository’s other task-spec trees. On this exact head, `docs/SETTINGS_SPINE/PROMPTS_AS_SETTINGS.md` has `task_id: SETTINGS-06` frontmatter yet `iter_docs()` emits `document`; the same live-repo check finds 300 `task_id:` Markdown files outside `docs/CAPABILITY_*` emitted as `document`. That makes the real-path kind-classification AC false and deprives downstream linkers of typed spec evidence. Detect the task frontmatter/spec-directory shape without hard-coding this one directory family, and add a real non-`CAPABILITY_*` regression fixture.

RasmusTho: Fixed in 339eb91f: document kind classification now recognizes a top-level `task_id:` inside the leading YAML frontmatter instead of hard-coding the `docs/CAPABILITY_*` prefix; ADR paths retain precedence. The real non-CAPABILITY task spec `docs/SETTINGS_SPINE/PROMPTS_AS_SETTINGS.md` is pinned as `spec` by regression.

app/builderops/ckm/ingest_repo.py
```diff
@@ -0,0 +1,238 @@
...
+
+def _tree_watermark(items: Iterable[tuple[str, str]]) -> str:
+    return _digest("\n".join(f"{key}:{value}" for key, value in sorted(items)).encode())
+
+
+def _file_artifact(
+    root: Path, path: Path, *, kind: str, summary: str, source: str
+) -> RepoArtifact:
+    relative = _relative(root, path)
+    content_hash = _digest(path.read_bytes())
```
RasmusTho: [P2] Hash the same bytes used for classification and summary

Each tree adapter reads a file once to derive `artifact_kind`/`payload_summary`, then `_file_artifact()` reads it again for `content_sha256`. If an editor or another agent changes the file between those reads, the row can store the old kind/summary under the new content hash; the next run sees that hash unchanged and skips the artifact forever until another edit. I reproduced this by changing a plain doc to task-frontmatter immediately before `read_bytes()`: the persisted row remained `document`, the following ingest reported `changed == 0`, while `_doc_kind()` on disk returned `spec`. Read each file once (or retry when a before/after stat/hash differs) and derive the summary, kind, provenance, and watermark from that single immutable byte snapshot.

RasmusTho: Fixed in 324ef2fb: docs, tests, and source adapters now read each file once as bytes and derive decoding, kind/summary, provenance hash, and watermark from that same immutable snapshot. A TOCTOU regression edits a doc immediately after its captured read, verifies the first row stays self-consistent with the old bytes, then verifies the next ingest detects and reclassifies the new task spec.

PR #3575, opened 2026-07-13
Add compatible Builder Control Plane boundary

RasmusTho:
Fixes #3574

## Problem och scope

Rescue-stashen innehöll ett äldre Control Plane-förslag som inte gick att applicera mot nuvarande dispatcher utan att regressa aktuella kontrakt. Den här PR:n levererar ett sammanhängande, kompatibelt Builder Control Plane-slice: revisionerad control-plane-state, observerbar hälsa, backup/isolated restore och CLI-status från samma SQLite-källa.

## Återanvänt från rescue-stashen

- Den logiska uppdelningen mellan dispatcher, GitHub och projections.
- Health-, backup- och restore-idén samt testinriktningen.
- Heartbeat-förnyelse med fail-loud avvisning efter expiry.

Det som ändrades från rescue: stateövergångar är revisionerade och atomiska; aktuell multi-repo store, stale takeover, Signboard-validering och CLI-kompatibilitet bevaras.

## Filer

- `app/dispatcher/control_plane.py`
- `app/dispatcher/cli.py`, `leases.py`, `store.py`
- `tests/dispatcher/test_control_plane.py`, `test_leases.py`
- `docs/AGENT_ISSUE_DISPATCHER.md`, `docs/development/BUILDER_CONTROL_PLANE.md`, `docs/DOCS_INDEX.md`

## State-, lease- och ownership-kontrakt

- SQLite är den lokala, logiska operational coordination-källan; GitHub Issues/PRs/CI är fortsatt durable delivery truth.
- Modeövergångar använder expected revision och `BEGIN IMMEDIATE`; stale, otillåtna eller ohälsosamma övergångar avvisas utan partiell state.
- Heartbeat förnyar bara giltiga leases; expired leases kan inte återupplivas och befintlig explicit `--takeover-stale` behålls.
- Signboard, Projects och BuilderOps Vault förblir projections/advisory surfaces, inte lås eller hot-path-beroenden.

## Tester och resultat

- `pytest -q tests/dispatcher/test_control_plane.py tests/dispatcher/test_leases.py tests/dispatcher/test_cli.py tests/dispatcher/test_store.py` — 71 passed
- `ruff check app/dispatcher/control_plane.py app/dispatcher/cli.py app/dispatcher/leases.py app/dispatcher/store.py tests/dispatcher/test_control_plane.py` — passed
- `python3 scripts/lint_skills_consistency.py` — passed
- `git diff --check` — passed

## Non-goals

- Ingen fallback-event writer/reconciler eller distribuerat lås.
- Ingen ny service/API, deployment, produktionstrafik eller extern-board-synk.
- Ingen ändring av GitHub-livscykelauktoritet eller branch protection.

## BuilderOps Routing

- Records/projections/receipts: LearningSignal `lrn_20260713081439_b8239f91`.
- Reason: rescue-stashens kontraktsdrift fångades innan implementation och formade det kompatibla slicet.

## Bekräftelse

Ingen deploy, produktionstrafik, branch protection eller aktivt PR-arbete har ändrats.


app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,122 @@
...
+    return next_state
+
+
+def backup(paths: DispatcherPaths, destination: Path) -> dict[str, str]:
+    if not paths.db_path.exists():
+        raise ValueError("Dispatcher database is missing")
+    destination.mkdir(parents=True, exist_ok=True)
+    target = destination / paths.db_path.name
+    with sqlite3.connect(paths.db_path) as source, sqlite3.connect(target) as output:
+        source.backup(output)
```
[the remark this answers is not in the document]
RasmusTho: Addressed in 61ef435b: backup now rejects destinations inside the dispatcher state root, including a target resolving to the source database.

app/dispatcher/cli.py
```diff
@@ -313,9 +313,19 @@ def _cmd_link_pr(args: argparse.Namespace, store: SqliteStore) -> int:
 def _cmd_status(args: argparse.Namespace, store: SqliteStore) -> int:
     paths = load_paths()
     status = singleton_status(paths)
+    control_plane: dict[str, Any]
+    if status["db_exists"]:
+        from app.dispatcher.control_plane import state
+        try:
+            control_plane = state(store)
```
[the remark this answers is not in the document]
RasmusTho: Addressed in 45cff247: status now returns the normal JSON fallback payload with `dispatcher_db_uninitialized` and a control-plane diagnostic instead of failing, so pickup can safely choose label-only coordination.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,122 @@
...
+    if paths.events_path.exists():
+        shutil.copy2(paths.events_path, destination / paths.events_path.name)
+    return {"db": str(target), "events": str(destination / paths.events_path.name), "created_at": _now()}
+
+
+def restore(backup_dir: Path, target: DispatcherPaths) -> dict[str, str]:
+    if target.db_path.exists() or backup_dir.resolve() == target.state_dir.resolve():
+        raise ValueError("Restore target must be a separate, empty state root")
```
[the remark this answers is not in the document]
RasmusTho: Addressed in 61ef435b: restore rejects any non-empty target state directory before creating or copying files, preventing side-file overwrite.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,149 @@
...
+    }
+    if not paths.db_path.exists():
+        result["db"]["error"] = "missing"
+        return result
+    if not paths.events_path.exists():
+        result["events"]["error"] = "missing"
+        return result
+    try:
+        with sqlite3.connect(paths.db_path, timeout=0) as conn:
+            integrity = conn.execute("PRAGMA integrity_check").fetchone()[0]
```
RasmusTho, line 255: **[P1] Verify dispatcher schema before declaring recovery state healthy**

`PRAGMA integrity_check` only proves that a file is structurally valid SQLite, not that it is a dispatcher database. On this head, an empty SQLite file plus an empty `events.jsonl` returns `ok: true`, and `backup()` then emits a successful receipt; `restore()` likewise accepts a database containing only an unrelated table and reports `integrity: ok`. An operator can therefore receive a successful recovery receipt for an unusable dispatcher state. Validate the required dispatcher schema/schema version here and in restore before returning health or restore success.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,149 @@
...
+        raise ValueError("Backup destination must be separate from dispatcher state")
+    proof = health(paths)
+    if not proof["ok"]:
+        raise ValueError("Dispatcher state is unsafe to back up")
+    destination.mkdir(parents=True, exist_ok=True)
+    target = destination / paths.db_path.name
+    with sqlite3.connect(paths.db_path) as source, sqlite3.connect(target) as output:
+        source.backup(output)
+    events_target = destination / paths.events_path.name
+    shutil.copy2(paths.events_path, events_target)
```
RasmusTho: **[P2] Reject an event-log backup-to-self before writing artifacts**

The destination guard covers the state root and database target, but not `events_target`. With the supported custom path configuration (`DISPATCHER_EVENTS_PATH=/x/logs/custom.jsonl`) and destination `/x/logs`, the DB backup is written first and this `copy2` then raises `SameFileError` because source and destination are the same file. The CLI does not catch that exception and leaves a partial backup directory behind. Resolve and reject event-source/event-target equality (including symlinks) before creating or writing any backup artifact.

RasmusTho (commented):
Local review gate on current head 45cff247696e7233b6a05130e8f25a64123d49f1 found two blocking recovery correctness issues.

app/dispatcher/control_plane.py
RasmusTho: Fixed in 69918df3: health and restore now require the dispatcher tables, required columns, and current schema_version in addition to SQLite integrity before emitting successful recovery evidence. Regression coverage rejects empty, unrelated, and shape-incomplete SQLite files without creating a restore target.

app/dispatcher/control_plane.py
RasmusTho: Fixed in 69918df3: backup resolves both DB/events sources and targets and rejects any target/source alias before mkdir or SQLite backup begins. Regression coverage uses a custom DISPATCHER_EVENTS_PATH and proves no partial DB artifact is written.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,203 @@
...
+    row = conn.execute(
+        "SELECT value FROM dispatcher_meta WHERE key = 'schema_version'"
+    ).fetchone()
+    if row is None:
+        return "missing dispatcher schema_version"
+    try:
+        version = int(row[0])
+    except (TypeError, ValueError):
+        return "invalid dispatcher schema_version"
+    if version != SCHEMA_VERSION:
```
RasmusTho: **[P1] Migrate supported v1 stores before requiring the current schema version**\n\nThe shipped `SqliteStore._ensure_schema()` path explicitly supports long-lived schema-v1 databases without rerunning `initialize()`: it adds/backfills `repo`, but it never updates `dispatcher_meta.schema_version`, so the database remains version 1. This new equality check therefore makes `health()` false and blocks backup/recovery for that otherwise supported existing dispatcher (the owner doc calls out that the shared instance is never re-initialized). Update the version atomically as part of the v1 migration, or validate an explicitly supported compatible shape/version here.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,203 @@
...
+        next_state = {"mode": mode, "revision": expected_revision + 1, "updated_at": _now(), "activation_id": activation_id, "health": proof}
+        conn.execute("INSERT INTO dispatcher_meta(key, value) VALUES (?, ?) ON CONFLICT(key) DO UPDATE SET value=excluded.value", (STATE_KEY, json.dumps(next_state, sort_keys=True)))
+        conn.commit()
+    return next_state
+
+
+def backup(paths: DispatcherPaths, destination: Path) -> dict[str, str]:
+    resolved_destination = destination.resolve()
+    target = destination / paths.db_path.name
+    events_target = destination / paths.events_path.name
```
RasmusTho: **[P1] Reject colliding custom backup artifact names**\n\n`DISPATCHER_DB_PATH` and `DISPATCHER_EVENTS_PATH` are independently supported overrides, so two source files in different directories can legitimately have the same basename. In that case `target == events_target`; the SQLite backup is created and then `copy2` overwrites it with the JSONL file, after which this function returns a successful receipt whose `db` and `events` are the same unusable artifact. Resolve and reject DB/event target equality (or use fixed distinct artifact names/a manifest) before writing.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,203 @@
...
+    if (
+        resolved_destination == paths.state_dir.resolve()
+        or resolved_destination.is_relative_to(paths.state_dir.resolve())
+        or bool(source_paths & target_paths)
+    ):
+        raise ValueError("Backup destination must be separate from dispatcher state")
+    proof = health(paths)
+    if not proof["ok"]:
+        raise ValueError("Dispatcher state is unsafe to back up")
+    destination.mkdir(parents=True, exist_ok=True)
```
RasmusTho: **[P2] Do not expose a partial backup before every artifact succeeds**\n\nThe database is written directly into the final destination before the event copy is attempted. For example, if an existing destination contains a directory at the event filename, `copy2` raises `IsADirectoryError` after a valid DB artifact has been left behind; other event-copy failures have the same result. Since that directory can later be mistaken for a completed backup, stage both artifacts in a temporary sibling and publish them only after both succeed (or otherwise clean up all newly written artifacts on failure).

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,203 @@
...
+    shutil.copy2(paths.events_path, events_target)
+    return {"db": str(target), "events": str(events_target), "created_at": _now()}
+
+
+def restore(backup_dir: Path, target: DispatcherPaths) -> dict[str, str]:
+    if backup_dir.resolve() == target.state_dir.resolve() or (
+        target.state_dir.exists() and any(target.state_dir.iterdir())
+    ):
+        raise ValueError("Restore target must be a separate, empty state root")
+    source = backup_dir / target.db_path.name
```
RasmusTho: **[P2] Restore custom-path backups using the source artifact names**\n\n`backup()` names its files from the active source `db_path`/`events_path`, but the CLI constructs this target with only `DISPATCHER_STATE_DIR`, so these lookups always use the default target basenames. A valid backup created while either supported custom path uses a non-default filename cannot be restored by the new CLI (`Backup database/events missing`). Record/read the source names in a backup manifest or use canonical backup artifact names independent of the restore target.

app/dispatcher/control_plane.py
RasmusTho: Fixed in 272f732c: the v1 migration now creates all current supporting tables/indexes and updates schema_version atomically in the same BEGIN IMMEDIATE transaction. Health explicitly accepts a complete compatible v1-shaped database until the normal store-open path upgrades it; incomplete/unsupported shapes remain rejected.

app/dispatcher/control_plane.py
RasmusTho: Fixed in 272f732c: backups now always use distinct canonical artifact names (`dispatcher.sqlite3`, `events.jsonl`) independent of custom source basenames. A regression uses identical custom source basenames in different directories and verifies distinct restorable artifacts.

app/dispatcher/control_plane.py
RasmusTho: Fixed in 272f732c: both artifacts are written to a temporary sibling directory, cleanup runs on SQLite/copy/publish failure, and the completed directory is atomically renamed into a destination that must not already exist. Regression injection proves copy failure leaves neither final nor staging artifacts.

app/dispatcher/control_plane.py
RasmusTho: Fixed in 272f732c: restore consumes the canonical backup pair rather than target path basenames, so CLI/default targets and programmatic custom targets can restore backups created from custom source paths. The custom-source round trip is covered.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,254 @@
...
+        return f"missing dispatcher tables: {', '.join(missing)}"
+    for table, required in _REQUIRED_COLUMNS.items():
+        columns = {
+            str(row[1]) for row in conn.execute(f"PRAGMA table_info({table})").fetchall()
+        }
+        missing_columns = sorted(required - columns)
+        if missing_columns:
+            return f"missing dispatcher columns in {table}: {', '.join(missing_columns)}"
+    row = conn.execute(
+        "SELECT value FROM dispatcher_meta WHERE key = 'schema_version'"
```
RasmusTho, line 124: **[P1] Validate dispatcher constraints, not only table/column names**

`_dispatcher_schema_error()` accepts any database with the named tables/columns and version, even if required primary-key/unique constraints are absent. A column-only schema returns `health()["ok"] == True` and can be backed up/restored successfully, but the first normal metadata write then fails with `ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint`. That makes the recovery receipt claim an unusable store is healthy. Verify the schema constraints/indexes required by `DDL_STATEMENTS` (or perform an equivalent compatibility check) before accepting health/restore.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,254 @@
...
+        shutil.copy2(paths.events_path, staged_events)
+        staging.replace(destination)
+    except (OSError, sqlite3.Error) as exc:
+        shutil.rmtree(staging, ignore_errors=True)
+        raise ValueError(f"Dispatcher backup failed: {exc}") from exc
+    return {"db": str(target), "events": str(events_target), "created_at": _now()}
+
+
+def restore(backup_dir: Path, target: DispatcherPaths) -> dict[str, str]:
+    if backup_dir.resolve() == target.state_dir.resolve() or (
```
RasmusTho, line 341: **[P1] Treat custom restore artifacts as part of the empty target**

This guard checks only `target.state_dir`, but `DispatcherPaths` explicitly allows `db_path` and `events_path` outside that directory. With an empty/nonexistent `state_dir` and existing custom SQLite/JSONL targets, `restore()` returns success after replacing both live files; the same gap also permits either target to alias an artifact inside `backup_dir`. Validate every resolved target path for emptiness/separation/source aliases before writing, not just the state directory.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,254 @@
...
+        raise ValueError(f"Backup database integrity failed: {source_integrity}")
+    if schema_error is not None:
+        raise ValueError(f"Backup database is invalid: {schema_error}")
+    target.state_dir.mkdir(parents=True, exist_ok=True)
+    with sqlite3.connect(source) as input_db, sqlite3.connect(target.db_path) as output:
+        input_db.backup(output)
+    with sqlite3.connect(target.db_path) as conn:
+        integrity = conn.execute("PRAGMA integrity_check").fetchone()[0]
+    if integrity != "ok":
+        raise ValueError(f"Restored database integrity failed: {integrity}")
```
RasmusTho: **[P2] Clean up a restore that fails after publishing the database**

The database is written directly into the final target before the event copy. If that copy fails (for example disk/permission failure, or a custom event destination whose parent is unavailable), the exception escapes the CLI as a traceback and a valid-looking DB remains in the target; the next restore is then rejected because the root is non-empty. Stage the restored pair and publish it together, or remove every newly created target artifact on failure and convert write errors into the command's structured failure receipt.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,254 @@
...
+        with sqlite3.connect(source, timeout=0) as input_db:
+            source_integrity = input_db.execute("PRAGMA integrity_check").fetchone()[0]
+            schema_error = _dispatcher_schema_error(input_db)
+    except sqlite3.Error as exc:
+        raise ValueError(f"Backup database is invalid: {exc}") from exc
+    if source_integrity != "ok":
+        raise ValueError(f"Backup database integrity failed: {source_integrity}")
+    if schema_error is not None:
+        raise ValueError(f"Backup database is invalid: {schema_error}")
+    target.state_dir.mkdir(parents=True, exist_ok=True)
```
RasmusTho: P1: Restore is not atomically published. After the input checks, this creates `target.state_dir` and writes the restored database before copying the events file. If that final copy fails (disk/full permissions/I/O), the target is left non-empty with a DB but no events log; a retry is then rejected by the empty-root guard, and the CLI leaks the raw `OSError` instead of its JSON error contract. Stage DB and events in a sibling temp directory, validate there, then atomically rename it into the target (and clean staging on failure).

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,350 @@
...
+    primary_key = tuple(
+        str(row[1])
+        for row in sorted(columns, key=lambda row: int(row[5]))
+        if int(row[5]) > 0
+    )
+    if primary_key == required_key:
+        return True
+
+    for index in conn.execute(f"PRAGMA index_list({table})").fetchall():
+        if not bool(index[2]):
```
RasmusTho, line 183: **[P1] Reject partial UNIQUE indexes before certifying recovery**

`PRAGMA index_list` exposes whether an index is partial, but this accepts every unique index without checking that flag. A schema with `UNIQUE(task_id) WHERE task_id IS NOT NULL` (and equivalent indexes for the other keys) therefore returns `health()["ok"] == true` and can be backed up/restored, while the first `upsert_task()` fails with `ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint`. Require a non-partial unique index (or validate the exact conflict-target semantics) before emitting a healthy recovery receipt.

app/dispatcher/control_plane.py
```diff
@@ -0,0 +1,350 @@
...
+    row = conn.execute(
+        "SELECT value FROM dispatcher_meta WHERE key = 'schema_version'"
+    ).fetchone()
+    if row is None:
+        return "missing dispatcher schema_version"
+    try:
+        version = int(row[0])
+    except (TypeError, ValueError):
+        return "invalid dispatcher schema_version"
+    if version not in {1, SCHEMA_VERSION}:
```
RasmusTho, line 132: **[P1] Exercise and support the actual canonical v1 database shape**

This advertises schema version 1 as compatible only after requiring the full v2 table/column set, but the canonical v1 fixture in `test_store.py::_write_legacy_v1_db` has only `dispatcher_tasks`/`dispatcher_events`, no `repo` column, and no meta/leases tables. Reproducing with that fixture makes `health()` report missing tables, readonly status fall back, and `backup()` reject the state as unsafe until an unrelated normal store access mutates it through migration. The new test merely rewrites a v2 database version to `1`, so it cannot cover upgrade/recovery of a real v1 store; either stage/migrate the real legacy shape safely for backup/restore or stop claiming v1 recovery compatibility.

app/dispatcher/leases.py
```diff
@@ -206,41 +206,46 @@ def heartbeat(
...
+            raise ValueError(
+                f"Cannot heartbeat lease {task['lease_id']}: held by {lease['holder']}, not {agent_id}"
+            )
+        if lease["released_at"] is not None or _parse_rfc3339(lease["expires_at"]) <= _parse_rfc3339(now):
+            raise ValueError(f"Cannot heartbeat lease {task['lease_id']}: lease has expired")
+        # A heartbeat proves that the current holder is still active; it must
+        # not renew the bounded claim window.  Keep the acquisition-time
+        # expiry so a continuously heartbeating worker can still be reclaimed
+        # once its finite TTL elapses.
+        expires = lease["expires_at"]
```
RasmusTho: **[P1] Preserve heartbeat renewal before allowing stale takeover**

The canonical root `AGENTS.md` dispatcher policy says the ~30-minute heartbeat renews the 90-minute lease before expiry; this change instead fixes expiry at acquisition and edits only the subordinate dispatcher doc to match. In a timed reproduction, successful heartbeats at +30m and +60m leave expiry at +90m, heartbeat at +90m is rejected, and `--takeover-stale` at +91m gives the task to another agent while the original worker followed the prescribed cadence. Rejecting a heartbeat after expiry is correct and satisfies the issue constraint, but a pre-expiry heartbeat must atomically extend both lease/task expiry so an active holder is not displaced.

app/dispatcher/control_plane.py
RasmusTho: Addressed by 28c74985 and 643f708c: recovery validates the required primary/unique conflict targets and rejects partial unique indexes; regression coverage proves unusable schemas are not certified.

app/dispatcher/control_plane.py
RasmusTho: Addressed by 6fa355f5: restore validates custom targets and source aliases before publication, with custom-path regression coverage.

app/dispatcher/control_plane.py
RasmusTho: Addressed by 6fa355f5: restore stages and validates the complete artifact pair, cleans failures, and atomically publishes the target.

app/dispatcher/control_plane.py
RasmusTho: Addressed by 6fa355f5: restore now stages both artifacts, validates the staged DB, atomically publishes, and removes staging on failure.

app/dispatcher/control_plane.py
RasmusTho: Addressed by 643f708c: schema validation now rejects partial UNIQUE indexes, with a regression that preserves the recovery evidence and prevents backup/restore certification.

app/dispatcher/control_plane.py
RasmusTho: Addressed by c870752f: recovery now recognizes the exact canonical pre-multi-repo v1 shape, permits health/backup/isolated restore without mutation, and the normal store path migrates the restored artifact to v2. The exact fixture round-trip passed in two fresh review rounds.

app/dispatcher/leases.py
RasmusTho: Addressed by c870752f: heartbeat now atomically renews both lease and task expiry for the original TTL while rejecting expired/released leases. Timed renewal and stale-takeover behavior passed in two fresh review rounds and matches root AGENTS.md.

PR #3584, opened 2026-07-13
Add zero-write contradiction triage adapter

RasmusTho:
Fixes #3544

## Summary
- add a typed read adapter over the delivered contradiction harness using `materialize=False`
- require two citation handles to resolve in the caller's current scope and fail closed with content-free diagnostics
- preserve the existing note/Panel proposal path without vault, outbox, receipt, or suppression writes
- register curation paths with subsystem-owned CI selection so the production, test, and owner-doc surfaces fail closed into the intended coverage

## Validation
- `pytest -q tests/scripts/test_select_pr_tests.py tests/curation/test_contradiction_triage_adapter.py tests/curation/test_contradiction_citations_resolve.py tests/curation/test_semantic_never_autowrites.py` (64 passed after rebase)
- `pytest -q tests/invariants/test_curation_invariants.py -k 'citations_resolve or semantic_curation_never_autowrites'` (1 passed)
- `ruff check app tests companion-ui/companion-app scripts/select_pr_tests.py` (passed)
- `python3 scripts/select_pr_tests.py ...` (curation/ops_deploy owned targets; no unowned paths)

## SBS Impact
Product/Runtime Curation + Retrieval admission. Read-only derived output only; no human-knowledge, memory, persistence, sync, or write authority is added. Citation and scope failures remain fail-closed.

## Owner-doc writeback
Updated `docs/CONTRADICTION_TRIAGE_BENCH/README.md :: Current foundation and boundary` with the supported adapter and diagnostic contract.

## BuilderOps Routing
- Records/projections/receipts: local epic run-state `cognitive-experiments-3543-20260713`; LearningSignal `lrn_20260713135209_cdd84ba6`
- Reason: the CI selector initially had no curation owner; the PR now repairs `scripts/select_pr_tests.py :: SUBSYSTEMS` and records the workflow divergence

### DIFF
--- a/app/curation/contradiction_triage.py
+++ b/app/curation/contradiction_triage.py
@@ -134,7 +134,9 @@ def run_contradiction_triage(
             continue
         admitted.append(projected)
 
-    if admitted:
+    if admitted and (excluded or report.suppressed_by_cross_scope_denial):
+        diagnostic = "partial_findings"
+    elif admitted:
         diagnostic = "ok"
     elif excluded or report.suppressed_by_cross_scope_denial:
         diagnostic = "finding_excluded"
--- a/app/curation/contradiction_triage.py
+++ b/app/curation/contradiction_triage.py
@@ -1,0 +1,157 @@
+"""Scope-safe, zero-write read adapter for contradiction findings.
+
+The adapter deliberately calls the delivered contradiction harness with
+``materialize=False``.  It exposes only findings whose two citation handles
+resolve in the caller's current admission scope; excluded details and scan
+exceptions are collapsed to content-free diagnostics.
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
+from typing import Callable
+
+from app.curation.contradiction import (
+    ContradictionPassConfig,
+    ContradictionPassReport,
+    run_contradiction_pass,
+)
+from app.curation.findings import CurationFinding, FindingClass
+
+
+class ContradictionTriageScanStatus(str, Enum):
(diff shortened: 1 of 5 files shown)


app/curation/contradiction_triage.py
```diff
@@ -0,0 +1,157 @@
...
+
+    admitted: list[ContradictionTriageFinding] = []
+    seen_ids: set[str] = set()
+    excluded = False
+    for finding in report.findings:
+        if finding.finding_id in seen_ids:
+            continue
+        seen_ids.add(finding.finding_id)
+        projected = _admit_finding(finding, admission)
+        if projected is None:
```
RasmusTho, line 132: [P1] Preserve exclusion posture when some findings are admitted

This branch returns `ok` whenever at least one finding is admitted, even if `excluded` is also true (for example, a report containing one current-scope pair and one denied/unavailable pair). The excluded logical finding then has no per-finding result and the global result is indistinguishable from a fully successful scan, which violates the specified uncertainty/exclusion posture and can let a consumer treat an incomplete scan as complete. Please make the content-free result outcome retain the exclusion/partial posture (and add a mixed admitted+excluded regression test), rather than letting `admitted` mask `excluded`.

RasmusTho: Fixed in 23354e743bb073fefdfd3bb812a0ef431af18363: mixed admitted+excluded results now return the content-free `partial_findings` posture, with a regression test proving admitted evidence remains visible while denied handles/content do not leak. Local post-rebase validation: 52 adapter+selector tests passed and ruff passed. CI and independent re-review are pending on this head.

PR #3696, opened 2026-07-14
Migrate legacy objects FK consumers to canonical store

RasmusTho:
Fixes #3510

## Summary
- add a forward-only Alembic cutover that inventories and transactionally backfills missing canonical parents and retargets every supported live `objects(id)` FK to `store_objects(object_id)` while preserving delete semantics
- fail before mutation on unknown FK consumers or orphaned canonical parents, with explicit repair/retry guidance
- remove the `PgObjects` compatibility-parent write and migrate decision receipt/projection plus audit producer assumptions
- make the live `FilesystemVaultAdapter` compatibility producer atomically write the canonical `store_objects` parent with its retained `objects` mirror, file state, and outbox event
- update all retention/export/rebuild fixtures plus schema and single-store owner docs

## SBS Impact
- Primary subsystem: Product/Runtime System — PDM durable persistence schema
- Secondary subsystems: ingestion, filesystem watcher, decision receipts/projection, audit
- Write class: mechanical durable / migration
- Authority impact: none
- Persistence impact: legacy FK ownership moves to canonical `store_objects`; historical `objects` remains readable and is retained as an atomic watcher compatibility mirror
- Derived/rebuildable impact: decision/audit projections preserve retention and relinking semantics
- Human knowledge / memory impact: none
- Sync/deployment impact: forward-only migration; rollout requires the normal reviewed promotion plan and operator acknowledgement. This PR performs no test/prod mutation or promotion.
- Contract change: all live ingest producers now establish the canonical FK parent; canonical vault-root ingest does not fabricate a legacy parent
- Owner-doc impact: [x] updated in this PR (`docs/DB_SCHEMA.md`, `docs/RUNTIME_CORRECTNESS_KERNEL/SINGLE_STORE_GENERATION.md`)
- Transition debt: reduces the #2901 deferred legacy-parent dependency; no new debt
- Fitness impact: strengthens single-store-generation enforcement with a closed FK inventory and production-entrypoint regression

## Acceptance Ledger
- [x] AC1 inventory: `tests/migrations/test_legacy_objects_fk_inventory.py::test_all_legacy_objects_fk_consumers_are_accounted_for`
- [x] AC2 migrated fresh ingest: `tests/ingest/test_vault_root_ingest_pg.py::test_ingest_persists_decisions_without_legacy_objects_parent_after_migration`
- [x] AC3 fail-loud unsupported state: `tests/migrations/test_legacy_objects_fk_migration.py::test_legacy_objects_fk_migration_fails_loudly_on_unsupported_state`
- [x] invariant→producers regression: `tests/ingest/test_vault_root_ingest_pg.py::test_filesystem_vault_adapter_upsert_creates_canonical_decision_parent`

## Validation
- review reproduction plus live adapter regression: `DATABASE_URL=<isolated-pg> pytest -q tests/stores/test_decisions_fk_semantics.py tests/jobs/test_decisions_export.py tests/integration/test_decisions_rebuild_from_log_only.py tests/ingest/test_vault_root_ingest_pg.py::test_filesystem_vault_adapter_upsert_creates_canonical_decision_parent` — 18 passed
- post-rebase PG migration/producer/atomicity set — 24 passed
- post-rebase not-pg producer lifecycle, tombstone, watcher, atomicity, and architecture set — 21 passed
- live schema assertion on the isolated DB — 7/7 legacy consumer FKs target `store_objects`; zero remain on `objects`
- `ruff check app tests` — passed
- `mypy app` — passed (733 files)
- historical full `pytest -q -m "not pg"` checkpoint before later review repairs — 8292 passed, 114 skipped, 18 xfailed; this is historical evidence only. Final authority requires fresh exact-head GitHub Unit CI.

## Review Repair
The blocking P1 at issue comment `4974357396` is addressed at head `63e7660d`: `sync_markdown`, `upsert_object_from_note`, rename, path update, and tombstone paths now keep the canonical row transactionally aligned through the sole `store_objects` SQL writer. Decision FK retention/export/rebuild fixtures seed/delete the canonical parent, and the owner docs no longer overstate removal of the watcher compatibility mirror. Clean-review count restarts from zero.

## Migration / Recovery
Revision `7e4f2a1c9d30` is forward-only. It transactionally backfills supported legacy rows into missing canonical parents, then aborts if an unreviewed FK or orphan remains. Reconcile any unsupported orphan from the vault/continuity source, verify zero orphans, then rerun `alembic upgrade head`; do not fabricate legacy parents or downgrade in place.

## Claim Receipt
- coordination mode: `github-label-only-fallback`
- fallback reason: `dispatcher_db_uninitialized`
- evidence: issue comment `4974185014`
- agent/session: `codex-issue-3510` / `ready-pool-20260714`

## BuilderOps Routing
- Records/projections/receipts: `lrn_20260718170347_a1cdb380`
- Reason: the existing topology merge missed a concurrently landed current-main migration head; captured as a workflow divergence for the `pr-integration` migration-topology readiness procedure.

## Decision-Writer Preflight Repair
The blocking round-1 finding at issue comment `4974562849` / inline thread `3582871732` is addressed at head `504bb7d6`. The active `app.services.decisions.insert_decision` path now performs a SELECT-only FK-shape preflight through `conn_ro` before `append_decision_receipt`. A database still at revision `4d1e0c9a3329` raises `DecisionsSchemaMigrationRequired` with `#3510` and `alembic upgrade head` guidance, with zero receipt records and zero decision rows. Receipt-before-ack remains unchanged after the schema gate.

Additional evidence:
- exact pre-cutover production-entrypoint regression — 1 passed
- full repaired PG AC/producer/receipt suite — 34 passed
- focused receipt/producer/architecture suite — 27 passed
- post-rebase migrated writer + export/rebuild suite — 15 passed
- post-rebase receipt suite — 6 passed
- `ruff check app tests` — passed
- `mypy app` — passed (733 files)

Clean-review count remains zero and restarts only after fresh independent review.

## Final Producer-Mechanism Repair
The blocking active-edit finding at issue comment `4974666611` is addressed at head `dae51bad`. `sync_markdown` now checks both `store_objects` and the retained `objects` mirror before its unchanged/ok early return. The explicit active-edit return remains `deferred` and does not mutate object/outbox state; the next non-active pass treats either missing producer row as a materialization change, atomically dual-writes both rows, and emits `ingest.object.created`. Pure-rename continues through `_update_path_only`, which already dual-writes; every completed-ok early return is therefore producer-safe.

Fresh Alembic-head production regression proves: first pass deferred with one file_state and zero object rows; second unchanged pass ok with both parent rows and one created outbox event; subsequent active `insert_decision` succeeds against the migrated FK.

Final evidence:
- exact deferred-first/unchanged-second production regression — 1 passed
- broader PG migration/producer/receipt/retention/export/rebuild/delete/audit suite — 41 passed
- focused not-pg receipt/producer/delete/architecture suite — 27 passed
- `ruff check app tests` — passed
- `mypy app` — passed (733 files)

Clean-review count remains zero. This is the final missing-producer mechanism repair; fresh independent review must stop rather than continue repair if the same mechanism recurs.

## Locator Backfill Repair
The blocking migration-backfill finding at issue comment `4974774851` is addressed at head `61e1fcac`, rebased on current `origin/main`. Revision `7e4f2a1c9d30` now chooses the locator row-by-row: non-null legacy `source_ref` wins; otherwise legacy watcher `path` is used; if both are null the canonical locator remains null. Kind, nested JSON payload, and timestamp/default handling remain faithful to the legacy row.

The `sync_markdown` completed-ok fast path now requires locator completeness as well as row presence. A present canonical parent with null or stale `source_ref` is routed through the existing atomic dual-write path and healed to the current note path; presence alone cannot cement incomplete backfill state.

Evidence:
- exact pre-cutover→head locator/payload regression plus unchanged-sync healing regression — 2 passed
- broad PG migration/producer/receipt/retention/export/rebuild/delete/audit suite — 43 passed
- focused receipt/producer/doctor/rebuild suite — 41 passed, 5 environment skips
- post-rebase critical migration/producer/preflight set — 5 passed
- `ruff check app tests` — passed
- `mypy app` — passed (733 files)

The prior decision-preflight inline thread `3582871732` was replied to with current commit evidence and resolved administratively. Clean-review count remains zero; two fresh independent rounds are required.

## Decision-Receipt Owner-Doc Repair
Round-1 owner-doc drift comment `4974907060` is addressed at head `1d770a2b`, rebased on current `origin/main`. `docs/DECISION_RECEIPT_LOG/README.md` now states the shipped identity contract: `decisions.object_id` is the canonical FK to `store_objects.object_id`; retained `objects.uuid` supplies continuity when a matching legacy row exists; fresh canonical-only objects do not invent a continuity UUID and therefore emit null unless a retained mapping exists; rebuild accepts an existing canonical object id first and otherwise re-links through the retained continuity UUID mapping. The operator-gated historical export posture and retained compatibility surface are unchanged.

Docs governance: existing owner spec updated; `DOCS_INDEX` already assigns ownership and requires no role/index change; MACHINE_MIRROR rule-4 projection contract remains aligned.

Evidence: `DOCS_GUARD_ALLOW_TEMPORAL_SKIP=1 python3 scripts/docs_guard.py` passed; relevant DOCS_INDEX/machine-mirror/docs-guard/receipt/rebuild tests 20 passed, 3 environment skips; diff check passed. Clean-review count remains zero; two fresh whole-diff rounds are still required.
Governing-Issue: #3510

## Alembic Topology Repair
The exact-head verifier found concurrent heads `a2f1c3e4d5b6` and `f5a6b7c8d9e0`. Revision `b7e8f9a0b1c2` is a forward-only, schema-noop topology merge over both; it restores exactly one `alembic upgrade head` target without a migration waiver, downgrade, or special case.

Validation: `alembic heads` and a `ScriptDirectory` assertion report only `b7e8f9a0b1c2`; `ruff check app tests` passed; isolated-PG migration/inventory/ingest suite completed successfully in the dedicated container-network runner.
## Canonical Decisions FK Parent Preflight Repair

The current head `97cddd67` fixes the blocking schema-preflight gap in the retained `PgDecisions` adapter. `_ensure_decisions` now requires the full FK identity `public.decisions(object_id) -> public.store_objects(object_id) ON DELETE SET NULL`, not merely a nullable column, UUID default, and delete rule. A legacy `objects(id) ON DELETE SET NULL` constraint therefore fails with the migration hint before any decision INSERT.

Regression coverage now includes both a schema-assertion unit test and an isolated-PostgreSQL pre-cutover database: the latter constructs the legacy parent, proves the adapter raises before writing a decision row, and preserves the existing production-path receipt-before-ack preflight proof. The #2788 SET NULL fixture was also aligned to the post-#3510 canonical parent.

Validation on `97cddd67`:
- isolated PostgreSQL migration/inventory/schema/ingest suite — 13 passed
- focused non-PG decision receipt/export/rebuild suite — 10 passed, 10 environment skips
- `ruff check app tests` — passed
- `mypy app` — passed (783 files)
- full `pytest -q -m 'not pg'` was attempted twice but the runner stream detached/stalled around 11%; its two local processes were stopped to avoid contention. No failure was observed; this is not a waiver and the exact-head CI gate remains required.

Clean-review count restarts from zero; fresh independent review and exact-head CI are still required.

## Static-Quality Repair
The current head `e4223354` moves the migration-owned decisions FK-shape preflight from the deprecated `app.stores` adapter into the neutral `app.db.decisions_schema` seam. The retained adapter and destructive projection rebuild now share the same fail-loud check without adding a new deprecated-store caller; the payload-sink census remains a closed, explicit classification with the relocated real sink line.

Validation on `e4223354`:
- exact static CI regressions — 5 passed (`test_no_new_store_callers`, `test_store_payload_sink_census_is_closed`, and decisions schema assertions)
- focused non-PG receipt/export/rebuild/property set — 23 passed
- fresh isolated PostgreSQL pre-cutover/rebuild set — 9 passed
- `ruff check app tests` and `mypy app` — passed

The local PostgreSQL service on port 15432 was corrupt and was not used as evidence; the focused PG result above used the dedicated healthy PR container on port 25433. Exact-head GitHub Unit CI remains required; no waiver is claimed.\n## Review Round 1 Consumer Repair

Exact head `468aaf87` addresses all four Sol/high findings: lifecycle events now carry the canonical key as `uuid`/`object_id` and retain frontmatter identity as `vault_uuid`; index/delete consumers prefer `object_id`; backfill scans canonical `store_objects`; store reset uses ordered transactional DELETEs rather than unsafe TRUNCATE CASCADE; receipt resolution returns null without a proven legacy UUID mapping.

Validation: focused watcher/outbox/backfill/reset/receipt/rebuild suite — 27 passed, 3 environment skips; `ruff check app tests` and `mypy app` (786 files) passed. Fresh exact-head PG/CI and two clean review rounds remain mandatory.

## Final Canonical-Identity and Migration-Shape Repairs
Exact head `a9a65019f388af4965ff08a73c3b7b5584b0add6` preserves retained-row identity across vault ingest, sync, and watcher deletion, and makes the forward-only backfill compatible with the active runtime-bootstrap `objects` shape where `updated_at` is absent (`created_at`, then database time, is the explicit fallback). The canonical continuity contract is null rather than fabricated when no retained UUID mapping exists.

Final local evidence on the dedicated PostgreSQL instance: 34 passed, 11 deselected across the migration, inventory, ingest, decision-FK, rebuild, retention, and reset gate; ruff, mypy, and diff-check passed. All active GitHub review threads are resolved. Fresh exact-head CI, current-main merge-ref validation, and two independent clean whole-diff reviews remain required before merge.

## Final Payload-Preservation and Transactional-Preflight Repairs
Exact head `10213f52197fbe6b76e2cf5e2e1ee1d17cf47698` integrates current main `317dca928444f38013671d8c12b45f52697c7f53`. Watcher edit and pure-rename paths merge their authoritative fields into the locked existing canonical payload instead of replacing richer metadata. Every connection-level canonical object write now asserts the full migration-owned store table and embedding-identity shape on the caller-owned transaction before any object/outbox mutation; boot-time data repairs remain boot-only.

Regression evidence before publication: focused lifecycle/identity set 28 passed; integrated PG checkpoint 42 passed plus one exact AC clean rerun after transient DB health-state contamination; Ruff, mypy, and diff-check passed. Fresh exact-head full PG, GitHub CI, merge-ref validation, and two independent clean whole-diff reviews remain mandatory before merge. BuilderOps: `lrn_20260718204613_d9cdfc8e`, `lrn_20260718211035_5eff0ea2`.

## Final CI Architecture and Census Repair
Exact head `37b7567def9298667979607b911c364546a26dee` integrates current main `f67070ea5c7bbd74b32d0d5edc89100167a2e030`. Retained UUID lookup is routed through the canonical Postgres store boundary, eliminating the forbidden direct DB dependency. Mirror and store-payload census entries are pinned to the real post-integration vault-alpha/indexer sinks; the episode-ref producer invariant remains closed. Local no-direct-import, census, episode-ref, and mirror fault-injection gates are green; Ruff, mypy, and the single Alembic head `c8d9e0f1a2b3` pass. Fresh exact-head CI, full PG, merge-ref validation, and two independent clean whole-diff reviews remain mandatory before merge.

## Current-Main Integration and Final Identity Snapshot Gate
Exact head `ad84f0c4d8b3c0d854b8f97c929406c9d12519ca` integrates current main `9c43820056cc1f20c1e7f62b0261d7e6c113d442`. The only new main paths are the dispatcher verification consumer and its tests; whole-diff review found no semantic overlap with #3510. The forward-only migration now locks both retained and canonical producer tables for the full preflight/backfill/FK-cutover transaction and fails before mutation when one retained `objects.uuid` names another retained `objects.id`.

Exact-head evidence: isolated PostgreSQL migration/ingest/FK matrix — 40 passed, 15 deselected; independent whole-diff review round 1 — clean with 281 passed/5 deselected, Ruff and single Alembic head `c8d9e0f1a2b3`. Exact-head CI and independent review round 2 remain mandatory before merge. BuilderOps: `lrn_20260718221654_e2c2f947`.

## Watcher Panel Canonical-Identity Repair
Exact head `df927a0cb23ae62e49ae82726fb7c453f80f9651` addresses the final review P1: after resolving a retained frontmatter UUID, every ObjectStore-facing watcher panel read, hydration, and executed-action write now uses the canonical `objects.id`; the retained vault UUID remains only the filesystem/companion identity. A production-entrypoint regression proves a checked panel action updates the sole canonical parent when `objects.uuid != objects.id` and never creates a split `store_objects` row.

Pre-publication evidence: focused regression plus existing watcher policy — 2 passed; broader watcher/lifecycle/indexer/delete set — 42 passed; Ruff and diff-check passed. Fresh exact-head PG, CI, merge-ref, and two clean whole-diff reviews remain mandatory.

## Forward-Only Recovery and Runtime-Bootstrap PK Repair
Exact head `a49e83bc22863babc6dc6d7c7a252026b55628fd` addresses the two final P1 findings. If an acknowledged forward-only migration has already run, a failed deployment retains the schema-compatible target image for a forward fix instead of automatically restoring an older, potentially incompatible producer. The runtime bootstrap now detects an existing exact `objects(id)` primary key and skips the legacy DROP/ADD pair, avoiding repeated AccessExclusive locking and index rebuild after FK cutover. The merged head also resolves raw-ingest object identifiers through the retained canonical mapping before indexing.

Pre-publication evidence on the dedicated worktree: combined DB/deploy/indexer/watcher suite — 36 passed; isolated PostgreSQL migration/ingest/FK matrix — 41 passed, 15 deselected; Ruff and diff-check passed. All previous CI and review evidence is historical after this head movement. Fresh exact-head GitHub CI, two independent clean whole-diff reviews, current-main review-thread census, and merge-ref validation remain mandatory.

## Producer-Drain, Applied-Migration Recovery, and Cross-Key Collision Repair
Exact head `ea76628e91f9455d39fc940d22e1bfb1e6a68aaa` addresses both independent P1 review clusters and the stale exact-head CI failures. Deploy now distinguishes a classified forward-only migration from one that completed successfully: pre-migration failures restore the previous pin, while only a successfully applied forward-only cutover retains the compatible target. When migrations changed, all runtime writers are stopped before the one-shot migration authority runs, so a legacy writer cannot resume after the cutover snapshot and create legacy-only state. Runtime retained-UUID resolution also fails loud when that UUID already names a different canonical object.

The moved indexer sinks are reclassified at their actual lines in both closed censuses; deploy tests now assert the shipped recovery contract. Pre-publication evidence: focused deploy/census/indexer/watcher/object-store set — 71 passed; isolated PostgreSQL migration/ingest/FK matrix — 42 passed, 15 deselected; Ruff, shell syntax, and diff-check passed. BuilderOps: `lrn_20260718225546_299b6c79`. Fresh exact-head CI, two independent clean reviews, active-thread census, and merge-ref validation remain mandatory.

## Rollback Separation and Resolver-Fake Repair
Exact head `525bcbffa5f3c11b4af2f920feea77b1f8f122db` resolves both exact-head review findings. The explicit changed-migration execution path now runs only for deploy; rollback remains under the governed rollback-promotion downgrade/compatibility authority and never asks an older target image to upgrade a newer stamped database. All three production-near lifecycle/deletion/tombstone transaction fakes now model the resolver query two-parameter input and `canonical_alias_exists` output.

Focused reproduction across deploy, lifecycle, deletion, tombstone, and closed censuses — 72 passed; Ruff and diff-check passed. All earlier reviews and CI are historical after this head movement. Fresh exact-head CI, two clean independent reviews, active-thread census, and merge-ref validation remain mandatory.

## Ambiguous Migration Outcome and Duplicate-FK Repair
Exact head `68d401b27d84c60a2f099e3d8367008e4e46af5c` resolves the latest P1/P2 review cluster. Forward-only deployment now distinguishes not-started, started, and applied states: a pull or pre-execution failure restores the previous pin, while any lost/nonzero result after execution begins is treated as possibly committed and retains the schema-compatible target until unchanged DB revision or compatibility is proven. The migration rejects multiple legacy FKs on one reviewed child column before backfill or retargeting, so it cannot stamp a schema that runtime writers reject for duplicate constraints.

Focused deploy/lifecycle/census evidence — 73 passed; isolated PostgreSQL migration/ingest/FK matrix — 43 passed, 15 deselected; Ruff, shell syntax, and diff-check passed. Fresh exact-head CI, two clean reviews, active-thread census, and merge-ref validation remain mandatory.

### Exact-head production-panel identity repair (b714500f)

- Exact head: `b714500f20e62168c07404037b6ed790698e1ff9`
- Canonicalizes active watcher and outbox panel paths before cognition/materialization while preserving the vault UUID for filesystem and latency identity.
- Regression coverage: `tests/watcher/test_panel_watcher_outbox_db.py::test_process_panel_note_uses_canonical_store_identity` and `tests/runtime/test_worker_ingest_event.py::test_handle_panel_scan_requested_uses_canonical_store_identity`.
- Local evidence: 34 focused panel/watcher/runtime tests passed; Ruff and `git diff --check` passed.
- Active unresolved non-outdated review threads at this head: 0.
- Exact-head CI and two independent exact-head review gates are in progress; their results will be recorded before merge.

## Canonical Identity Review Closure
Exact head `6dd8e8c8ca9cc84edb4ca7d7befa5fd7ed2e64d7` resolves the late exact-head P1/P2/P3 review cluster. Retained vault UUID is threaded separately to companion hash refresh; lifecycle event identity and retained/canonical lookup share single constructors/joins; the migration SQL allowlist is rendered from its governed frozenset; decision schema validation reuses the shared seam and caches successful checks per DSN; reset names every CASCADE child explicitly; the isolated vault-sync test now uses the production schema producer; and rename avoids duplicate canonical resolution.

Pre-publication evidence: focused review-regression set — 43 passed, 7 deselected; Ruff, mypy (789 files), import-linter (745 files / 1,921 dependencies), compileall, and diff-check passed. Local PostgreSQL/Docker was unavailable because the machine-local Colima daemon is down, so exact-head GitHub CI plus fresh isolated-PG evidence remain mandatory. All earlier clean reviews are historical after this head movement; two new independent whole-diff rounds are required.

## Checkbox Projection and Deferred-Baseline Repair
Exact head `f3c2af09fc200ef7216b3a2089652be2881b396c` resolves both local review-gate P1 findings. The active-edit deferred baseline again calls `_upsert_file_state` with its real signature; the already-resolved canonical id is passed only to `_update_path_only`. Companion checkbox projection now resolves the retained frontmatter UUID before ObjectStore refresh/execution and threads the retained value separately as `vault_uuid`, matching watcher/outbox identity semantics.

Focused evidence: panel/worker/runtime/vault-sync set — 42 passed, 5 deselected; vault-sync lifecycle — 10 passed; Ruff, targeted mypy, and diff-check passed. All earlier CI and clean-review evidence is historical after this head movement; exact-head CI, PG acceptance, and two fresh independent whole-diff reviews remain mandatory.


app/stores/postgres.py
```diff
@@ -47,27 +47,38 @@ def _ensure_decisions(conn) -> None:
...
                 WHERE tc.table_schema = 'public'
                   AND tc.table_name = 'decisions'
                   AND tc.constraint_type = 'FOREIGN KEY'
                   AND kcu.column_name = 'object_id'
                 LIMIT 1;
 
+                IF fk_name IS NOT NULL AND referenced_table IS DISTINCT FROM 'store_objects' THEN
+                    RAISE EXCEPTION USING
+                        MESSAGE = '#3510 decisions FK migration is required',
+                        HINT = 'run alembic upgrade head before starting the runtime';
```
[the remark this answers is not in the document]
RasmusTho: Addressed in the active writer at current rebased head `61e1fcac` (original repair commit lineage: `504bb7d6`). `app.services.decisions.insert_decision` now runs a SELECT-only `decisions.object_id` FK-shape preflight through `conn_ro` before `append_decision_receipt`. The production regression upgrades to exact pre-cutover revision `4d1e0c9a3329` and proves the call raises `DecisionsSchemaMigrationRequired` with `#3510` / `alembic upgrade head`, while writing zero receipt records and zero decision rows. Migrated receipt-before-ack behavior remains covered by the focused receipt and PG suites.

docs/DB_SCHEMA.md
```diff
@@ -192,6 +192,20 @@ The tables below are the legacy AMG-core schema defined by Alembic migrations un
 remaining shapes here are historical lineage. Do not read these shapes as the current contract. Note
 there is no `search_vector` column anywhere in the schema (active or legacy).
 
+## #3510 legacy-FK cutover (current reality)
+
+Alembic revision `7e4f2a1c9d30` moves every inventoried live single-column FK that referenced
+`objects.id` to `store_objects.object_id`, preserving its `ON UPDATE`, `ON DELETE`, and deferral
```
RasmusTho, line 213: **[P2] Keep the owner schema consumer rows consistent with this cutover**

This new current-reality paragraph says every inventoried live FK is retargeted to `store_objects.object_id`, but the table-by-table rows immediately below still describe the post-migration consumers as `objects.id`: `chunks.object_id` (line 218), `embeddings.object_id` (line 240), and both `membership.object_id` / `membership.set_id` (lines 277-278). Those are four of the constraints this migration actually moves, so the owner doc currently presents both the old and shipped FK targets at once. Please update those rows to `store_objects.object_id` (while retaining the legacy-lineage qualifier for their origin) so operators and later migrations do not read the pre-cutover topology as current.

RasmusTho: Fixed in `b7968ab4`: all four current-reality FK rows now name `store_objects.object_id` after #3510, matching the cutover section. `DOCS_GUARD_ALLOW_TEMPORAL_SKIP=1 python3 scripts/docs_guard.py` and `git diff --check` pass. Fresh independent review rounds are pending.

tests/migrations/test_legacy_objects_fk_migration.py
```diff
@@ -0,0 +1,462 @@
...
+
+from __future__ import annotations
+
+import uuid
+from pathlib import Path
+
+import psycopg
+import pytest
+
+pytestmark = pytest.mark.pg
```
RasmusTho, line 11: **[P1] Attach a green PostgreSQL execution to the current head for the database-backed ACs.**

This module-level `pytestmark = pytest.mark.pg` means the AC3 Verify target here is excluded from the exact-head PR CI: the only full PR test check runs `pytest -m "not pg ..."`; AC2 in `tests/ingest/test_vault_root_ingest_pg.py` is excluded for the same reason. My exact-head targeted run produced `1 passed, 12 skipped` because no local PostgreSQL is configured. The closure contract does not allow a behavioral `Verify:` target that is skipped or excluded from the CI suite that ran to satisfy its AC. Please attach a green isolated-PostgreSQL execution of AC2 and AC3 (and preserve its exact head SHA) before merge.

RasmusTho: Current-head PG evidence attached for `b7968ab4`: started isolated throwaway pgvector/PostgreSQL and ran the three declared #3510 Verify targets exactly: `test_all_legacy_objects_fk_consumers_are_accounted_for`, `test_legacy_objects_fk_migration_fails_loudly_on_unsupported_state`, and `test_ingest_persists_decisions_without_legacy_objects_parent_after_migration` — **3 passed in 5.46s**. The full related package also completed before this focused confirmation. This satisfies AC1–AC3 on the current head; no migration/runtime code changed after the preceding clean review.

docs/DB_SCHEMA.md
```diff
@@ -259,8 +274,8 @@ Interpretation:
 
 ### `membership` (legacy)
 The legacy baseline retains the **composite** key form:
-- `object_id` (`uuid`, FK → `objects.id`, `ON DELETE CASCADE`)
-- `set_id` (`uuid`, FK → `objects.id`, `ON DELETE CASCADE`) (sets are stored as objects in this baseline)
+- `object_id` (`uuid`, FK → `store_objects.object_id`, `ON DELETE CASCADE` after #3510)
+- `set_id` (`uuid`, FK → `store_objects.object_id`, `ON DELETE CASCADE` after #3510; sets are stored as objects in this baseline)
```
RasmusTho: **[P2] Keep the conditional `membership.set_id` target truthful**

This row now claims that `membership.set_id` points to `store_objects.object_id` after #3510, but the migration explicitly preserves two supported shapes: on fresh databases `set_id` references `sets(id)`, while only retained objects-as-sets schemas have an `objects(id)` FK that this migration retargets (`7e4f2a1c9d30...py:23-25`). The fresh Alembic lineage creates `membership.set_id REFERENCES sets(id)` in `202510241200_sot41_amg_core.py`, and the later `CREATE TABLE IF NOT EXISTS` cannot replace it. Please document the target conditionally (fresh: `sets.id`; retained objects-as-sets: `store_objects.object_id` after #3510) so the current-reality schema does not direct operators or future migrations to the wrong parent.

RasmusTho: Fixed in f4a62751. The owner schema now distinguishes fresh lineage (`sets.id`) from retained objects-as-sets lineage retargeted by #3510 (`store_objects.object_id`). Docs guard and diff check pass after rebasing onto current main.

RasmusTho:
Review repair published for the runtime cross-key alias finding. `resolve_vault_uuid_with_connection` now fails loud when a retained `objects.uuid` already names a different canonical `store_objects.object_id`, with PostgreSQL regression coverage for a collision introduced after cutover. Exact head: `7a04244b1e61c8a60ff9abb12b33bff7df1319d9`. Verification: focused ruff passed; `ruff check app tests` passed; `mypy app` passed (789 files). The two focused PostgreSQL tests were collected but skipped locally because no test DSN was configured; CI remains authoritative. Clean isolated review count restarts at zero.

app/workers/outbox_worker.py
```diff
@@ -1497,15 +1502,20 @@ def handle_panel_scan_requested(
     except (AttributeError, TypeError, ValueError) as exc:
         raise InvalidPanelNoteUUIDDispatchError(note_uuid) from exc
 
+    # The scan event carries the retained vault UUID. Historical rows may map
+    # it to a different canonical objects.id; all ObjectStore-facing refresh,
+    # intent, and writeback work must use that canonical key.
+    canonical_note_id = resolve_canonical_object_id(note_uuid)
```
RasmusTho, line 1508: **[P1] Companion-note hash refresh silently no-ops for retained legacy notes (canonical id substituted for vault uuid).**

`resolve_canonical_object_id(note_uuid)` threads the canonical `store_objects.object_id` into `refresh_panel_note_object` and `run_panel_note_execution`; the original vault `note_uuid` is never passed downstream. It propagates `DomainObject.uuid` → `NoteRef.uuid` (no separate vault-uuid field) → `_refresh_companion_hash(note_uuid)` → `read_companion(vault_root, note_uuid)` in `app/agents/panel_agent/runtime.py`. Companion files are keyed by the vault-native frontmatter uuid (`app/services/companion_note.py::companion_path`, and `app/ingest/vault_alpha.py` writes them by `frontmatter_uuid`).

**Failure:** for a retained note where `objects.uuid != objects.id` (exactly the #3510 cutover scenario), panel writeback updates the real note file (path-based resolution still works) but `read_companion(vault_root, canonical_id)` looks for `<canonical_id>.md`, which does not exist, returns `None`, and `_refresh_companion_hash` silently no-ops — the companion `content_hash` stays stale and later staleness/conflict checks produce false mismatches. The new test `tests/runtime/test_worker_ingest_event.py::test_handle_panel_scan_requested_uses_canonical_store_identity` locks in the substitution but monkeypatches away the downstream functions, so the companion consequence is untested.

**Fix direction:** thread the vault uuid alongside the canonical id (e.g. keep the original event uuid for companion/vault-file identity, use the canonical id only for store lookups), or give `NoteRef` an explicit vault-identity field; add a retained-note regression test for `_refresh_companion_hash`.

app/stores/pg.py
```diff
@@ -322,18 +340,114 @@ def truncate_pg_tables() -> None:
...
-            cur.execute("TRUNCATE TABLE store_relation_memberships")
-            cur.execute("TRUNCATE TABLE vector_index_meta")
+            # Do not CASCADE a truncate through durable consumers (decisions,
+            # audit, etc.).  Ordered DELETE keeps their declared FK semantics
+            # (for example SET NULL) and leaves the reset atomic/fail-loud.
+            cur.execute("DELETE FROM store_vector_index")
+            cur.execute("DELETE FROM store_relation_memberships")
+            cur.execute("DELETE FROM store_relations")
+            cur.execute("DELETE FROM vector_index_meta")
+            cur.execute("DELETE FROM store_objects")
```
RasmusTho, line 365: **[P2] `DELETE FROM store_objects` now cascade-deletes chunks/embeddings/relations/membership — contradicting this function's own no-cascade intent.**

Migration `7e4f2a1c9d30` repoints those four tables' FKs onto `store_objects(object_id)` preserving their original `ON DELETE CASCADE` (from `202510241200_sot41_amg_core.py`), while `decisions`/`audit` are `SET NULL` (via `1a739d9494af`). So this final DELETE silently empties all four tables even though none are named here. Verified empirically against a live Postgres with the post-migration schema: chunks/embeddings/relations/membership went 1→0 rows; audit survived with `object_id` NULL.

`tests/stores/test_pg_truncate_reset.py` mocks the cursor and asserts only the SQL statement list, so the cascade is structurally unobservable there. Callers: ~150+ test fixtures via `reset_store_backends()` plus `ops/benchmarks/run_benchmark.py` (no production path today).

**Fix direction:** either delete the four cascade children explicitly (making the comment true and the blast radius visible), or update the comment/docs to state that store reset intentionally cascades through canonical-FK children post-#3510 — and add one pg-marked test observing real post-reset row counts.

app/alembic/versions/7e4f2a1c9d30_migrate_legacy_objects_fk_consumers.py
```diff
@@ -0,0 +1,291 @@
...
+revision: str = "7e4f2a1c9d30"
+down_revision: Union[str, None] = "4d1e0c9a3329"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+reversibility: str = "forward-only"
+
+# Every schema shape observed in the supported migration lineage.  ``set_id``
+# points at ``sets`` on fresh databases, but some retained v4.5 databases used
+# objects-as-sets; accepting it here keeps that live historical shape explicit.
+LEGACY_OBJECTS_FK_CONSUMERS = frozenset(
```
RasmusTho, line 26: **[P2] The FK-consumer inventory exists in three unlinked copies, and the AC1 inventory test cannot detect drift in either direction.**

`LEGACY_OBJECTS_FK_CONSUMERS` (this frozenset) is referenced nowhere — repo-wide grep finds only its definition. The enforced copy is the hardcoded SQL tuple literal in the `NOT IN (...)` allowlist inside the `DO $$` block below. The AC1 Verify target `tests/migrations/test_legacy_objects_fk_inventory.py` adds a third copy (`ACCOUNTED_CONSUMERS`) and asserts per-pair substring presence over the migration text — satisfied by the SQL literal alone — plus a `discovered <= ACCOUNTED_CONSUMERS` check that only scans OTHER migration files. Consequences in the required (not-pg) CI gate: (1) editing the frozenset changes nothing and stays green; (2) appending an unreviewed pair directly to the SQL `NOT IN` list widens the migration's acceptance and also stays green. The tests that actually execute the enforcement are all `pytest.mark.pg` (dispatch-only).

**Fix direction:** render the SQL tuple list from `LEGACY_OBJECTS_FK_CONSUMERS` via `format()` inside `op.execute` so one Python source governs enforcement, and make the inventory test import that frozenset and assert exact equality with a value parsed from the SQL — killing both drift directions in the required gate.

app/services/decisions.py
```diff
@@ -21,6 +21,62 @@
 _ALLOWED_BACKENDS = {"memory", "pg"}
 
 
+class DecisionsSchemaMigrationRequired(RuntimeError):
+    """The active decisions projection still has its pre-#3510 FK parent."""
+
+
+def _assert_decisions_fk_cutover() -> None:
```
RasmusTho, line 29: **[P2] `_assert_decisions_fk_cutover` is a hand-copied, narrower fork of `app.db.decisions_schema.assert_decisions_schema` — and the drift it invites already happened once inside this PR.**

The shared helper (used by `app/stores/postgres.py` and `app/jobs/decisions_projection.py`) also asserts table existence, required columns, nullability, `gen_random_uuid` default, and exact FK cardinality; this copy checks only the FK rows and raises a different exception type. The FK-cardinality tightening was added to the shared helper 17 minutes after this copy was extracted and was never mirrored here — the exact split-brain this PR's own seam extraction was meant to end. `assert_decisions_schema` only issues SELECTs, so it runs fine on the existing `conn_ro()` connection; no circular-import risk (`app/db/decisions_schema.py` has no internal imports).

**Fix direction:** call `assert_decisions_schema(conn)` from `insert_decision` (optionally moving `DecisionsSchemaMigrationRequired` into the shared module and raising it there; it subclasses RuntimeError so existing callers are unaffected), and delete this copy.

app/stores/pg.py
```diff
@@ -322,18 +340,114 @@ def truncate_pg_tables() -> None:
...
+        )
+    if canonical_alias_exists and str(object_id) != str(vault_uuid):
+        raise RuntimeError(
+            "retained vault UUID already names a different canonical object; reconcile the "
+            "cross-key identity collision before retrying"
+        )
+    return str(object_id or vault_uuid)
+
+
+def resolve_vault_uuid(vault_uuid: str) -> str:
```
RasmusTho, line 402: **[P2/perf] `resolve_vault_uuid` opens a brand-new Postgres connection per call — and it is called per note per watcher tick and per drained outbox event.**

Confirmed live call sites: `app/watcher/vault_watcher.py:647` (inside the per-note panel loop each tick), `app/watcher/registry.py:386` (`_process_panel_note`, per note), `app/workers/outbox_worker.py:1508` (per `PANEL_SCAN_REQUESTED` event in the daemon loop). Each call pays TCP+auth connection setup for an identity mapping that is immutable for a note's lifetime.

**Fix direction:** accept an existing connection at the call sites that already hold one, and/or memoize uuid→canonical-id per process (the mapping only changes at migration time; the deploy contract restarts processes after `alembic upgrade head` — compose gates services on the migrate container completing).

app/services/decisions.py
```diff
@@ -99,6 +155,12 @@ def insert_decision(object_id: str, key: str, value: dict[str, Any], trace_id: s
     # Durable path — decision-receipt log is CANONICAL, Postgres is the
     # projection (feat #2969, docs/DECISION_RECEIPT_LOG/README.md).
     #
+    # The read-only #3510 schema preflight must run before the canonical log is
+    # mutated. Otherwise a pre-cutover FK can accept the receipt and then reject
+    # its projection with a raw ForeignKeyViolation, leaving an avoidable
+    # durable-but-unprojected decision.
+    _assert_decisions_fk_cutover()
```
RasmusTho, line 136: **[P3/perf] Per-write schema preflight: every `insert_decision` opens a fresh `conn_ro()` and runs a 4-table `information_schema` join, unmemoized.**

`insert_decision` is the live write path for the classifier/reviewer/set_evaluator agents and loops inside `BackfillJob._process_reviews/_process_evaluations`. The schema fact it re-verifies cannot change mid-process under the deploy contract (compose gates watcher/worker/api on the migrate service completing; promotion restarts after migrating). The sibling backend check in this same file is already `@lru_cache`; the analogous store-schema check in `app/stores/pg.py` uses a process-lifetime `_TABLES_READY` flag.

**Fix direction:** cache first success per process (module flag or `lru_cache`), preserving the fail-loud first call. Same applies to `PgDecisions.put`'s per-call `assert_decisions_schema` (currently no live production caller, but the same pattern).

app/jobs/decisions_projection.py
```diff
@@ -62,18 +63,24 @@ def _parse_ts(raw: Any) -> datetime | None:
 
 
 def _existing_object_ids(cur) -> set[str]:
-    cur.execute("SELECT id FROM objects")
-    return {str(r["id"] if isinstance(r, dict) else r[0]) for r in cur.fetchall()}
+    cur.execute("SELECT object_id FROM store_objects")
+    return {str(r["object_id"] if isinstance(r, dict) else r[0]) for r in cur.fetchall()}
 
 
 def _uuid_to_id_map(cur) -> dict[str, str]:
```
RasmusTho: **[P3/reuse] The `store_objects ↔ objects` identity join is now copy-pasted in two files added by this PR.**

`_uuid_to_id_map` here and `app/receipts/decision_receipt_log.py::resolve_vault_uuid` both run `FROM store_objects canonical LEFT JOIN objects legacy ON legacy.id = canonical.object_id` (previously two different simpler queries). Directionality and null-handling differ intentionally (bulk COALESCE-fallback map vs single lookup returning None), but the join predicate — the part that encodes the retained-identity model — is duplicated, so a future identity-model change can update one and miss the other, making projection rebuild re-linking disagree with receipt-time vault_uuid resolution.

**Fix direction:** factor the join into one shared helper (parameterizing the null-handling), e.g. next to `resolve_vault_uuid_with_connection` in `app/stores/pg.py`.

tests/integration/test_vault_sync_atomicity.py
```diff
@@ -71,6 +71,19 @@ def _configure_isolated_pg_test(monkeypatch) -> tuple[str, str]:
     schema = f"pgtest_{uuid4().hex}"
     _create_schema(base_dsn, schema)
     dsn = _dsn_with_search_path(base_dsn, schema)
+    with psycopg.connect(dsn, autocommit=True) as conn:
+        conn.execute(
+            """
+            CREATE TABLE store_objects (
```
RasmusTho: **[P3/reuse] Hand-copied `CREATE TABLE store_objects` DDL will drift from the migration-owned schema.**

`_configure_isolated_pg_test` re-types the exact 6-column DDL from `app/stores/pg.py::_ensure_tables`. Two existing reusable mechanisms: (a) `_ensure_tables()` itself — `tests/conftest.py`'s autouse `store_schema_autocreate_for_pg_tests` already sets `STORE_SCHEMA_AUTOCREATE=1` for every pg-marked test; (b) `tests/migrations/decisions_schema_helpers.py::migrated_decisions_db` (real `alembic upgrade head`), already used by three other test files. A future store_objects column change leaves this copy stale, so the atomicity tests keep passing against a shape production never has.

app/services/vault_sync.py
```diff
@@ -549,6 +795,24 @@ def sync_markdown(path: str) -> dict[str, Any]:
...
+                    conn,
+                    object_id=canonical_object_id,
+                    updates=json.loads(payload_json),
+                ),
+                source_ref=str(note_path),
+                created_at=mtime,
+            ),
+        )
+
+        obj_payload["uuid"] = canonical_object_id
```
RasmusTho: **[P3/altitude] The `{uuid: canonical, object_id: canonical, vault_uuid: retained}` event-identity trio is inline-built at four sites with no shared constructor.**

Sites at this head: `delete_note` (~458–460), `upsert_object_from_note` (~577–579), `sync_markdown` (here), and `app/watcher/vault_watcher.py:380–388` — whose docstring says it is “shaped exactly like vault_sync.delete_note’s”, a mirror enforced only by comment. A fifth producer (or an edit to one of the four) that forgets a key or maps `uuid` to the retained value re-creates the split-parent defect this PR exists to fix, with no signal at the write site.

**Fix direction:** one `canonical_event_identity(canonical_id, vault_uuid) -> dict` helper used by all four producers.

app/services/vault_sync.py
```diff
@@ -106,8 +243,17 @@ def _upsert_file_state(
...
+    *,
+    old_path: str,
+    new_path: str,
+    uuid_value: str,
+    payload: dict[str, Any],
+    fm_hash: str,
+    body_hash: str,
+    mtime: datetime,
 ) -> None:
+    canonical_object_id = _canonical_object_id(conn, uuid_value)
```
RasmusTho: **[P3] Rename path resolves the canonical id twice in one transaction — and note: `sync_markdown`/`handle_rename` currently have zero non-test callers.**

`sync_markdown` resolves `canonical_object_id` up front, then the pure-rename branch calls `_update_path_only`, which re-resolves the same uuid on the same connection instead of receiving the value. Wasted round-trip per rename; under READ COMMITTED the two statements use different snapshots, so `result["id"]` can in principle disagree with the id actually written.

Also worth an explicit decision: repo-wide search finds no production caller of `sync_markdown`/`handle_rename` — the live registry watcher routes renames through `vault_alpha` re-ingest, and the legacy `scripts/fs_watcher.py` uses `update_path()`/`upsert_object_from_note` via `FilesystemVaultAdapter`. The atomicity tests label `sync_markdown` “the real production entrypoint”, but nothing wires it in. Either wire it, or document it as the adapter-contract surface it currently is.

app/workers/outbox_worker.py
RasmusTho: Fixed at exact head `6dd8e8c8`. The canonical ObjectStore id remains the panel cognition/write id, while the retained vault UUID is now threaded separately through `run_panel_note_execution` and `execute_panel_intent` to `_refresh_companion_hash`. Both the worker boundary and the full execution pipeline have regressions for the distinct identities.

app/stores/pg.py
RasmusTho: Fixed at exact head `6dd8e8c8`. `truncate_pg_tables` now explicitly deletes the four canonical-FK CASCADE consumers (`chunks`, `embeddings`, `relations`, `membership`) before deleting `store_objects`; decisions/audit retain their SET NULL semantics. The reset test asserts the complete visible scope.

app/alembic/versions/7e4f2a1c9d30_migrate_legacy_objects_fk_consumers.py
RasmusTho: Fixed at exact head `6dd8e8c8`. `LEGACY_OBJECTS_FK_CONSUMERS` now renders the SQL allowlist used by the migration, and the required not-pg inventory test imports the migration and proves exact equality between the governed frozenset and rendered enforcement list.

app/services/decisions.py
RasmusTho: Fixed at exact head `6dd8e8c8`. The active writer now delegates to `app.db.decisions_schema.assert_decisions_schema`; the copied narrower query is removed and the caller preserves its domain-specific migration exception.

app/services/decisions.py
RasmusTho: Fixed at exact head `6dd8e8c8`. The shared schema preflight caches only successful validation, keyed by resolved DSN, so each process/database fails loud on first use and avoids repeated information_schema scans thereafter.

app/workers/outbox_worker.py
RasmusTho: Disposition: **fixed** in `29b9dc92` — `_refresh_companion_hash` now resolves the companion from the note's own frontmatter uuid (fallback: caller id), with a retained-note regression test proven red pre-fix (`tests/agents/panel_agent/test_companion_hash_identity.py`).

app/stores/pg.py
RasmusTho: Disposition: **fixed** in `b6ab5782` — cascade children are deleted explicitly (to_regclass-guarded), the comment states the retained SET NULL semantics truthfully, and a pg-marked live test observes both effects on a migrated database.

app/alembic/versions/7e4f2a1c9d30_migrate_legacy_objects_fk_consumers.py
RasmusTho: Disposition: **fixed** in `74d6bf2e` — the SQL NOT IN allowlist is rendered from `LEGACY_OBJECTS_FK_CONSUMERS` via a sentinel, and the inventory test now imports the module, asserts the sentinel wiring and SQL round-trip, and bases the historical checks on the imported set (third copy removed). Rendered migration executed live against a fresh database.

app/services/decisions.py
RasmusTho: Disposition: **fixed** in `87648b9b` — `_assert_decisions_fk_cutover` delegates to the shared `assert_decisions_schema` on the same read-only connection, preserving `DecisionsSchemaMigrationRequired` and the #3510 guidance while inheriting the full shape contract.

app/stores/pg.py
RasmusTho: Disposition: **partially fixed** in `c28c1f02` — retained (non-identity) mappings are cached process-lifetime, eliminating reconnects for exactly the retained-note traffic this finding targets; identity results stay uncached by design. Remaining identity-path connection reuse: follow-up #3987 (item 4).

app/services/decisions.py
RasmusTho: Disposition: **deferred with reason** to #3987 (item 4) — memoizing the per-write preflight is safe under the deploy contract but must not weaken the pg tests that intentionally mutate schema states mid-process; the per-call fail-loud check is retained for now.

app/jobs/decisions_projection.py
RasmusTho: Disposition: **deferred** to #3987 (item 1) — one shared identity-join helper with parameterized null-handling.

tests/integration/test_vault_sync_atomicity.py
RasmusTho: Disposition: **deferred** to #3987 (item 2) — replace the hand-typed DDL with an owned bootstrap (`_ensure_tables` via STORE_SCHEMA_AUTOCREATE or the alembic fixture).

app/services/vault_sync.py
RasmusTho: Disposition: **deferred** to #3987 (item 3) — one `canonical_event_identity` constructor for the four producer sites.

app/services/vault_sync.py
RasmusTho: Disposition: double-resolve **fixed** in `d520aee9` (sync_markdown threads its resolved id into `_update_path_only`); the sync_markdown/handle_rename no-live-caller decision is **deferred** to #3987 (item 5).

app/services/vault_sync.py
```diff
@@ -429,34 +647,56 @@ def sync_markdown(path: str) -> dict[str, Any]:
                 fm_hash=fm_hash,
                 body_hash=body_hash,
                 mtime=mtime,
+                canonical_object_id=canonical_object_id,
```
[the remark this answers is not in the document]
RasmusTho: Fixed in prepared direct-child candidate `5d15a4e58aa3b01f566cee92f7527984b261b654` (parent `ec3b32e4`, which already combines live `6dd8e8c8` with the required deploy/migration repairs). The unsupported argument is removed. Direct regression proves an active edit returns `deferred`, persists the file-state baseline, and creates neither parent row; `tests/services/test_vault_sync_lifecycle.py` is 11/11 green and Ruff passes. The active #3906 claim-holder has the exact SHA/tree; thread remains unresolved until that candidate is the live head and exact-head evidence is rerun.

PR #3732, opened 2026-07-15
Add governed conversational journal draft staging

RasmusTho:
Governing-Issue: #3349

## Summary

- stage one transcript-grounded journal proposal per day under the vault system `drafts/journal` area
- preserve separate conversation and system-context citations, fail loudly on unresolved sources, and never write the canonical daily entry
- require every source to independently reach cited-proposal admission: draft/provisional stays raw, revised stays read-only, and rejected/unknown is excluded before cognition or staging
- harden staging against symlink escapes and concurrent same-day lost updates with no-follow traversal, atomic replacement, and in-process plus cross-process locking
- consume the reasoning runtime's explicit outcome/degradation contract without inferring success from counts or fabricating synthesis
- preserve conservative source posture and durable content-free activation receipts across restart

Fixes #3349

## SBS impact

- System: Product/Runtime
- Primary charter: Human Knowledge Authority
- Secondary charters: Context Assembly & Orchestration; Governance
- Artifact class: candidate/proposal only
- Authority effect: none; acceptance remains owned by JRNL-04
- Storage/retrieval: vault-local staging only; staged drafts remain excluded from production ingest
- Contract boundaries: no Create output-kind expansion and no canonical journal mutation

## Validation

- trust-posture production regression: reproduced rejected/revised/queued capture laundering 3/3 before repair; now all three block before cognition/staging with canonical reasons
- proposal-builder backstop: rejected/revised/unknown all fail loud
- explicit reasoning integration regressions: provider failure, empty output, and mixed degraded cognition preserve explicit posture and fabricate no synthesis
- `134 passed`: journal concurrency/security/retrieval/cognition/trust/receipt/store-census plus shared reasoning/provider and Create lifecycle suites
- fresh-vault spawn multiprocessing barrier regression: passed 10/10 consecutive runs while retaining both sessions, owner text/provenance, and both durable receipts
- pre-dependency head `80bddb155fcbdf274fd5fa49caac8468631042db`: full non-pg suite `8398 passed, 114 skipped, 85 deselected, 18 xfailed`
- current-head full non-pg was not started locally because disk preflight found only about 420 MiB free; exact-head GitHub CI non-pg is required below
- `ruff check app tests companion-ui/companion-app`
- `mypy app` (740 files)
- `lint-imports --config importlinter.ini` (1 contract kept)
- `python -m app.cli settings-validate --json`
- `python3 scripts/docs_guard.py --check` (`Docs guard: OK`)
- `git diff --check`

## Coordination

Dispatcher unavailable because the worktree database was uninitialized; used the governed GitHub-label-only resume claim fallback. Current pickup receipt: https://github.com/RasmusTho/agentic-pkm-mvp/issues/3349#issuecomment-4976198805

## BuilderOps routing

- Records/projections/receipts: `none`
- Reason: bounded dependency integration and trust-posture repair are captured in repo truth and PR review threads; no upstream workflow divergence warrants a LearningSignal.

## Owner-document writeback

- `docs/FRONTMATTER.md` — machine-synthesis provenance keys and the JRNL-03 proposal posture
- `docs/CONVERSATIONAL_JOURNALING/DRAFT_JOURNAL_ENTRY.md` — source admission, hardened staging, explicit reasoning outcome/degradation, and durable-receipt behavior
- `docs/STATUS.md` — shipped Conversational Journaling status with explicit trust boundaries


app/journaling/draft.py
```diff
@@ -0,0 +1,517 @@
...
+
+def _draft_path(vault_root: Path, for_date: date, *, is_addendum: bool) -> Path:
+    system_dir = Path(get_vault_system_dir_rel(vault_root))
+    suffix = "-addendum" if is_addendum else ""
+    path = (
+        vault_root
+        / system_dir
+        / JOURNAL_DRAFTS_SUBDIR
+        / f"{for_date.isoformat()}{suffix}.md"
+    ).resolve()
```
RasmusTho: [P1] Reject symlinked staging components instead of resolving through them. `_draft_path()` follows the configured staging tree with `resolve()` and only checks that the final target remains somewhere inside the vault. A symlink `_system/drafts/journal -> Inbox` therefore writes `Inbox/YYYY-MM-DD.md`; the real ingest candidate selector then indexes it, violating both the fixed staging location and the anti-laundering retrieval invariant. I reproduced this at this exact SHA: `result.path == Inbox/2026-07-15.md` and `_select_candidates(...)` contained that path. Validate every staging path component with no-follow semantics (and the final target collision) before the guarded mutation, and add a production-selector regression.

app/journaling/draft.py
```diff
@@ -0,0 +1,517 @@
...
+    # Re-resolve immediately before the guarded mutation so a source removed
+    # during cognition cannot be laundered into the staged proposal.
+    _validate_session_citations(vault_root, sessions)
+    _validate_context_citations(vault_root, context_items)
+
+    # Production mutation seam: the guard is immediately before the first
+    # possible filesystem mutation (directory creation).  The same-directory
+    # staged file plus os.replace makes both create and redraft atomic.
+    write_guard.assert_writes_allowed(JOURNAL_DRAFT_WRITE_ACTION)
+    _atomic_write(draft_path, note_text)
```
RasmusTho: [P1] The read/compose/replace sequence is not collision-safe, so concurrent same-day redrafts lose a contributing session. Two calls can both read the same prior frontmatter, independently compose, and atomically replace the same file; atomic replacement prevents torn bytes but not lost updates. With a barrier in `reasoning_fn`, two calls for sessions `a` and `b` both returned the same path while the final file contained only `session:a`. The AC says repeated same-day composition extends/redrafts the one candidate; add a per-draft lock or CAS/retry around read+compose+replace and a concurrent production-path test.

app/journaling/draft.py
```diff
@@ -0,0 +1,517 @@
...
+            f"Journal addendum candidate {for_date.isoformat()}"
+            if is_addendum
+            else f"Journal draft {for_date.isoformat()}"
+        ),
+    )
+    existing_uuid = str(existing_frontmatter.get("uuid") or "").strip()
+    if existing_uuid:
+        compilation = compilation.model_copy(update={"artifact_id": existing_uuid})
+
+    cognition = _run_cognition(reasoning_fn, source_ids, activation.receipt.receipt_id)
```
RasmusTho: [P1] The claimed cognition integration cannot consume JRNL-02/JRNL-01 identifiers. `run_multi_note_reasoning` loads UUID-keyed objects from the object store, while this call supplies `session:<uuid>` and vault-relative paths/fragments; those always resolve to zero objects. The function returns an empty `ReasoningOutput` without raising, so `_run_cognition` records `degraded: false`, and the stored body is only deterministic collation, not cognition-backed ghost-writing. This is also the real #3350 compatibility shape. Adapt the transcript/context into the cognition input contract (or use a synthesis seam that accepts their text), consume the synthesis result, and test the default production function rather than only allowing an injected fake.

app/journaling/draft.py
```diff
@@ -0,0 +1,517 @@
...
+            role="conversation",
+            review_state="reviewed",
+        )
+        for session in sessions
+    ) + tuple(
+        SourceRef(
+            artifact_id=item.provenance_ref,
+            note_path=_reference_path(item.provenance_ref),
+            role="system_context",
+            review_state="reviewed",
```
RasmusTho: [P1] Do not manufacture `reviewed` provenance for every day-context item. JRNL-01 deliberately includes Knowledge Acquisition captures whose shipped frontmatter is `review_state: draft` / `requires_review: true`; converting those to `SourceRef(review_state="reviewed")` makes `proposal_builders._reject_hidden_authority` accept a posture the source never acquired. The activation adapter repeats the same unconditional REVIEWED claim. Preserve/derive the actual trust and review posture (or explicitly model cited-proposal admissibility without asserting review), then test a real draft capture through both activation and CompilationDraft construction.

app/journaling/draft.py
```diff
@@ -0,0 +1,517 @@
...
+        "journal_candidate_type": "addendum" if is_addendum else "primary",
+        "for_date": for_date.isoformat(),
+        "derived_by": "conversation",
+        "authority_state": "proposal",
+        "proposed_by": {
+            "capability": JOURNAL_DRAFT_CAPABILITY_ID,
+            "cognition": cognition,
+        },
+        "sources": list(source_ids),
+        "activation_receipt_id": activation.receipt.receipt_id,
```
RasmusTho: [P1] `activation_receipt_id` currently points to an in-memory `GateDecisionReceipt` that is discarded. Unlike the established Expansion gate path (`emit_expansion_gate_receipt`) and ASK activation receipt, JRNL-03 emits no durable record, so the staged file contains an unresolvable receipt reference and cannot prove which candidates/outcome produced activation. Persist the content-free gate receipt through the governed/atomic staging transaction (or a contractually equivalent durable seam), verify id resolution after restart, and cover failure atomicity so receipt and draft cannot diverge.

docs/FRONTMATTER.md
```diff
@@ -124,6 +125,13 @@ write path (`app/expansion/create.py`); `authority_state` (=`accepted`), `accept
 `acceptance_receipt_id`, and `decision_token_ref` are written by EXP-4's acceptance path
 (`app/expansion/accept.py`) — the only path from a staged draft to a canonical note.
 
+Conversational Journaling's JRNL-03 staging path (`app/journaling/draft.py`) uses the same proposal
+posture without extending Create's closed output-kind enum: it writes `derived_by: conversation`,
+`authority_state: proposal`, and a `sources` list containing the reflection-session id plus every
+day-context provenance reference folded into the draft. These fields do not make the draft human
```
RasmusTho, line 131: [P1] The recorded validation is not reproducible at this head: bare `python3 scripts/docs_guard.py --check` exits 1 because runtime code changed without any high-risk temporal owner doc (`STATUS`, `ARCHITECTURE`, etc.). The PR body currently claims that exact command passed. Either add the required shipped-current-state writeback (while keeping parent acceptance truth honest) or use and document a contractually justified temporal-skip path; then rerun the exact command/evidence. FRONTMATTER alone does not satisfy the repository gate.

RasmusTho (commented):
Fresh independent local review gate at exact SHA `558da397bb1617917610a5a3ecd609ed57366bdc`: **reject / merge blocked**. Exact-head CI is green on the latest runs and the focused suite passed locally (22 tests), but six unresolved inline findings remain: staging symlink escape into retrieval-visible vault content; lost-update collision for concurrent same-day redrafts; default cognition integration is structurally empty while reporting non-degraded; false `reviewed` posture on draft capture sources; orphaned/non-durable activation receipt reference; and the claimed bare docs-guard command actually exits 1. No merge authority was exercised.

```yaml
tcd_review:
  verdict: reject
  risk_level: high
  model_used: gpt-5.6-sol
  reasoning_effort_used: medium
  under_modeling_detected: false
  over_modeling_detected: false
  blocking_issues: [symlink staging/retrieval escape, concurrent lost update, nonfunctional cognition seam, provenance-review laundering, non-durable activation receipt, failing docs guard]
  non_blocking_issues: []
  missing_tests: [symlinked staging tree, concurrent same-day collision, default #3350 transcript-to-cognition path, draft-capture trust posture, restart-resolvable activation receipt]
  hidden_defect_risks: [candidate authority laundering, candidate indexed as knowledge, orphan receipt ids]
  recommended_fixes: repair all six mechanisms and rerun a fresh independent exact-SHA review
  recommended_model_for_fix: Sol
  recommended_reasoning_for_fix: high
  residual_risk: high until repair and re-review
```

app/journaling/draft.py
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: staging now walks directory components by dir-fd with O_NOFOLLOW, rejects symlink ancestors and symlink/non-regular final targets, and performs same-directory atomic replacement. Regressions cover an ancestor symlink into Inbox and a final-target symlink.

docs/FRONTMATTER.md
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: docs/STATUS.md now records the shipped Conversational Journaling surface, and bare python3 scripts/docs_guard.py --check passes with .

app/journaling/draft.py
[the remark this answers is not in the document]
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: draft capture now preserves conservative source posture; chat sessions and draft captures default to unreviewed, and the dedicated cited-unreviewed proposal builder permits only explicitly unreviewed cited sources while rejecting stale, ranked, machine-derived, rejected/unknown, or action-authorizing inputs.

app/journaling/draft.py
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: the full read/compose/replace transaction is serialized by a per-draft in-process RLock plus fcntl.flock for cross-process writers. A barrier-based concurrent same-day regression verifies that both sessions survive.

app/journaling/draft.py
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: reasoning receives real resolvable object IDs via the existing allowlisted substrate; stored payloads carry the required unbound episode default, output is rendered only when claims/inferences exist, and no-op/unavailable cognition is truthfully degraded.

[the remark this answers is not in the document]
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: journal cognition now materializes stable UUID-addressable inputs through the existing reasoning store substrate, verifies they resolve, consumes actual claims/inferences into the machine-cognition section, and reports degraded when inputs or output are unavailable.

app/journaling/draft.py
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: source review state is derived and preserved instead of coerced to reviewed. Conversational and draft-capture inputs default to unreviewed, with a narrow cited-unreviewed builder that retains authority guards.

app/journaling/draft.py
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: the content-free activation receipt record is embedded in the atomically persisted draft frontmatter, prior same-day receipt records are retained on redraft, and resolve_journal_draft_activation_receipt resolves it after process restart.

docs/FRONTMATTER.md
RasmusTho: Fixed in ca14f0a31f8892345721e16f7ffa543c539c7699: docs/STATUS.md now records the shipped Conversational Journaling surface, and the bare docs guard passes with the terminal result Docs guard: OK.

app/journaling/draft.py
```diff
@@ -0,0 +1,809 @@
...
+                        "journal draft staging component is a symlink or not a directory"
+                    ) from exc
+                raise
+            if current_fd != root_fd:
+                os.close(current_fd)
+            current_fd = next_fd
+
+        lock_name = f".{relative_path.name}.lock"
+        try:
+            lock_fd = os.open(
```
RasmusTho: [P1] Make first-use cross-process locking reliable. The checked-in barrier test only uses threads, but two separate processes hitting a fresh vault race here while the staging tree/lock file is created: in 4 of 5 exact-head reproductions one process raised FileNotFoundError (Errno 2) for .2026-07-15.md.lock at this os.open, so one same-day session never reaches the serialized transaction. Pre-creating the system drafts/journal directory makes the same cross-process barrier retain both sessions, both text bodies, and both receipts, isolating the defect to first-use setup. Add a real multiprocessing/subprocess regression on a fresh vault and make directory/lock creation collision-safe so both callers complete.

RasmusTho: Fixed in 80bddb155fcbdf274fd5fa49caac8468631042db: added a real spawn multiprocessing barrier regression on a fresh vault; it reproduced ENOENT 5/5 before the fix and passed 10/10 after. First-use now atomically creates or opens a no-follow regular bootstrap lock anchored by the existing vault root dirfd, holds its flock through secure staging-tree creation and per-draft flock acquisition, then releases it before the serialized transaction. Both callers succeed and the final candidate retains both session IDs, owner text/provenance, and both durable receipts. Bootstrap and per-draft lock symlink attacks also fail closed.

app/journaling/draft.py
```diff
@@ -0,0 +1,883 @@
...
+            f"{record.get('object_id')}:{record.get('key')}:{record.get('created_at')}"
+        )
+        if candidate == fragment:
+            return
+    raise UnresolvableJournalCitationError(
+        f"citation fragment does not resolve: {reference}"
+    )
+
+
+def _review_state(frontmatter: Mapping[str, object]) -> ReviewState:
```
RasmusTho: [P1] Preserve rejected/revised source posture instead of converting it to admissible raw input. This normalizer maps every value other than reviewed/accepted/protected to UNREVIEWED, including canonical ReviewState.REJECTED and REVISED values. The shipped day-context capture reader does not filter on review state, so a capture marked review_state: rejected is included, becomes unreviewed here, passes the journal activation gate at cited-proposal tier, and is accepted by build_cited_unreviewed_compilation_draft. That reverses the activation gate's explicit rejected exclusion and contradicts this PR's claim that rejected/unknown postures fail loudly. I reproduced the normalization at this exact SHA: rejected -> unreviewed and queued -> unreviewed. Parse the full posture without weakening negative/unknown states, reject non-admissible inputs before materialization/staging, and add a production-path rejected-capture regression through activation and CompilationDraft construction.

RasmusTho: Fixed in 4730209d9815ce860ace41c95c40226aa83a8e7d: production parsing now preserves rejected and revised, maps only canonical draft/provisional/raw values to unreviewed, and keeps missing/unknown as unknown. Every source must independently reach cited-proposal admission before cognition or staging; rejected reports contradicted_or_rejected, revised remains read-only with revised_read_only, and unknown reports provenance_unverified. The exact production path is covered for rejected/revised/queued, plus the proposal-builder backstop rejects all three postures.

app/journaling/draft.py
```diff
@@ -0,0 +1,915 @@
...
+    context_items: tuple[DayContextItem, ...],
+) -> dict[str, ReviewState | None]:
+    states = {session.source_id: session.review_state for session in sessions}
+    for item in context_items:
+        path = vault_root / _reference_path(item.provenance_ref)
+        state: ReviewState | None = None
+        if path.suffix.lower() == ".md":
+            frontmatter, _body = load_frontmatter(path.read_text(encoding="utf-8"))
+            state = _review_state(frontmatter, default=None)
+        states[item.provenance_ref] = state
```
RasmusTho, line 607: [P1] Reject colliding source identities before trust evaluation. This mapping lets a later day-context item overwrite an earlier session review state when their external IDs collide, and the admitted-ID set below then treats both occurrences as admitted. I reproduced the production `draft_journal_entry` path with a rejected transcript whose `session_id` was `collision.md` and a distinct accepted context artifact whose `provenance_ref` was `session:collision.md`: cognition was called before the proposal-builder finally rejected the transcript. That violates the contract that every source independently reaches cited-proposal admission before materialization/cognition. Make source identity typed/collision-free (or reject duplicates) and evaluate each source occurrence with its own posture; add the mixed-source regression.

RasmusTho: Fixed in 8ee08d0e: the combined source-id sequence is now checked for collisions before review-state mapping, activation, cognition, or staging. Added the reproduced mixed session/context collision regression and asserted cognition remains untouched.

PR #3767, opened 2026-07-15
Filter provisional recall stopword matches

RasmusTho:
Fixes #3766

Governing-Issue: #3766

## Summary
- Filter common stopwords out of provisional-memory recall token scoring.
- Add focused regressions proving stopword-only overlap does not recall while meaningful terms still match.

## Validation
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q tests/agent_memory/test_provisional_memory_recall.py` — 9 passed, including decomposed-stopword and cross-case combining-mark regressions
- `ruff check app tests companion-ui/companion-app` — All checks passed
- `mypy app/agent_memory/provisional_recall.py` — Success: no issues found in 1 source file
- `git diff --check` — passed

## BuilderOps Routing
- Records/projections/receipts: `receipt_20260715052144_45f3fe86`
- Reason: closure-audit receipt records terminal review-thread supersession/discard classifications; delivery truth for this residual repair is carried by issue #3766, this PR, and the PR #3743 review-thread source anchor.

## Claim
- Issue pickup: `github-comment:4977122608`
- Coordination mode: GitHub-label-only fallback (`dispatcher_db_uninitialized`)

### DIFF
--- a/app/agent_memory/provisional_recall.py
+++ b/app/agent_memory/provisional_recall.py
@@ -52,6 +52,38 @@ DEFAULT_PROVISIONAL_RECALL_RECEIPTS_PATH = Path(
 )
 PROVISIONAL_RECALL_RECEIPT_EVENT = "agent_memory.provisional_recall.evaluated"
 _TOKEN_RE = re.compile(r"[A-Za-z0-9]+")
+_STOPWORDS = {
+    "a",
+    "an",
+    "and",
+    "are",
+    "as",
+    "at",
+    "att",
+    "av",
+    "det",
+    "en",
+    "ett",
+    "for",
+    "för",
+    "i",
+    "in",
+    "is",
+    "it",
+    "med",
+    "of",
+    "och",
+    "om",
+    "on",
+    "or",
+    "på",
+    "som",
+    "the",
+    "to",
+    "vad",
+    "which",
+    "with",
+}
 
 
 @dataclass(frozen=True)
@@ -412,7 +444,11 @@ def _tier_rank(tier: AdmissionTier) -> int:
 
 
 def _tokens(value: str) -> set[str]:
-    return {match.group(0).lower() for match in _TOKEN_RE.finditer(value)}
+    return {
+        token
+        for match in _TOKEN_RE.finditer(value)
+        if (token := match.group(0).lower()) not in _STOPWORDS
+    }
 
 
 def _score(record: ProvisionalMemoryRecord, query_tokens: set[str]) -> float:
--- a/app/agent_memory/provisional_recall.py
+++ b/app/agent_memory/provisional_recall.py
@@ -13,6 +13,7 @@ import json
 import os
 from pathlib import Path
 import re
+import unicodedata
 from uuid import UUID, uuid4
 
 from app.activation.gate import (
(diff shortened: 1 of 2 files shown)


app/agent_memory/provisional_recall.py
```diff
@@ -52,6 +52,38 @@
...
+    "in",
+    "is",
+    "it",
+    "med",
+    "of",
+    "och",
+    "om",
+    "on",
+    "or",
+    "på",
```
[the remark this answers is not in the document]
RasmusTho: Fixed in cf4fad95: tokenization now preserves Unicode letters and case-folds before stopword filtering. The governing AC regression now includes Swedish “för” and “på” and fails on the prior ASCII tokenizer; the full focused file passes 8/8.

app/agent_memory/provisional_recall.py
```diff
@@ -412,7 +444,11 @@ def _tier_rank(tier: AdmissionTier) -> int:
 
 
 def _tokens(value: str) -> set[str]:
-    return {match.group(0).lower() for match in _TOKEN_RE.finditer(value)}
+    return {
+        token
+        for match in _TOKEN_RE.finditer(value)
```
RasmusTho: Normalize Unicode before tokenization/filtering. `[^\W_]+` recognizes precomposed `för`/`på`, but Python's `\w` does not include combining marks. Canonically equivalent NFD text therefore splits into surviving fragments: `_tokens(unicodedata.normalize("NFD", "för på")) == {"fo", "r", "pa"}`. If both the query and provisional content use that representation, their stopword-only overlap still yields a positive score and admission, so the Unicode fix does not satisfy the AC for ordinary decomposed input. Normalize (for example to NFC) before matching/case-folding, and cover the decomposed form in the regression.

RasmusTho: Fixed in b53234f4: `_tokens` now NFC-normalizes input before Unicode matching and case-folding. The governing AC test now uses decomposed NFD `fo\u0308r pa\u030a`; it reproduced the bypass before the fix and the focused suite passes 8/8 after it.

app/agent_memory/provisional_recall.py
```diff
@@ -412,7 +445,11 @@ def _tier_rank(tier: AdmissionTier) -> int:
 
 
 def _tokens(value: str) -> set[str]:
-    return {match.group(0).lower() for match in _TOKEN_RE.finditer(value)}
+    return {
+        token
+        for match in _TOKEN_RE.finditer(unicodedata.normalize("NFC", value))
```
RasmusTho: **[P2] Case-fold before token boundary detection**

NFC-before-matching fixes the Swedish NFD stopword bypass, but matching before `casefold()` still makes canonically case-insensitive forms tokenize differently when folding introduces or absorbs combining marks. At this head, `_tokens("ǰ") == {"j\u030c"}` while `_tokens("J\u030c") == {"j"}`, even though `NFC(value.casefold())` is identical for both; a query/content pair using those meaningful forms gets zero overlap. Case-fold the full string and normalize the folded result before `_TOKEN_RE.finditer(...)`, then add a cross-case combining-mark regression.

RasmusTho: Fixed on current rebased head 540d01ca: `_tokens` now case-folds the complete input, NFC-normalizes the folded result, and only then detects token boundaries. `test_cross_case_combining_mark_tokens_match` reproduced the `ǰ` versus `J\u030c` mismatch before the fix; the focused suite now passes 9/9.

PR #3786, opened 2026-07-15
Add CKM public snapshot contract

RasmusTho:
Governing-Issue: #3776

Fixes #3776

## Summary

- add lifecycle-safe CKM public identities and schema-v5 state contracts
- version current assessment fingerprints over material public evidence, selected formulae, and exact watermarks
- authenticate the latest pre-v5 fingerprint with the exact old producer algorithm before normalizing it to v2
- preserve stale-but-representable old rows as stable legacy snapshots so the normal engine appends a fresh v2 assertion
- align migration latest-row selection exactly with runtime ordering (asserted_at, then SQLite rowid)
- fail loudly and transactionally roll back ambiguous historical state whose only discriminator was an internal-row-ID-derived fingerprint
- support zero-evidence and formula-unselected current producer shapes
- keep Q1 blocked on Q1b

## Validation

Exact head `44c99a94b5813a4d10736c4e6c83f0d29c91f4bd` contains current main `c8be43b1f541f6c52035460d83c59445a142316a`.

- 145 CKM tests passed after the final legacy-state authentication repair
- exact pre-v5 match normalizes/skips; same-watermark post-assessment evidence mutation remains stale and appends exactly one fresh v2 assertion
- tied-timestamp ordering, ambiguous-history rollback-without-half-init, supported multi-history, zero-evidence, and rebuild regressions pass
- ruff, mypy (768 files), and diff checks passed
- exact-SHA CI 19/19 acceptable; two consecutive Sol/high reviews clean

## BuilderOps Routing
- Records/projections/receipts: epic run `ckm-ma-3775-20260715` and linked LearningSignals
- Reason: recovery and migration-review divergence is recorded without changing Issue #3776 authority

## TCD

Sol/high for schema migration, historical identity, and rebuild determinism.


app/builderops/ckm/store.py
```diff
@@ -177,6 +430,16 @@ def rebuild(self) -> dict[str, Any]:
             conn.execute("PRAGMA foreign_keys = ON")
             for statement in CKM_DDL_STATEMENTS:
                 conn.execute(statement)
+            now = utc_now()
```
RasmusTho, line 888: **Blocking — make rebuild and revision one transaction.** Python sqlite3 does not implicitly begin a transaction for DROP/CREATE statements. In this sequence every table DROP and CREATE autocommits, and only the later INSERT into ckm_state begins the transaction that writes revision 1. A concurrent reader/writer can therefore observe or mutate a half-rebuilt schema, while the revision is not atomic with the rebuild it claims to represent. Start an explicit write transaction before the first DDL mutation and commit the DDL plus the singleton state row together; add a regression test that proves rollback leaves the prior schema/state intact.

app/builderops/ckm/store.py
```diff
@@ -101,6 +128,232 @@ def ensure_schema(self) -> dict[str, Any]:
...
+        except (OSError, ValueError):
+            # A custom/partial deployment may not carry the repository manifest.
+            # Existing rows still receive deterministic legacy identities below.
+            seed_identity_by_name = {}
+
+        for row in conn.execute("SELECT * FROM ckm_capability ORDER BY id").fetchall():
+            identity_key = str(row["identity_key"] or "")
+            if not identity_key:
+                identity_key = seed_identity_by_name.get(str(row["name"])) or (
+                    f"legacy:{row['existence_provenance']}:{row['id']}"
```
RasmusTho: **Blocking — migrated inferred IDs are not rebuild-stable.** This fallback incorporates the random internal row ID, but the normal inferred producer derives identity_key as inferred:{provenance}:{boundary_ref-or-name}. A pre-v5 inferred row therefore gets one public ID during migration and a different one after rebuild/reproduction. For example, legacy receipt:123 / cap_random migrates to legacy:receipt:123:cap_random, then rebuild produces inferred:receipt:123:Legacy inferred. Derive the migration fallback from the same durable producer key and extend the real pre-v5 migration fixture to assert the public ID survives rebuild.

app/builderops/ckm/seed.py
```diff
@@ -128,7 +140,8 @@ def seed_one(entry: SeedCapability) -> None:
             seed_one(by_slug[entry.parent])
             parent_id = seeded[entry.parent].id
         provenance = f"seeded:{entry.seed_source}"
-        existing = store.get_capability_by_name(entry.name)
+        identity_key = f"seed:{entry.stable_key}"
+        existing = store.get_capability_by_identity_key(identity_key)
         if existing is not None and (
```
RasmusTho, line 145: **Blocking — a stable-key rename is silently ignored.** The unchanged guard compares every seeded field except entry.name, so changing the manifest slug/name with the same stable_key returns the old row, reports changed=0, and never persists the new display name or revision. The acceptance test only checks public_id and misses that the rename did not happen. Include name in this comparison and assert renamed_seed.name plus the single revision advance.

app/builderops/ckm/store.py
```diff
@@ -634,6 +978,7 @@ def _set_inferred_edge_confirmed(self, edge_id: str) -> CkmEvidenceEdge:
         if edge.lifecycle == "confirmed":
             return edge
         with self._connect() as conn:
+            conn.execute("BEGIN IMMEDIATE")
```
RasmusTho, line 1541: **Blocking — revalidate confirmation state under the write lock.** The candidate/not-found checks occur before BEGIN IMMEDIATE. Two concurrent confirmations can both observe candidate; after the first commits, the second re-reads an already-confirmed row here, archives it, rewrites the same lifecycle, and advances the revision again. A concurrent delete can likewise leave row=None while this path still advances revision. After acquiring the lock, fail if absent and return unchanged if already confirmed before archiving/updating/bumping; cover the interleaving so a logical confirmation advances exactly once.

app/builderops/ckm/store.py
RasmusTho: Fixed in 618a8140: rebuild now starts BEGIN IMMEDIATE before the first DROP and commits DDL plus epoch/revision state atomically. The acceptance test injects invalid DDL and proves the prior schema, state identity, and capability row survive rollback.

app/builderops/ckm/store.py
RasmusTho: Fixed in 618a8140: legacy inferred fallback now uses the same inferred:{provenance}:{boundary-or-name} key as the producer. The migration fixture removes the v5 columns, migrates, rebuilds, reproduces the inferred capability, and asserts the public ID is unchanged.

app/builderops/ckm/seed.py
RasmusTho: Fixed in 618a8140: the seed unchanged guard now includes display name. The slug/name rename regression asserts changed=1, the new name persists, public ID is unchanged, and state revision advances exactly once.

app/builderops/ckm/store.py
RasmusTho: Fixed in 618a8140: confirmation now acquires BEGIN IMMEDIATE before reading lifecycle, refuses missing/non-inferred rows, and returns unchanged for already-confirmed rows. A two-thread regression proves one history row and exactly one revision advance.

app/builderops/ckm/store.py
```diff
@@ -101,6 +128,232 @@ def ensure_schema(self) -> dict[str, Any]:
...
+            conn.execute(
+                "UPDATE ckm_evidence_edge SET public_id = ? WHERE id = ?", (public_id, row["id"])
+            )
+        for row in conn.execute("SELECT * FROM ckm_evidence_edge_history ORDER BY history_id").fetchall():
+            public_id = str(row["public_id"] or "")
+            if not public_id:
+                active = conn.execute(
+                    "SELECT public_id FROM ckm_evidence_edge WHERE id = ?", (row["edge_id"],)
+                ).fetchone()
+                public_id = str(active["public_id"]) if active else CkmStore._edge_public_id(
```
RasmusTho: **Blocking — migrate retired history whose artifact was deleted.** `delete_artifacts_not_in()` intentionally archives each edge and then deletes both the active edge and its artifact, leaving a valid history-only row. In a pre-v5 database that row has no `public_id`; this fallback finds no active edge and calls `_edge_public_id()`, which requires the now-deleted `ckm_artifact` row and aborts the whole migration (`CKM identity preflight failed: ckm_artifact row ... has no public identity`). I reproduced this by deleting a stale artifact, stripping the v5 columns/state, and running `ensure_schema()`. Derive the retired edge identity from the history snapshot (its persisted artifact source ref plus the capability public ID), or otherwise preserve the missing referenced identity, and extend the real legacy fixture with a history-only edge produced by stale-artifact cleanup.

app/builderops/ckm/store.py
```diff
@@ -101,6 +128,232 @@ def ensure_schema(self) -> dict[str, Any]:
...
+        for row in conn.execute("SELECT * FROM ckm_assessment ORDER BY id").fetchall():
+            public_id = str(row["public_id"] or "") or CkmStore._assessment_public_id(conn, row)
+            conn.execute(
+                "UPDATE ckm_assessment SET public_id = ? WHERE id = ?", (public_id, row["id"])
+            )
+        for row in conn.execute("SELECT * FROM ckm_finding ORDER BY id").fetchall():
+            public_id = str(row["public_id"] or "") or CkmStore._finding_public_id(
+                conn, row["kind"], row["capability_id"], row["dimension"]
+            )
+            conn.execute("UPDATE ckm_finding SET public_id = ? WHERE id = ?", (public_id, row["id"]))
```
RasmusTho, line 339: **Blocking — migrate the persisted citation snapshots, not only table columns.** A real pre-v5 finding stores `artifact.to_dict()` / `edge.to_dict()` JSON without the newly required `public_id` fields. This backfill updates the artifact/edge/finding table rows but leaves those serialized snapshots unchanged; after migration `CkmFinding.validate()` reaches the new `CkmArtifact.from_row()` / `CkmEvidenceEdge.from_row()` contract and raises `KeyError: public_id`. The current migration test is created with the v5 models and only drops SQL columns, so its nested JSON already contains v5 identities and masks the failure. Backfill the persisted citation snapshots (including assessment citations where applicable), and make the fixture genuinely pre-v5 by stripping/asserting the nested payload shape before migration.

app/builderops/ckm/store.py
RasmusTho: Rebase follow-up: the atomic rebuild fix is now reachable on the current PR as 763942ae; PR #3786 remains the fixing PR. The rollback regression is unchanged and green after rebase.

app/builderops/ckm/store.py
RasmusTho: Rebase follow-up: the legacy inferred-identity fix is now reachable on the current PR as 763942ae; PR #3786 remains the fixing PR. The real missing-column migration/rebuild regression is green.

app/builderops/ckm/seed.py
RasmusTho: Rebase follow-up: the stable-key rename fix is now reachable on the current PR as 763942ae; PR #3786 remains the fixing PR. Name persistence and exactly-one revision remain covered.

app/builderops/ckm/store.py
RasmusTho: Rebase follow-up: the confirmation-race fix is now reachable on the current PR as 763942ae; PR #3786 remains the fixing PR. The two-thread history/revision regression is green.

app/builderops/ckm/store.py
RasmusTho: Fixed in current-head commit b72b77b2: retired history now derives artifact identity from its persisted source_ref when cleanup removed the artifact row, then combines it with the migrated capability identity and basis. The real missing-column fixture creates a history-only edge through stale-artifact cleanup and asserts the recovered ID.

app/builderops/ckm/store.py
RasmusTho: Fixed in current-head commit b72b77b2: migration recursively enriches serialized assessment and finding artifact/edge snapshots after table identities are backfilled. The pre-v5 fixture strips nested public_id fields, migrates, validates findings, and asserts assessment citation identities.

app/builderops/ckm/store.py
```diff
@@ -101,6 +128,342 @@ def ensure_schema(self) -> dict[str, Any]:
...
+        CkmStore._backfill_serialized_citation_identities(conn)
```
RasmusTho: **Blocking — serialized citation repair mutates v5 state without advancing revision.** `append_assessment()` currently accepts a pre-v5-shaped edge citation with no `public_id`; the next `ensure_schema()` enriches that persisted JSON here, but because the state row is already schema v5 it leaves `state_revision` unchanged. I reproduced revision 4 → 4 while the stored citation changed. Either reject incomplete citation snapshots at the producer or have the repair path detect changes and advance exactly once in this transaction; add a current-v5 repair regression, not only the state-less pre-v5 migration fixture.

app/builderops/ckm/store.py
```diff
@@ -101,6 +128,342 @@ def ensure_schema(self) -> dict[str, Any]:
...
+            "capability": CkmStore._referenced_public_id(
+                conn, "ckm_capability", str(row["capability_id"])
+            ),
+            "scores": {dimension: row[dimension] for dimension in MATURITY_DIMENSIONS},
+            "candidate_shares": row["candidate_shares"],
+            "formula_ids": row["formula_ids"],
+            "aggregate": row["aggregate"],
+            "aggregate_formula_id": row["aggregate_formula_id"],
+            "low_confidence": row["low_confidence"],
+            "edge_fingerprint": row["edge_fingerprint"],
```
RasmusTho: **Blocking — assessment public IDs still depend on internal row IDs.** This identity includes `edge_fingerprint`, while the production `assessment_fingerprint()` hashes `edge.id` and `artifact_id` and even sorts by those generated row IDs. Rebuilding and recreating identical capability/artifact/edge resources gives those resources the same public IDs but new row IDs, so the fingerprint and assessment `public_id` change even with identical scores, watermarks, and explicit valid/asserted timestamps. Derive the assessment identity input from public IDs/canonical semantic fields and extend the rebuild test to cover assessments.

app/builderops/ckm/store.py
```diff
@@ -101,6 +128,342 @@ def ensure_schema(self) -> dict[str, Any]:
...
+                )
+        return migrated
+
+    @staticmethod
+    def _preflight_schema(conn: sqlite3.Connection) -> None:
+        state_rows = conn.execute("SELECT * FROM ckm_state").fetchall()
+        if len(state_rows) != 1 or int(state_rows[0]["schema_version"]) != CKM_SCHEMA_VERSION:
+            raise CkmValidationError("CKM state preflight failed: missing or unsupported state row")
+        checks = {
+            "ckm_capability": ("public_id", "identity_key"),
```
RasmusTho, line 505: **Blocking — a partial schema with the full table-name set passes preflight.** Starting from a valid populated store, `ALTER TABLE ckm_artifact DROP COLUMN watermark`; `ensure_schema()` succeeds because the table set matches and this preflight checks only the new identity columns, then `list_artifacts()` fails with `IndexError: No item with that key`. That is the half-initialized unsupported state the migration AC says must refuse. Validate the required column/schema shape of every CKM table before commit and cover a missing-column fixture, not only a missing-table fixture.

app/builderops/ckm/store.py
```diff
@@ -216,43 +479,81 @@ def upsert_capability(
         parent_id: str | None = None,
         lifecycle: str = "candidate",
         boundary_ref: str | None = None,
+        identity_key: str | None = None,
     ) -> CkmCapability:
         now = utc_now()
         candidate_id = new_id("cap")
+        resolved_identity_key = identity_key or (
+            f"inferred:{existence_provenance}:{boundary_ref or name}"
+        )
```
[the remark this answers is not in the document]
RasmusTho: Fixed in current-head commit 2337ab48: upsert_capability now requires every producer to provide an explicit durable identity_key; the mutable name/provenance fallback is gone. All app/test producers were migrated, and rename/rebuild coverage uses the same permanent key.

app/builderops/ckm/store.py
RasmusTho: Fixed in current-head commit 2337ab48: assessment_fingerprint now uses edge/artifact public IDs and sorts by public ID, eliminating internal row IDs from the identity input. The acceptance test rebuilds capability/artifact/edge/assessment with new row IDs and asserts the fingerprint and assessment public ID remain identical.

app/builderops/ckm/store.py
RasmusTho: Fixed in current-head commit 2337ab48: serialized citation backfill reports whether it mutated data; ensure_schema advances current-v5 state revision exactly once in the same transaction and remains idempotent on the next ensure. A dedicated current-v5 malformed-citation regression proves both transitions.

app/builderops/ckm/store.py
RasmusTho: Fixed in current-head commit 2337ab48: schema.py now declares the required column set for every CKM table, with an explicit legacy-added allowlist. ensure_schema validates shape before mutation and full shape during preflight; a full-table-set fixture missing ckm_artifact.watermark now refuses loudly.

app/builderops/ckm/store.py
```diff
@@ -101,6 +140,367 @@ def ensure_schema(self) -> dict[str, Any]:
...
+            "capability": CkmStore._referenced_public_id(
+                conn, "ckm_capability", str(row["capability_id"])
+            ),
+            "scores": {dimension: row[dimension] for dimension in MATURITY_DIMENSIONS},
+            "candidate_shares": row["candidate_shares"],
+            "formula_ids": row["formula_ids"],
+            "aggregate": row["aggregate"],
+            "aggregate_formula_id": row["aggregate_formula_id"],
+            "low_confidence": row["low_confidence"],
+            "edge_fingerprint": row["edge_fingerprint"],
```
RasmusTho: **Blocking — migrate legacy assessment fingerprints before deriving public identity.** A real pre-v5 assessment’s `edge_fingerprint` was produced by the old algorithm that hashed `edge.id` and `artifact_id`. `_assessment_public_id()` still includes that recorded legacy fingerprint verbatim, so migration gives the assessment a row-ID-derived public ID even though new rebuilds use the repaired public-ID fingerprint. I reproduced an identical capability/artifact/edge/assessment (same scores, watermarks, `valid_from`, and `asserted_at`) migrating to `ckm_assessment_efcf…`, then rebuilding to `ckm_assessment_3d39…`. The current migration fixture is created with the new fingerprint and therefore masks this. Recompute the legacy fingerprint deterministically from the persisted citation snapshots/public identities (or fail loudly if the stored evidence cannot support it) before deriving the migrated assessment ID, and add a genuinely pre-v5 fingerprint → migrate → rebuild equality regression.

app/builderops/ckm/store.py
```diff
@@ -87,10 +94,42 @@ def ensure_schema(self) -> dict[str, Any]:
...
             self._migrate_assessment_explainability(conn)
             for statement in CKM_DDL_STATEMENTS:
+                if "CREATE UNIQUE INDEX" in statement and (
+                    "public_id" in statement or "identity_key" in statement
+                ):
+                    continue
                 conn.execute(statement)
+            initial_revision = 1 if legacy else 0 if not existing_tables else None
+            self._initialize_or_validate_state(conn, initial_revision=initial_revision)
+            citations_changed = self._backfill_public_identities(conn)
```
RasmusTho, line 135: **Blocking — do not silently repair current-v5 identity rows without a revision.** `_backfill_public_identities()` mutates blank `public_id`/`identity_key` values for both legacy and current schemas, but this branch advances the current-v5 revision only when serialized citations changed. On a valid v5 store I set one capability `public_id=''`; `ensure_schema()` repaired it while state revision remained 1 → 1, and the later blank-identity preflight could no longer fail loud. Restrict identity backfill to the explicit legacy migration path and reject damaged current-v5 identity state before mutation, or report identity changes and advance exactly once in this transaction. Add a current-v5 blank-identity regression alongside the citation-repair test.

app/builderops/ckm/store.py
RasmusTho: Fixed at 5c650ce6. Assessment public identity no longer includes the legacy edge_fingerprint field, which could embed pre-v5 internal row IDs. The migration test now seeds a row-ID-derived legacy fingerprint, migrates it, rebuilds the store with newly allocated internal IDs and a current public-ID fingerprint, and asserts the assessment public ID is unchanged.

app/builderops/ckm/store.py
RasmusTho: Fixed at 5c650ce6. ensure_schema now validates all persisted public_id/identity_key values for a current v5 schema before any migration/backfill runs. A regression test corrupts a current public ID, verifies typed refusal, verifies the state revision is unchanged, and verifies the blank value is not silently repaired.

app/builderops/ckm/store.py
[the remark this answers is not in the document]
RasmusTho: Fixed on the current head (introduced in 2337ab48 and retained at 5c650ce6). Legacy inferred capabilities derive identity_key from stable provenance plus boundary/name fallback, never the generated internal row ID; live producers must now pass an explicit identity_key. The migration/rebuild test verifies the inferred capability public ID survives newly allocated internal IDs.

app/builderops/ckm/store.py
```diff
@@ -101,6 +147,436 @@ def ensure_schema(self) -> dict[str, Any]:
...
+            ),
+            "scores": {dimension: row[dimension] for dimension in MATURITY_DIMENSIONS},
+            "candidate_shares": row["candidate_shares"],
+            "formula_ids": row["formula_ids"],
+            "aggregate": row["aggregate"],
+            "aggregate_formula_id": row["aggregate_formula_id"],
+            "low_confidence": row["low_confidence"],
+            "watermark_set": row["watermark_set"],
+            "valid_from": row["valid_from"],
+            "asserted_at": row["asserted_at"],
```
RasmusTho: [P1] Keep assessment IDs stable across the real rebuild producer

`asserted_at` (and `valid_from`) are included in the public-ID digest, but `assess_capabilities()` calls `append_assessment()` without either value, so both default to `utc_now()`. Rebuilding identical capability/artifact/edge/watermark inputs in a later second therefore produces a different assessment public ID. I reproduced this through the normal producer: the first assessment was `ckm_assessment_70b8386bb62245aad378ec76ae38e8ba` at `00:34:37Z`, the identical post-rebuild assessment was `ckm_assessment_5789a36befc43b9312a10cc236cd2682` at `00:34:38Z`, and both registry entries remained `active`. The AC requires every public resource identity to survive rebuild; please derive assessment identity from rebuild-stable inputs or make the producer restore a durable assertion identity, and add a test using `assess_capabilities()` rather than hand-supplying identical timestamps.

app/builderops/ckm/store.py
```diff
@@ -172,12 +648,52 @@ def rebuild(self) -> dict[str, Any]:
...
         with self._connect() as conn:
             conn.execute("PRAGMA foreign_keys = OFF")
+            conn.execute("BEGIN IMMEDIATE")
+            preserved_identities = []
+            preserved_successors = []
+            if conn.execute(
+                "SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = 'ckm_public_identity'"
+            ).fetchone():
+                preserved_identities = conn.execute(
+                    "SELECT * FROM ckm_public_identity ORDER BY public_id"
```
RasmusTho, line 838: [P1] Reconcile preserved active identities that are absent after rebuild

Preserving the entire registry keeps non-reuse history, but it also preserves every identity as `active` after all content tables are dropped. If an artifact (or its edges/findings/assessments) disappeared from the governed source before the rebuild, normal ingestion cannot tombstone it: `delete_artifacts_not_in()` only scans newly rebuilt `ckm_artifact` rows, so the old row is invisible and its registry entry stays active forever. A minimal sequence is create artifact A -> `rebuild()` -> run the source reconciliation with A absent; `identity_lifecycle(A.public_id)` still returns `active`. This violates the accepted deletion/content-free-tombstone semantics and makes registry lifecycle truth diverge from the projection. Please introduce an explicit rebuild reconciliation/finalization that tombstones preserved active identities not recreated, or otherwise retain enough producer state to retire them, with an end-to-end missing-after-rebuild test.

app/builderops/ckm/contracts.py
```diff
@@ -0,0 +1,384 @@
...
+    """Proves whether every declared object class was accounted for completely."""
+
+    object_classes: Sequence[ObjectClassCompleteness]
+    complete: bool
+
+    def __post_init__(self) -> None:
+        names = [item.object_class for item in self.object_classes]
+        if not names or len(names) != len(set(names)):
+            raise ValueError("completeness manifest needs distinct object classes")
+        actually_complete = all(
```
RasmusTho, line 189: [P1] Bind completeness accounting to the actual declared/read scope

`complete=True` is accepted whenever the entries supplied here have zero `omitted`/`truncated`; the contract neither requires all declared object classes nor checks `included` against the actual `read_set`/`ResultEnvelope.resources`. Thus a caller can build a measurement-eligible complete snapshot with only `capability(included=999)` and return one capability, or omit artifact/edge/etc. accounting entirely. That defeats the AC that incomplete captures cannot claim completeness and leaves Q1b no enforced proof of exact scope. Please make the declared scope explicit and validate per-class counts against the canonical read set/resources before allowing `complete=True`, with negative tests for missing classes and mismatched included counts.

app/builderops/ckm/store.py
```diff
@@ -213,46 +729,170 @@ def upsert_capability(
...
+        if relation not in {"split_successor", "merge_successor"}:
+            raise CkmValidationError(f"unsupported successor relation: {relation}")
+        with self._connect() as conn:
+            conn.execute("BEGIN IMMEDIATE")
+            capability = conn.execute(
+                "SELECT id FROM ckm_capability WHERE public_id = ?", (public_id,)
+            ).fetchone()
+            if capability is None:
+                raise CkmValidationError(f"active capability identity not found: {public_id}")
+            for successor_public_id in successor_public_ids:
```
RasmusTho, line 1039: [P1] Reject the tombstoned source as its own successor

The successor lookup runs while the source is still active, so callers may pass `successor_public_ids=[public_id]`. The method then tombstones that identity and records a split/merge successor link from the tombstone back to itself. This succeeds today even though the accepted lifecycle says split/merge creates new identifiers and links the tombstoned input to those new identities. Reject `successor_public_id == public_id` (and cover both relations) before mutating.

RasmusTho (commented):
Round 1 local review gate: blocking persistence/identity/completeness findings.

app/builderops/ckm/store.py
```diff
@@ -101,6 +159,594 @@ def ensure_schema(self) -> dict[str, Any]:
...
+                    conn.execute(
+                        f"UPDATE ckm_assessment SET {column} = ? WHERE id = ?",
+                        (_dumps(migrated), row["id"]),
+                    )
+        for row in conn.execute("SELECT id, citations FROM ckm_finding ORDER BY id").fetchall():
+            original = _loads(row["citations"])
+            migrated = CkmStore._backfill_citation_value(conn, original)
+            if migrated != original:
+                changed = True
+                conn.execute(
```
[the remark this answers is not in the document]
RasmusTho: Daily review-comment closure audit follow-up: this residual is now routed to #3967 with strict-ready `Verify:` targets. Classification: actionable residual missed before #3776 / PR #3786 closure; not discarded.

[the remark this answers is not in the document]
RasmusTho: Fixed by PR #3974, merged as 6ccc9dea62b7a731242a3e91b6bd1b42da19a253 and reachable from main. History-only v4 evidence-edge rows now receive stable tombstoned identities; ambiguous mappings and active/history domain mismatches roll back transactionally; already-buggy v5 lifecycle rows are repaired idempotently with valid timestamp chronology.

Issue #2143, opened 2026-06-18
epic: instance manages multiple vaults (registry, default/active selection, dimensions)


RasmusTho:
## Context

The shipped runtime has no-vault and one-active-vault behavior, an app-local known-vault seed, and
a v0 ActiveContextSet adapter, but still binds most work through a process-global selection. The
owner direction in #2143 is one instance managing several content vaults, including explicit
default, request/session selection, and non-authoritative dimensions. This parent validates the
bounded delivery defined in `docs/MULTI_VAULT_RUNTIME/`; it is never directly implemented.

TCD route: Sol/high–xhigh for registry migration, resolution authority, ActiveContextSet,
concurrency, and background lifecycle design; Terra/medium–high only for mechanically bounded
consumer migration and validation after those contracts are frozen.

## Scope

- Establish a durable instance-local registry keyed by stable binding identity while preserving
  logical-vault and local-clone identity.
- Add an explicit default with one-time last-active compatibility migration and fail-closed
  precedence.
- Version request/session ActiveContextSet resolution with immutable context generations and
  full-context isolation.
- Add non-authoritative dimensions over binding identities.
- Migrate request and background production consumers, reusing #3163 for watcher/settings rebind.
- Prove no-vault/one-vault compatibility, update owner docs/debt, and assemble the parent ledger.

## Source Anchors

- `docs/MULTI_VAULT_RUNTIME/README.md :: Capability boundary`
- `docs/contracts/ACTIVE_CONTEXT_SET.md :: Multi-Vault Runtime V1 Decision`
- `docs/CONCEPTS/VAULT_TOPOLOGY_CONTRACT.md :: Topology rules`
- `docs/CONCEPTS/VAULT_AND_SETTINGS_CONTEXT.md :: Future Multi-Vault`
- `app/vault/app_local.py :: AppLocalSettings / KnownVaultRef`
- GitHub issue #2143

## SBS Impact

- Primary subsystem: WSP
- Secondary subsystem(s): GOV, SFC, PDM, EBF, HKA, RCA, HIX, OEF
- Write class: mechanical durable instance registry/default/dimension/background intent; existing governed HKA writes only
- Authority impact: selection/grouping never grants authority; every binding is independently GOV-authorized
- Persistence impact: versioned instance-local Markdown registry plus lossless/compatible migrations
- Derived/rebuildable impact: context snapshots, caches, retrieval results, lifecycle instances, and health projections remain rebuildable
[...]

PR #3871, opened 2026-07-16
Add dormant instance vault registry core

RasmusTho:
## Change Lane
- [x] Implementation lane
- [ ] Docs authoring lane
- [ ] Governance lane

## Linked Issue
Governing-Issue: #3853
Closes #3853

## SBS Impact
- Primary subsystem: WSP
- Secondary subsystem(s): SFC, PDM, EBF
- Write class: mechanical durable instance-local state
- Authority impact: none; registry membership does not grant content access and authority activation remains sealed
- Persistence impact: versioned private Markdown state with atomic replace, file/directory fsync, verified last-good snapshots, migration, and CAS
- Derived/rebuildable impact: database/cache projections remain rebuildable and non-authoritative
- Human knowledge impact: none; no content-vault write or knowledge authority
- Memory impact: none
- Retrieval/context impact: supplies typed binding identity for later ActiveContextSet resolution without activating it
- Sync/deployment impact: dormant store contract only; MVR-01B still owns protected instance-state volumes and cross-consumer deployment
- External boundary impact: paths remain EBF binding metadata, not vault identity or authority
- New or changed contract: first-class dormant instance vault registry package/schema and legacy compatibility re-export
- Owner-doc impact: updated VAULT_AND_SETTINGS_CONTEXT, DEPLOYMENT_AND_ENVIRONMENTS, and RELEASE_CHANNELS
- Transition debt impact: reduces global/app-local package ambiguity; explicit re-export remains bounded compatibility debt
- Fitness rule impact: adds identity, migration, recovery, producer/preflight, permissions, locking/CAS, import-boundary, and read-purity proofs
- Boundary risk: high data/migration/concurrency risk, contained by dormant authority and fail-closed activation

## Owner-Doc Writeback

## Summary
- Adds the dormant, versioned instance-vault registry with stable binding identity, physical-root alias resolution, lossless legacy/settings-rebind migration, prepared/committed crash recovery, private atomic persistence, cross-process locking, and CAS.
- Moves production imports to app.instance.vault_registry while retaining app.vault.app_local as a compatibility re-export; the legacy scalar picker remains authoritative until MVR-01B/01C.
- Updates the three named owner-doc surfaces with the truthful dormant rollout boundary.

## Implementation Scope Check
- [x] Change stays within the linked Issue scope.
- [x] Constraints from the linked Issue were followed.
- [x] Acceptance Criteria from the linked Issue are satisfied.
- [x] Docs were updated in the same change when behavior/contracts changed.
- [x] Owner docs and roadmap/plan wording were updated when this PR turns a tracked backlog item into shipped reality.

## Docs Authoring Check
- [ ] This PR only changes approved docs-authoring surfaces.
- [ ] No code, runtime behavior, contracts, or shipped reality changed.
- [ ] This PR prepares or clarifies authoritative docs/specification and may later feed `docs-to-issue` extraction.

## Governance Lane Check
- [ ] This PR only changes approved governance surfaces.
- [ ] The change is limited to repo governance, agent workflow, or lightweight enforcement.
- [ ] No product/runtime implementation or shipped feature behavior changed.

## Validation
Implementation lane:
- [x] 12 exact Issue Verify targets: 12 passed
- [x] property/read-purity and focused registry coverage: 88 passed; docs_guard OK; DOCS_INDEX 4 passed
- [x] ruff check app tests companion-ui/companion-app: passed
- [x] mypy app: passed
- [x] CI selector repair: 2 focused tests passed; app/instance and tests/instance now resolve to vault coverage with no unowned paths
- [x] full non-PG suite: 9272 passed, 36 skipped, 18 xfailed; the one moved census line was repaired and rerun green; five unrelated startup tests were blocked by the host disk preflight (353-365 MiB free vs 1024 MiB required)

## BuilderOps Routing
- Records/projections/receipts: LearningSignals lrn_20260716043947_dd60faf4 and lrn_20260716051539_ed851601
- Reason: one signal routes missing subsystem-selector ownership to issue-to-code; the second routes Verify targets that underproved explicit physical-identity and crash-recovery clauses to feature-breakdown. Durable delivery evidence remains in Issue #3853, this PR, and parent #2143 receipts.

## Notes
Claim receipt: coordination_mode=github-label-only-fallback; fallback_reason=dispatcher_db_uninitialized; evidence=github-comment:4987935753. Parent #2143 remains an unclaimed blocked validation hub.


app/instance/vault_registry.py
```diff
@@ -0,0 +1,827 @@
...
+        return snapshot, payload
+
+    def _write_locked(self, snapshot: RegistrySnapshot) -> None:
+        if snapshot.schema != CURRENT_REGISTRY_SCHEMA or snapshot.authority != REGISTRY_AUTHORITY_DORMANT:
+            raise RegistryError("MVR-01A may persist only the dormant current registry schema")
+        payload = _render_markdown_settings(
+            self._frontmatter_from_snapshot(snapshot),
+            "# Instance Vault Registry\nMechanical instance-local state; registration does not grant vault authority.\n",
+        ).encode("utf-8")
+        _atomic_private_write(self.path, payload)
```
RasmusTho: **[P1] Keep the previous payload intact until the whole commit is durable**

`_write_locked` replaces the registry first and only then writes the last-good payload and checksum. If either later write fails, this method raises even though the main registry has already committed the new state. I reproduced this by failing the `.last-good` write during legacy migration: `load_or_migrate()` raised, but the original legacy payload had been replaced by the new schema. That violates the AC requiring write failure to leave the original payload recoverable and also makes ordinary mutation callers observe failure after commit. Stage and fsync the complete artifact set before the main-file commit point (or use an explicit recoverable journal) so a failed transaction cannot expose partially advanced state.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,827 @@
...
+                snapshot = self._empty_snapshot()
+                self._write_locked(snapshot)
+                return snapshot
+            try:
+                document = _read_document(self.path)
+            except (OSError, RegistryParseError):
+                return self._read_current_locked(recover=True)
+            schema = _optional_str(document.frontmatter.get("schema"))
+            if schema == CURRENT_REGISTRY_SCHEMA:
+                return self._snapshot_from_frontmatter(document.frontmatter)
```
RasmusTho, line 159: **[P1] Enforce private-file checks on the current-schema fast path**

This return bypasses `_read_current_locked`, so `load_or_migrate()` accepts a current registry with unsafe permissions. A focused probe created a valid registry, changed `registry.md` to `0644`, and this path returned it successfully, whereas the issue requires unsafe mode/ownership to fail loud. Validate the registry file through the same private/no-unsafe-path read path used by `load()` before accepting the current schema.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,827 @@
...
+        except OSError:
+            pass
+        temporary.unlink(missing_ok=True)
+        raise
+
+
+def _assert_private(path: Path, *, directory: bool) -> None:
+    if not path.exists():
+        raise RegistryError(f"required registry path is missing: {path}")
+    _assert_private_stat(path.stat(), path, directory=directory)
```
RasmusTho: **[P1] Reject symlinked registry transaction paths**

`Path.stat()` follows symlinks, so the regular-file/type check here validates the target rather than the registry path itself. I reproduced both `VaultRegistryStore.load()` and `preflight_registry_payload()` accepting `registry.md` as a symlink to an owner-owned `0600` file. That bypasses the fail-loud unsafe-path contract and permits registry state to live outside the governed private directory. Use `lstat()`/`O_NOFOLLOW` (and fd-based validation where races matter) for the registry, snapshots, checksum, and parent directory.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,827 @@
...
+        expected_revision: int | None = None,
+    ) -> RegistrySnapshot:
+        """Update mutable binding metadata without changing stable identities."""
+
+        with self._locked():
+            current = self._read_current_locked(recover=True)
+            self._assert_revision(current, expected_revision)
+            existing = current.registrations.get(registration.vault_binding_id)
+            if existing is None:
+                raise RegistryError(f"unknown vault_binding_id: {registration.vault_binding_id}")
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 7f244db8: update_registration now shares the same non-blank binding/ref/path validation as register. Regression: tests/instance/test_vault_registry.py::test_registry_rejects_blank_update_fields.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,827 @@
...
+                    raise RegistryMigrationError("one provisional binding identity matches multiple registrations")
+                binding_by_ref[matched_ref] = binding_id
+                value["ref"] = matched_ref
+                value["vaultBindingId"] = binding_id
+        registrations: dict[str, VaultRegistration] = {}
+        for ref, raw in candidates.items():
+            binding_id = binding_by_ref.get(ref) or f"binding-{uuid4()}"
+            if binding_id in registrations:
+                raise RegistryMigrationError(f"duplicate vault_binding_id during migration: {binding_id}")
+            registrations[binding_id] = VaultRegistration(
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 7f244db8: legacy candidates are grouped by canonical path, aliases map to one deterministic primary, and conflicting metadata fails closed. Regression: tests/instance/test_vault_registry_migration.py::test_migration_coalesces_canonical_path_aliases.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,827 @@
...
+                document = _read_document(self.path)
+            except (OSError, RegistryParseError):
+                return self._read_current_locked(recover=True)
+            schema = _optional_str(document.frontmatter.get("schema"))
+            if schema == CURRENT_REGISTRY_SCHEMA:
+                return self._snapshot_from_frontmatter(document.frontmatter)
+            if schema != APP_LOCAL_SCHEMA:
+                raise RegistryMigrationError(f"unsupported registry migration schema: {schema or '<missing>'}")
+            migrated = self._migrate_legacy_frontmatter(document.frontmatter)
+            self._write_locked(migrated)
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 7f244db8: migration first verifies an owned regular legacy source, then upgrades the owned legacy directory to 0700 and atomically replaces the 0644 file with 0600 state. Regression covers ordinary 0755/0644 producer modes.

app/instance/vault_registry.py
RasmusTho: Fixed in 7f244db8: the transaction captures all prior artifacts before any write and rolls back every attempted path on a late failure, including fsync. Regression injects snapshot failure and proves the original main remains byte-identical and ancillary files are not exposed.

app/instance/vault_registry.py
RasmusTho: Fixed in 7f244db8: load_or_migrate validates the current-schema file through the private-file gate before parsing it. Regression proves 0644 current state fails loud.

app/instance/vault_registry.py
RasmusTho: Fixed in 7f244db8: registry and transaction paths use lstat-based owner/type/mode validation; ancillary snapshot/checksum writes also reject symlinks before mutation. Regressions cover registry and snapshot symlinks and prove external targets remain unchanged.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,945 @@
...
+            raise RegistryError("registry revision must be a non-negative integer")
+        raw_registrations = frontmatter.get("registrations") or {}
+        if not isinstance(raw_registrations, dict):
+            raise RegistryError("registry registrations must be a mapping")
+        registrations: dict[str, VaultRegistration] = {}
+        refs: set[str] = set()
+        for binding_id, raw in raw_registrations.items():
+            if not isinstance(raw, dict):
+                raise RegistryError(f"registration {binding_id} must be a mapping")
+            registration = _registration_from_frontmatter(str(binding_id), raw)
```
RasmusTho, line 602: **[P1] Re-validate persisted registrations before accepting the snapshot**

This load path only rejects duplicate textual refs. It never applies the write-time canonical-path uniqueness invariant (or even `_validate_registration`) to registrations parsed from disk. I reproduced a current-schema payload containing `binding-a` at `/vault/a` and `binding-alias` at `/vault/../vault/a`; `VaultRegistryStore.load()` returned both successfully. That lets a hand-edited, partially migrated, or externally restored file bypass the identity-collision fail-closed AC and persist two binding identities for one root. Validate every parsed registration and reject duplicate canonical identities before returning the snapshot.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,945 @@
...
+                elif value is not None and merged[key] != value:
+                    raise RegistryMigrationError(
+                        f"conflicting alias metadata for legacy registrations: {primary}, {ref}"
+                    )
+        coalesced[primary] = merged
+    return coalesced, aliases
+
+
+def _canonical_path(value: str) -> str:
+    return str(Path(value).expanduser().resolve(strict=False))
```
RasmusTho: **[P1] Resolve physical-root aliases, not only lexical/symlink aliases**

The migration AC requires aliases of one physical root to coalesce, and the source contract explicitly includes bind-mount aliases/filesystem identity. `Path.resolve(strict=False)` normalizes `..` and follows symlinks, but two bind-mount paths to the same underlying directory remain different strings, so this code generates two registrations instead of coalescing them. Use the governed filesystem-identity resolver (for existing roots, at least stable device/inode identity with the required fallback/fail-closed behavior) so migration cannot mint duplicate bindings for a bind-mounted root.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,945 @@
...
+        previous = {
+            self.path: _read_previous_transaction_file(self.path, allow_legacy_mode=True),
+            self.snapshot_path: _read_previous_transaction_file(self.snapshot_path),
+            self.snapshot_checksum_path: _read_previous_transaction_file(self.snapshot_checksum_path),
+        }
+        attempted: list[Path] = []
+        try:
+            for path, contents in writes:
+                attempted.append(path)
+                _atomic_private_write(path, contents)
```
RasmusTho: **[P1] Make the multi-file commit recoverable across process/power loss**

The exception rollback fixes injected Python write errors, but the main registry is still replaced before the snapshot/checksum pair. If the process or host dies after the first replace, revision N is exposed with an N-1 recovery snapshot; if it dies after the snapshot replace, the snapshot and checksum disagree. The next lock holder has no journal/commit marker to distinguish or repair either state, and later corruption can lose a successfully exposed revision or fail recovery despite a complete payload. Because this store explicitly fsyncs for crash durability and the contract says every committed revision refreshes a checksum-verified last-good state, stage the artifact set behind a recoverable manifest/commit marker (or equivalent protocol) and verify restart recovery at each interruption point.

app/instance/vault_registry.py
RasmusTho: Fixed in 7124db7e: every persisted registration now runs non-blank validation and the same ref/physical-root uniqueness invariant used by writes. Regression hand-writes canonical and materialized-inode collisions and proves neither main nor last-good can return both bindings.

app/instance/vault_registry.py
RasmusTho: Fixed in 7124db7e: app.instance now owns a filesystem-root resolver that combines canonical path with materialized device/inode identity. Migration, rebind matching, writes, and reads all consume it; mocked bind-mount identity coalesces to the exact provisional binding while offline roots remain canonical-only.

app/instance/vault_registry.py
RasmusTho: Fixed in 7124db7e: each main/snapshot/checksum update persists a private digest-validated prepared/committed journal before mutation. The next lock holder rolls prepared generations back and committed generations forward before reading. Adversarial tests interrupt after snapshot, checksum, main, and commit marker; corrupt/digest-invalid/symlinked journals fail closed.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,1111 @@
...
+        if not isinstance(raw_known, dict):
+            raise RegistryMigrationError("legacy knownVaults must be a mapping")
+        raw_candidates: dict[str, dict[str, Any]] = {}
+        for ref, raw in raw_known.items():
+            if not isinstance(raw, dict):
+                raise RegistryMigrationError(f"legacy registration {ref} must be a mapping")
+            path = _optional_str(raw.get("path"))
+            if path is None:
+                raise RegistryMigrationError(f"legacy registration {ref} has no path")
+            raw_candidates[str(ref)] = dict(raw)
```
RasmusTho: [P1] Reject invalid legacy refs before committing migration. `str(ref)` permits an empty/whitespace legacy `knownVaults` key, and `_write_locked` never re-validates the generated snapshot. I reproduced `load_or_migrate()` returning success with `ref == ""`, followed by `load()` failing because both the main file and checksum-valid last-good snapshot are unreadable under `_validate_registration`. This violates lossless/fail-closed migration: validate every generated registration (or the complete snapshot) before writing, and add a blank-ref migration regression.

app/instance/vault_registry.py
```diff
@@ -0,0 +1,1111 @@
...
+    def __init__(self, path: Path) -> None:
+        self.path = path
+        self.lock_path = path.with_suffix(path.suffix + ".lock")
+        self.snapshot_path = path.with_suffix(path.suffix + ".last-good")
+        self.snapshot_checksum_path = path.with_suffix(path.suffix + ".last-good.sha256")
+        self.transaction_path = path.with_suffix(path.suffix + ".transaction")
+
+    def load(self) -> RegistrySnapshot:
+        with self._locked():
+            if not self.path.exists():
```
RasmusTho, line 143: [P1] Recover a missing populated main from last-good instead of reseeding empty. Both `load()` and `load_or_migrate()` take this branch whenever the main path is absent, even if `.last-good` plus its checksum contain a valid populated revision. I reproduced a revision-1 registry losing its only registration: unlinking only the main and calling `load()` returned revision 0, then overwrote the valid snapshot with empty state. The governing recovery contract allows empty reseed only for provably pre-MVR/empty state with no validated populated snapshot. Check/recover the verified snapshot before initialization, and cover missing-main restart for both entrypoints.

RasmusTho (commented):
Independent clean review A found blocking correctness issues; inline details follow.

app/instance/vault_registry.py
RasmusTho: Fixed in 6fbae7a4: legacy mapping keys are normalized and blank/duplicate normalized refs fail before migration builds or journals a snapshot; _write_locked also round-trips every candidate snapshot through the full persisted-state validator before any transaction begins. Regression proves the legacy bytes remain unchanged and no last-good is created.

app/instance/vault_registry.py
RasmusTho: Fixed in 6fbae7a4: a missing main now restores the checksum-verified last-good generation before load, load_or_migrate, or register may initialize/mutate state. Any partial or invalid snapshot pair fails closed. The exact recovery selector now deletes the populated main and proves load, migration, and mutation preserve all registrations and monotonic revision.

PR #3877, opened 2026-07-16
fix(deploy): preflight Companion UI pytest smoke before deploy mutation

RasmusTho:
## Change Lane
- [x] Implementation lane
- [ ] Docs authoring lane
- [ ] Governance lane

## Linked Issue
Fixes #3872

## SBS Impact
- Primary subsystem: Builder System release-channel deployment guard
- Secondary subsystem(s): Companion UI deploy smoke
- Write class: deployment-script behavior and regression test
- Authority impact: strengthens pre-mutation deploy validation without changing release authority
- Persistence impact: none beyond code and test files
- Derived/rebuildable impact: none
- Human knowledge impact: none
- Memory impact: none
- Retrieval/context impact: none
- Sync/deployment impact: prevents late tooling failure after a deploy mutation
- External boundary impact: checks local Python/pytest/test-module availability before mutation
- New or changed contract: Companion UI deploy preflight covers the smoke command import/start path, not only browser launch
- Owner-doc impact: none (no operator prerequisites changed; preflight remains offline and dependency-free)
- Transition debt impact: reduces
- Fitness rule impact: adds pre-mutation regression coverage
- Boundary risk: low

## Owner-Doc Writeback

## Summary
- `scripts/companion_ui_postdeploy_smoke.sh :: preflight` now also proves the actual post-deploy smoke command can import/start: an offline `pytest --collect-only -q` on `tests/companion_ui/test_companion_ui_live_smoke.py`, run before any deploy mutation. Pytest exit 5 (the module's intentional module-level self-skip when `COMPANION_UI_SMOKE_URL` is unset — the expected offline outcome) passes; any other nonzero exit (pytest missing, module import failure) fails loud with exit 86 before `write_pin`/Compose recreation. No gateway required, nothing installed, full smoke never runs pre-deploy.
- New harness regression `tests/deploy/test_deploy_channel.py::test_deploy_preflights_companion_pytest_smoke_before_pin_or_compose_mutation` proves the failure lands before pin writes and any docker invocation. The fake-python wrapper gained an additive env-driven branch (`FAKE_PYTEST_SMOKE_PREFLIGHT=fail`, default pass) so all existing tests keep their behavior.
- `scripts/deploy_channel.sh` unchanged: it already calls `preflight` before `write_pin`/compose mutation, and the existing exit-86 handling covers the extended preflight.

Resolves the unresolved PR #3832 review thread `PRRT_kwDOQEip6s6RNyS-` finding (browser-only preflight). Thread reply/resolution is coordinated post-merge per the issue AC.

## Implementation Scope Check
- [x] Change stays within the linked Issue scope.
- [x] Constraints from the linked Issue were followed.
- [x] Acceptance Criteria from the linked Issue are satisfied.
- [x] Docs were updated in the same change when behavior/contracts changed. (No doc-owned operator contract changed.)
- [x] Owner docs and roadmap/plan wording were updated when this PR turns a tracked backlog item into shipped reality. (Not applicable — review-follow-up fix, no roadmap item.)

## Validation
Implementation lane:
- [x] `ruff check app tests` — All checks passed!
- [x] Lint output included below.
- [ ] `mypy app` — not run: no `app/` files changed (bash script + deploy test only).
- [ ] `pytest -q -m "not pg"` — not run on this machine (laptop is not the runtime env by design); targeted full-file run below covers the only touched test surface, and the change touches no `app/` store-payload sink.
- [x] Additional targeted checks run as needed.

```
$ pytest -q tests/deploy/test_deploy_channel.py -k companion
2 passed, 5 deselected in 4.36s

$ pytest -q tests/deploy/test_deploy_channel.py
7 passed in 28.11s

$ bash -n scripts/companion_ui_postdeploy_smoke.sh scripts/deploy_channel.sh
(exit 0)

$ ruff check app tests
All checks passed!
```

Offline behavior of the real preflight verified directly on this host: `bash scripts/companion_ui_postdeploy_smoke.sh preflight` exits 0 with no gateway running (module self-skips, rc 5 accepted), and a genuine collection/import error exits 86.

## BuilderOps Routing
- Reason: bounded review-follow-up fix inside an existing deploy-guard contract; no plan divergence, no operational state to record.

## Notes
- Dispatcher unavailable (control plane: unsupported schema_version 3) — used GitHub-label-only claim (claim receipt comment on #3872, evidence=github-comment:4989060951).
- Residual risk: if a future environment's pytest maps a different meaning onto exit 5, the preflight would treat it as the self-skip; the error message includes the raw pytest exit code to keep diagnosis one-step.
- PR #3832 review thread `PRRT_kwDOQEip6s6RNyS-` resolution is deliberately left to the coordinator after merge (issue AC 3 verifies against post-merge thread state).

---


RasmusTho (commented):
Local review gate (deliver-issue-set coordinator) — round 1, 4 findings (2 inline, 2 on files outside the diff):

**1. [blocking] Stale preflight failure message in `scripts/deploy_channel.sh` (~line 374-377, not in this diff):** the generic handler still prints "browser preflight failed before channel mutation" when the new pytest-collection gate is what failed, misattributing an import/pytest failure to Playwright. Update the message to cover both gates (e.g. "companion UI preflight failed before channel mutation").

**2. [non-blocking, bundle if cheap] `docs/deployment/DEPLOYMENT_AND_ENVIRONMENTS.md:176`** still describes the pre-mutation gate as a browser-only preflight; the contract now also covers pytest smoke collection. One-line doc update, bundled in this PR per owner-doc bundling policy.

scripts/companion_ui_postdeploy_smoke.sh
```diff
@@ -39,6 +39,23 @@ except Exception:
...
+  # and never requires a running gateway. Without COMPANION_UI_SMOKE_URL set,
+  # the module intentionally self-skips at module level (pytest exit 5, "no
+  # tests collected"), which is the expected offline/pre-mutation outcome and
+  # must not be treated as a preflight failure; any other nonzero exit means
+  # pytest is missing or the module failed to import.
+  local pytest_status=0
+  "${PYTHON}" -m pytest \
+    "${ROOT}/tests/companion_ui/test_companion_ui_live_smoke.py" \
+    --collect-only -q || pytest_status=$?
+  if [ "${pytest_status}" -ne 0 ] && [ "${pytest_status}" -ne 5 ]; then
```
RasmusTho: [blocking — altitude] Accepting bare exit 5 conflates two states that must diverge: (a) legit module-level self-skip when COMPANION_UI_SMOKE_URL is unset (collect reports `1 skipped`), and (b) a genuinely emptied/broken-but-importable smoke module (collect reports `0 items`, **no** skip). Case (b) currently passes preflight and only fails after pin write + service recreation — the exact fail-after-mutation this issue exists to prevent. Distinguish them (e.g. capture collect output and require a skip marker or >=1 collected test before treating exit 5 as pass), and cover case (b) in the regression test.

tests/deploy/test_deploy_channel.py
```diff
@@ -137,6 +141,20 @@ def test_deploy_preflights_companion_browser_before_pin_or_compose_mutation(
     assert not (tmp_path / "docker-called").exists()
 
 
+def test_deploy_preflights_companion_pytest_smoke_before_pin_or_compose_mutation(
```
RasmusTho, line 168: [optional] Near-verbatim copy of test_deploy_preflights_companion_browser_before_pin_or_compose_mutation, differing only in env var + expected stderr. Consider one pytest.mark.parametrize over (env_var, expected_stderr) to stop the copy-per-gate pattern.

scripts/companion_ui_postdeploy_smoke.sh
RasmusTho: Fixed in e1bb2b8d. preflight() now captures the collect output (`--collect-only -q -ra`) and accepts exit 5 only when the standard `SKIPPED` summary marker proves the module imported and hit its intentional COMPANION_UI_SMOKE_URL self-skip. Case (b) — emptied-but-importable module, exit 5 with no skip marker — now prints the collect output and fails loud with exit 86 before any mutation. Regression covered: `test_deploy_preflights_companion_pytest_smoke_before_pin_or_compose_mutation` is parametrized over both failure modes (`fail` = import/pytest failure, `empty` = skip-less 'no tests collected' exit 5), and the fake pytest wrapper's collect-only branch emulates the realistic SKIPPED+exit-5 offline outcome by default, so every passing deploy test also exercises the acceptance path.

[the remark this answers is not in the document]
RasmusTho: Addressed in e1bb2b8d via outcome verification rather than env clearing: exit 5 now passes only with the `SKIPPED` summary marker (`-ra`), so any inherited selection option that deselects the module without importing it fails loud pre-mutation (exit 86) instead of passing silently. Verified on this host that `PYTEST_ADDOPTS=--ignore=<module>` specifically does not deselect an explicitly-passed file path (pytest collects it anyway and the self-skip marker appears), while a genuinely skip-less exit-5 outcome takes the fail branch. Env clearing was deliberately not done: the later real smoke run inherits the same env, so the preflight verifying the same-inheritance outcome is the faithful proof.

tests/deploy/test_deploy_channel.py
RasmusTho: Partially taken in e1bb2b8d: the pytest-smoke test is now parametrized over its two failure modes (fail/empty), which is where the new duplication pressure was. The cross-gate parametrize (merging the browser test into the same function) was deliberately declined: both function names are binding Verify targets in issue #3872's Acceptance Criteria (`::test_deploy_preflights_companion_browser_before_pin_or_compose_mutation` and `::test_deploy_preflights_companion_pytest_smoke_before_pin_or_compose_mutation`), and collapsing them would break the named-node contract. If the copy-per-gate pattern grows further, a shared assertion helper is the cleaner refactor than merging the named tests.

scripts/companion_ui_postdeploy_smoke.sh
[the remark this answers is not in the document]
RasmusTho: Triage classification: addressed; the original thread already records the fix in e1bb2b8d, including the skip-marker distinction before mutation. This is stale GitHub resolution state only, not open work. Evidence: https://github.com/RasmusTho/agentic-pkm-mvp/pull/3877#discussion_r3593587762

PR #3884, opened 2026-07-16
Canonicalize the vault settings location

RasmusTho:
Governing-Issue: #3161

Fixes #3161
Closes #3908

## Summary

- establish `<vault>/settings/` as the only canonical root with canonical-wins compatibility reads
- provide WriteGuard-gated migration, atomic publication/recovery, and durable SETTINGS-04 receipts
- migrate all producers and exclude legacy control roots from note/index flows
- receipt UAT canonical materialization before publication and skip byte-identical reruns

## Validation

Exact head `a226617e401c30c4345866176c17d53dcac60cff` contains base `0320859ff7bed12a927d576223f2def9de76ab16`.

- 170 passed, 1 skipped
- integrated settings UAT, ruff, mypy, and diff checks passed
- prior full non-pg candidate: 9,717 passed, 46 skipped, 18 xfailed

Earlier rejected reviews do not count. Exact-SHA CI and two clean rounds restart here.

## BuilderOps Routing
- Reason: no new BuilderOps object was created by this final receipt-only repair

## TCD

Sol/high for vault migration, durable publication, recovery, and WriteGuard.


tests/architecture/test_settings_single_location.py
```diff
@@ -0,0 +1,42 @@
...
+        left = _literal_path(node.left)
+        right = _literal_path(node.right)
+        if left is not None and right is not None:
+            return f"{left.rstrip('/')}/{right.lstrip('/')}"
+    return None
+
+
+def test_no_new_settings_paths() -> None:
+    violations: set[str] = set()
+    for path in sorted((ROOT / "app").rglob("*.py")):
```
RasmusTho: [P1] Extend the gate to cover settings producers outside app/

This walk only inspects `app/**/*.py`, so the new SET-2 gate passes while `scripts/dev/init_vault.py` still creates `vault/_system/settings`, writes `_system/settings/system-settings.yaml`, and labels/links that retired file as canonical. Running that fresh-init producer therefore immediately violates the AC that newly initialized vaults contain no legacy directories. Update the producer and make this gate cover `scripts/` (or explicitly enumerate all init/scaffold producers) so the invariant cannot regress outside `app/`.

app/settings/migration.py
```diff
@@ -0,0 +1,132 @@
...
+        published = True
+
+        for legacy in (
+            root / LEGACY_COMPILED_DIR,
+            (root / LEGACY_SYSTEM_SETTINGS).parent,
+            (root / LEGACY_HEALTH_SETTINGS).parent,
+        ):
+            if legacy.exists():
+                shutil.rmtree(legacy)
+        if backup.exists():
```
RasmusTho: [P1] Preserve unrelated files in the uppercase compatibility directory

`_migration_files()` deliberately relocates only `_system/Settings/health.md` from the uppercase spelling, but this cleanup recursively deletes the entire `_system/Settings` directory on case-sensitive filesystems. Any other operator file in that directory is never staged and is silently lost. Remove only the migrated `health.md` and prune the directory only when empty; on case-insensitive filesystems keep the existing alias/dedup behavior.

app/settings/migration.py
```diff
@@ -0,0 +1,132 @@
...
+def _markdown_from_yaml(raw: str) -> str:
+    return f"---\n{raw.rstrip()}\n---\n\n# System settings\n"
+
+
+def _migration_files(vault_root: Path) -> list[tuple[Path, Path, str | None]]:
+    mappings: list[tuple[Path, Path, str | None]] = []
+    compiled = vault_root / LEGACY_COMPILED_DIR
+    if compiled.exists():
+        for source in sorted(compiled.rglob("*")):
+            if source.is_file():
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 338a25e3: legacy @Settings/system-settings.yaml now normalizes to settings/system-settings.md with YAML-to-Markdown conversion. Covered by test_migration_converts_compiled_legacy_system_yaml.

app/settings/migration.py
```diff
@@ -0,0 +1,132 @@
...
+    staged = Path(tempfile.mkdtemp(prefix=".settings-migrate-", dir=canonical.parent))
+    backup = canonical.parent / ".settings-before-migration"
+    published = False
+    try:
+        if canonical.exists():
+            shutil.copytree(canonical, staged, dirs_exist_ok=True)
+        for source, relative, transform in mappings:
+            target = staged / relative
+            target.parent.mkdir(parents=True, exist_ok=True)
+            target.write_text(_target_text(source, transform), encoding="utf-8")
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 338a25e3: all normalized target mappings are compared before WriteGuard or mutation; differing retired sources fail closed and identical aliases collapse. Covered by test_migration_refuses_colliding_legacy_sources_before_guard.

tests/architecture/test_settings_single_location.py
RasmusTho: Fixed in 338a25e3: scripts/dev/init_vault.py and adjacent guard scripts now use settings/system-settings.md, and SET-2 scans scripts/**/*.py as well as app/**/*.py. The architecture/guard regression suite passes.

app/settings/migration.py
RasmusTho: Fixed in 338a25e3: cleanup now removes only the named uppercase health.md and prunes its directory only when empty; unrelated operator files survive (or are migrated once when lower/upper spellings alias). Covered by test_migration_preserves_unrelated_uppercase_system_files.

tests/architecture/test_settings_single_location.py
```diff
@@ -0,0 +1,46 @@
...
+    return None
+
+
+def test_no_new_settings_paths() -> None:
+    violations: set[str] = set()
+    code_paths = [
+        *(ROOT / "app").rglob("*.py"),
+        *(ROOT / "scripts").rglob("*.py"),
+    ]
+    for path in sorted(code_paths):
```
RasmusTho, line 209: [P1] Make SET-2 catch the operational scripts that still read retired roots

This gate now walks `scripts/**/*.py`, but `_literal_path` returns `None` as soon as the left side is a variable, so it misses `scripts/dev/ingest_samples.py` (`ROOT / "vault" / "_system" / "settings" / "system-settings.yaml"`). It also never parses embedded Python in shell scripts: both `scripts/export_runtime_env.sh` blocks still read `@Settings/providers.md`, and `scripts/lib/start_full_system_env.sh` still reads `_system/settings/system-settings.yaml`. The first script now crashes against this PR fixture because that YAML was removed; the startup/export scripts silently ignore canonical provider/path settings. Update these consumers and strengthen the gate (including shell/embedded-Python coverage) so SET-2 actually blocks retired operational reads.

app/settings/migration.py
```diff
@@ -0,0 +1,174 @@
...
+                continue
+            if source.name == ".gitkeep" or source.name.casefold() == "health.md":
+                continue
+            relative = source.relative_to(legacy_system_root)
+            if source == vault_root / LEGACY_SYSTEM_SETTINGS:
+                mappings.append((source, Path("system-settings.md"), "yaml_to_markdown"))
+            else:
+                mappings.append((source, relative, None))
+
+    legacy_health = vault_root / LEGACY_HEALTH_SETTINGS
```
RasmusTho: [P1] Migrate health.md from the configured legacy system directory

The pre-PR health loader resolved `<get_vault_system_dir_rel(vault)>/Settings/health.md`, and the new compatibility loader deliberately still supports that dynamic path. This migration only checks the hard-coded `_system/Settings/health.md`. Repro: a legacy system-settings file with `paths.system_dir_rel: Meta` plus `Meta/Settings/health.md` loads before migration; after `migrate-location`, `settings/health.md` is absent and the legacy file remains the active warned source. The command therefore does not complete canonicalization for supported custom-layout vaults. Resolve the legacy system dir before publication and include its named health artifact in the same conflict/preflight logic.

tests/architecture/test_settings_single_location.py
RasmusTho: Fixed in a1c8870c: live Python and embedded-shell readers now use the canonical compatibility resolver; SET-2 scans variable-prefixed Python paths and shell bodies. Operational export/start/canonical suite passes 60 tests.

app/settings/migration.py
RasmusTho: Fixed in a1c8870c: migration discovers every health path accepted by get_vault_system_dir_rel, migrates configured paths to settings/health.md, and removes only the migrated health artifact. Covered by test_migration_moves_health_from_configured_system_directory.

app/settings/migration.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def _legacy_health_paths(root: Path) -> tuple[Path, ...]:
+    paths = {root / LEGACY_HEALTH_SETTINGS}
+    try:
+        configured_system_dir = Path(get_vault_system_dir_rel(root))
+    except (OSError, ValueError):
+        pass
+    else:
+        paths.add(root / configured_system_dir / "Settings" / "health.md")
```
RasmusTho: **[P1] Constrain the configured health path to the selected vault**

`get_vault_system_dir_rel()` returns the configured string without enforcing that it is relative or contained by `root`. Because `Path` joining discards `root` for an absolute value (and `..` can also escape), this migration will treat an external `<configured>/Settings/health.md` as a legacy vault artifact, copy it into `settings/health.md`, and then unlink the external file in `_remove_legacy_sources()`. I reproduced this at this SHA with `paths.system_dir_rel` set to an absolute sibling directory: the external health file was deleted after an allowed migration. Resolve the candidate, require `candidate.relative_to(root.resolve())` (and reject absolute/traversing values) before it enters the mapping/cleanup set, with a regression proving no out-of-vault read or deletion.

RasmusTho: Fixed in 9186bf80: configured health paths are resolved and required to remain beneath the selected vault before WriteGuard or mutation. Traversal/absolute escape now fails closed, leaving external content untouched. Covered by test_migration_rejects_configured_health_path_outside_vault_before_guard (canonical matrix 10/10).

app/settings/health_settings.py
```diff
@@ -167,7 +160,14 @@ def load_health_settings(
                 configured_vault_root=None,
             )
 
-    target = vault_root / _settings_rel_path(vault_root)
+    target = resolve_settings_file(
+        vault_root,
+        "health.md",
+        legacy_paths=(
+            Path(get_vault_system_dir_rel(vault_root)) / "Settings" / "health.md",
```
RasmusTho: [P1] Contain the configured legacy health path before resolving it

`get_vault_system_dir_rel()` accepts both the environment override and the value stored in system settings. If that value is absolute, `root / legacy_path` discards `root`; if it contains `..`, it can likewise escape. With no canonical `health.md`, `resolve_settings_file()` will then read an arbitrary existing `.../Settings/health.md` outside the selected vault and use it as runtime health policy. The migration path correctly resolves and checks `is_relative_to(vault_root)`, but the production reader does not. Apply the same containment check here (including symlink resolution) and add absolute/`..`/symlink regressions.

app/settings/migration.py
```diff
@@ -0,0 +1,193 @@
...
+    published = False
+    try:
+        if canonical.exists():
+            shutil.copytree(canonical, staged, dirs_exist_ok=True)
+        for _source, relative, text in prepared:
+            target = staged / relative
+            target.parent.mkdir(parents=True, exist_ok=True)
+            target.write_text(text, encoding="utf-8")
+
+        if backup.exists():
```
RasmusTho: [P1] Do not recursively delete an unowned fixed backup path

Any pre-existing `<vault>/.settings-before-migration` directory is unconditionally removed here before publication. That may be an unrelated operator directory, or the only recovery copy left by an interrupted prior run, so the explicit settings migration can destroy data that was neither mapped nor collision-checked. Use a uniquely allocated sibling backup (or fail closed on collision before the guard/mutation) and preserve/recover an interrupted backup rather than deleting a fixed path.

scripts/lib/start_full_system_env.sh
```diff
@@ -73,6 +73,13 @@ import os
 import sys
 from pathlib import Path
 
+from app.settings.locations import (
```
RasmusTho: [P1] Bootstrap app imports while provider enforcement is active

This helper is sourced after `config/runtime.defaults.env`, which sets `LLM_PROVIDER_ENFORCE=1`. In infra/startup configurations where `LLM_PROVIDER` is intentionally not set yet, importing any `app.*` module executes `app.__init__.ensure_provider()` and raises before layout inference; under `set -e` that aborts the full-system script. `export_runtime_env.sh` already wraps these new settings imports with a temporary `LLM_PROVIDER=mock` and restores the operator value. This operational reader needs the same bootstrap (plus a no-provider/enforce=1 regression).

tests/architecture/test_settings_single_location.py
```diff
@@ -0,0 +1,54 @@
...
+        if left is not None and right is not None:
+            return f"{left.rstrip('/')}/{right.lstrip('/')}"
+        if right is not None:
+            return right
+    return None
+
+
+def test_no_new_settings_paths() -> None:
+    violations: set[str] = set()
+    code_paths = [
```
RasmusTho, line 203: [P2] Extend SET-2 enforcement to operational workflow paths

The gate only scans Python plus shell, so it currently passes while `.github/workflows/settings-ci.yaml` still names `vault/@Settings/**` in both path filters. A PR that only edits the new canonical `vault/settings/**` tree therefore does not trigger the dedicated settings compile/tests, and YAML/compose/config producers can regrow retired roots without this invariant noticing. Include the relevant operational text surfaces in the gate and update the settings workflow to trigger on `vault/settings/**` (and ensure the SET-2 test itself is attached to CI).

scripts/lib/start_full_system_env.sh
RasmusTho: Fixed in 75336639: the embedded startup reader temporarily bootstraps the mock provider only around app imports and restores the exact operator state. A provider-enforced and unset-provider startup regression passes.

app/settings/migration.py
RasmusTho: Fixed in 75336639: migration allocates an owned unique sibling backup and never removes the fixed operator backup path. The regression preserves an operator sentinel while exercising migration of an existing canonical tree.

tests/architecture/test_settings_single_location.py
RasmusTho: Fixed in 75336639: SET-2 scans operational YAML in GitHub workflows, config, scripts, and root compose files; settings-ci watches the canonical vault settings path and runs the SET-2 architecture test directly.

app/settings/health_settings.py
RasmusTho: Fixed in 75336639: the configured legacy health path is resolved and containment-checked against the selected vault before compatibility resolution. Absolute, parent traversal, and symlink escapes fail loud; regressions use real external health files and prove they remain untouched.

app/settings/migration.py
```diff
@@ -0,0 +1,217 @@
...
+                    mappings.append((source, relative, None))
+
+    legacy_system_root = contained_settings_path(
+        vault_root, (vault_root / LEGACY_SYSTEM_SETTINGS).parent
+    )
+    if legacy_system_root.exists():
+        for source in sorted(legacy_system_root.rglob("*")):
+            if not source.is_file():
+                continue
+            if source.name == ".gitkeep" or source.name.casefold() == "health.md":
```
RasmusTho: [P1][verification review round 2 — blocking] Case-sensitive filesystems lose `_system/settings/health.md` permanently.

`_migration_files` skips any `_system/settings` file whose name casefolds to `health.md` (to dedupe the `_system/Settings` alias), but `_legacy_health_paths` only re-adds `<root>/_system/Settings/health.md` and the configured `<system_dir>/Settings/health.md` (capital `S`). On Linux, a real `_system/settings/health.md` is therefore excluded from the prepared mappings, and `_remove_legacy_sources` then `shutil.rmtree`s `_system/settings` — the file is deleted without migration, backup, or conflict error, while the receipt reports success.

app/ingest/config.py
```diff
@@ -12,7 +12,7 @@
 
 DEFAULT_VAULT_ROOT = Path("vault")
 
-_BASE_IGNORES = [".obsidian/**", ".trash/**", "_system/companions/**"]
+_BASE_IGNORES = [".obsidian/**", ".trash/**", "settings/**", "_system/companions/**"]
```
RasmusTho, line 16: [P1][verification review round 2 — blocking] The ingest override reader was not moved to the canonical root, so overrides silently stop applying.

`_override_path` (line 63) still resolves `<vault>/<system_folder>/settings/ingest.override.md`, but this PR renames the packaged artifact to `vault/settings/ingest.override.md`, and `migrate_settings_location` relocates every `_system/settings/*` file to `<vault>/settings/` and removes `_system/settings`. After this PR (and after any migration), `_resolve_override` finds no file and returns `{}`: operator `include_folders`/`ignore_glob` overrides silently revert to layout defaults, and the relocated file is never read by anything because `settings/**` is now in `_BASE_IGNORES`. `app/cli/uat.py:_ingest_override_path` also still writes the legacy location.

app/settings/migration.py
RasmusTho: Fixed on current head ba5254fe: the case-sensitive lowercase `_system/settings/health.md` path is included in migration comparison/publication and covered by `test_migration_preserves_lowercase_legacy_health`; it can no longer be removed unmapped.

app/ingest/config.py
RasmusTho: Fixed in ba5254fe: ingest and UAT now resolve canonical `<vault>/settings/ingest.override.md` through `resolve_settings_file`, retain the loud legacy compatibility path, and prove canonical shadowing in `test_ingest_override_reads_canonical_root`.

app/settings/migration.py
```diff
@@ -0,0 +1,777 @@
...
+            canonical,
+            transaction,
+            expected_manifest=published_manifest,
+        )
+    if backup.exists():
+        if canonical.exists():
+            raise RuntimeError(
+                "canonical settings reappeared during recovery; operator recovery required"
+            )
+        os.replace(backup, canonical)
```
RasmusTho: [P1] Reject symlinked or tampered recovery backups before restoring

Any vault directory named .settings-migration-* with a transaction marker is treated as owned, but this recovery branch renames canonical-before into vault/settings without applying the fresh-source symlink/containment checks or validating the backup tree. On this exact head, I created a prepared transaction whose canonical-before was a symlink to an outside directory; _recover_interrupted_transaction returned normally and vault/settings became that outside symlink. This bypasses the containment invariant during recovery and can make a later migration read or report success against data outside the vault. Fail closed on symlinked/tampered transaction artifacts and validate the backup tree against a trusted manifest before os.replace; add a regression for a forged symlink backup.

app/settings/migration.py
```diff
@@ -0,0 +1,777 @@
...
+            os.fsync(descriptor)
+        finally:
+            os.close(descriptor)
+    for directory in reversed(directories):
+        _fsync_directory(directory)
+    _fsync_directory(root)
+
+
+def _markdown_from_yaml(raw: str) -> str:
+    return f"---\n{raw.rstrip()}\n---\n\n# System settings\n"
```
RasmusTho: [P1] Parse legacy YAML before emitting Markdown frontmatter

Wrapping the legacy bytes verbatim corrupts a valid YAML document that already starts with the standard --- marker. Exact reproduction: raw input --- followed by sync.debounce_ms: 99 produces two adjacent opening delimiters, and read_settings_mapping returns an empty mapping. The migration can still receipt the transaction and quarantine the only legacy source, silently replacing real settings with defaults. Parse the legacy YAML into a mapping before entering the write transaction, fail closed on invalid or non-mapping content, and serialize one normalized frontmatter block; add a leading-document-marker regression.

app/settings/migration.py
```diff
@@ -0,0 +1,777 @@
...
+            not _legacy_manifest_matches(root, legacy_fingerprints)
+            or _canonical_manifest(root, canonical) != canonical_fingerprints
+        ):
+            raise RuntimeError("settings changed during migration preparation; retry required")
+
+        _fsync_tree(staged)
+        staged_manifest = _canonical_manifest(root, staged)
+
+        if canonical.exists():
+            os.replace(canonical, backup)
```
RasmusTho: [P1] Keep a complete canonical tree visible to live readers during publication

This first rename removes vault/settings before the staged tree is renamed into place, while normal compiler, watcher, and startup readers do not take this migration lock. I instrumented this exact gap and called resolve_compiled_sources: it returned only the legacy global.md and silently omitted a pre-existing canonical watchers.md, so a live reload can publish a partial/default bundle instead of the previous valid bundle. The crash-recovery test only proves a later migration can repair disk state; it does not provide the required atomic visibility to concurrent readers. Use an atomic indirection/exchange strategy, or a shared read/write protocol that makes every settings reader wait/recover/fail loud across this swap.

tests/architecture/test_settings_single_location.py
```diff
@@ -0,0 +1,254 @@
...
+    "global.md",
+    "flows.settings.yaml",
+    "health.md",
+    "ingest.override.md",
+    "llm_routing.md",
+    "models.md",
+    "providers.md",
+    "routing.md",
+    "standards.md",
+    "system-settings.md",
```
RasmusTho, line 41: [P2] Include the retired YAML artifact in the SET-2 location gate

The gate enumerates system-settings.md but not the actual retired artifact system-settings.yaml. On this exact head, _python_path_violations(Path("app/example.py"), "path = Path(\\"rogue/system-settings.yaml\\")") returns an empty list, so CI permits a new non-canonical legacy settings location despite the acceptance criterion. Add system-settings.yaml to the guarded artifact set and a negative regression, while keeping only the explicitly enumerated compatibility references allowlisted.

app/settings/migration.py
RasmusTho: Fixed in 76ffb6ad: legacy YAML is safe-loaded as a mapping and safe-dumped into exactly one normalized frontmatter block. A leading YAML document marker is preserved semantically instead of creating adjacent delimiters. Regression: test_migration_normalizes_legacy_yaml_document_marker.

app/settings/migration.py
RasmusTho: Fixed in 76ffb6ad: an existing canonical directory is published with one same-filesystem kernel exchange (renameatx_np on Darwin / renameat2 on Linux), so readers see the complete old or complete new tree. Crash and ordinary-error recovery preserve the prior tree; exact reader, crash, fsync, and persistent backup-rename fault checks are green.

app/settings/migration.py
RasmusTho: Fixed in 76ffb6ad: recovery markers now persist original/published manifests, and every restore candidate must be a contained real directory with the exact recorded fingerprint. Symlinked or tampered backup trees fail closed before exchange. Regression: test_recovery_rejects_symlinked_backup_before_restoring_it (49-test exact suite green).

tests/architecture/test_settings_single_location.py
RasmusTho: Fixed in 76ffb6ad: SET-2 now enumerates retired system-settings.yaml and the path-shape regression proves rogue/system-settings.yaml is rejected while the explicit locations.py compatibility literal remains allowlisted.

app/settings/migration.py
```diff
@@ -0,0 +1,977 @@
...
+                        transaction,
+                        expected_manifest=staged_manifest,
+                        backup_manifest=canonical_fingerprints,
+                        had_canonical=had_canonical,
+                    )
+                except Exception:
+                    recovery_collision = True
+                    rollback_incomplete = True
+            if transaction.exists() and not recovery_collision:
+                _write_transaction_state(
```
RasmusTho: [P1][fresh high-risk review round 1 — blocking] Always release the migration lock when rollback-marker persistence fails

This `_write_transaction_state(..., "rolled_back")` call runs inside the exception handler but before lines 931-932 release/close `lock_descriptor`. If marker persistence itself raises (for example an fsync/replace error), control exits the handler immediately, so the original root flock and file descriptor remain live for the lifetime of the process. I reproduced this on exact head `76ffb6ad208e96b307a260df20d082fcf46be0fc` by forcing the durable receipt to fail after publication and forcing only the `rolled_back` marker write to raise: the previous canonical tree was restored, the call raised `OSError`, and a separate subprocess `LOCK_EX|LOCK_NB` probe on the vault root failed with `BlockingIOError` (probe rc 73). Any later in-process migration then blocks behind the leaked lock, so the ordinary-error recovery state machine can wedge permanently. Put unlock/close in an outer `finally` (without masking the recovery error), and add this double-failure lock probe as a regression.

RasmusTho: Fixed in bf6ceb8e. Ordinary rollback cleanup is now enclosed by an exception-safe finally that always calls `_release_lock`; `_release_lock` closes the descriptor even if `LOCK_UN` raises, while the rollback-marker error still propagates and the prior canonical remains restored. Exact subprocess regression `test_migration_releases_lock_when_rollback_marker_write_fails` proves a separate process can immediately acquire `LOCK_EX|LOCK_NB`. Exact suite 50/50; broader matrix 121/121.

Issue #3843, opened 2026-07-15
feature: local Keychain secret provisioning


RasmusTho:
## Topology reconciliation (2026-08-30)

Current owner authority supersedes every future-use instruction below that treats Demerzel/Mac mini,
local Colima, `workspace-prod`, or workstation loopback as the `dev`, `test`, or `prod` Product
Runtime. The intended placement is the TARS-hosted Linux VM topology; exact per-channel VM, engine,
source/image, ingress, health, data, backup, and rollback identity must come from the repository-only
#5237 contract and a later governed qualification receipt. Demerzel remains a control,
development, client, and operator computer. Historical observations below remain evidence, but are
not current placement or acceptance authority.

Until #5237 is delivered and the issue-specific channel prerequisites are evidenced, no host,
container, service, deployment, model, secret, database, or release mutation is authorized by this
Issue. Any later human operation or authorization gate starts only after those prerequisites pass.
## Context

PR #3838 merged the Local Secret Provisioning capability specification. The existing #3830 dev capture-watch receipt is prior evidence only; this feature delivers the reusable Keychain-backed boundary without introducing a product subsystem or cloud secret service.

## Scope

Deliver the HSP-01 and HSP-02 child slices, collect their validation receipts here, then validate that bootstrap preserves the healthy dev capture-watch path.

## Source Anchors

- `docs/LOCAL_SECRET_PROVISIONING/README.md :: Capability acceptance criteria`
- `docs/LOCAL_SECRET_PROVISIONING/README.md :: Cross-task invariants`

## SBS Impact

- Primary subsystem: Product/Runtime host/deploy boundary
- Secondary subsystem(s): Builder System credential consumer; Heimdal capture runtime
- Write class: boundary infrastructure, configuration, and security controls
- Authority impact: no Product memory or routing authority; carries declared credentials only
- Persistence impact: macOS Keychain entries and temporary runtime material; no repository secret store
- Derived/rebuildable impact: redacted validation receipts only
- Human knowledge impact: none
- Memory impact: none
- Retrieval/context impact: none
- Sync/deployment impact: channel-scoped process bootstrap
- External boundary impact: macOS Keychain; no cloud secret service
- New or changed contract: Local Secret Provisioning HSP-01/HSP-02
- Owner-doc impact: update `docs/SECURITY.md` only after feature acceptance
- Transition debt impact: reduces plaintext/deploy-file secret drift
[...]

PR #3888, opened 2026-07-16
Add host secret contract

RasmusTho:
Governing-Issue: #3845
Fixes #3845
Fixes #3905

## Summary
- Define a deterministic, channel-scoped Keychain account mapping for declared host-secret tuples.
- Fail closed on undeclared, noncanonical, and duplicate JSON contract values; register the `app/ops` test-selection owner.
- Percent-encode account components so distinct declared tuples cannot collide.

## Validation
- pytest -q tests/ops/test_host_secret_contract.py tests/scripts/test_select_pr_tests.py tests/governance/test_select_pr_tests.py (84 passed after rebase)
- PYTHONPATH="$PWD:$PWD/companion-ui/companion-app" pytest -q -m "not pg" (9801 passed, 46 skipped, 225 deselected, 18 xfailed before the clean rebase)
- mypy app
- ruff check app tests scripts
- python3 scripts/public_seam_lint.py --mode gate
- python3 scripts/select_pr_tests.py --changed-file app/ops/host_secret_contract.py --changed-file tests/ops/test_host_secret_contract.py

## BuilderOps Routing
- Records/projections/receipts: runtime/builderops/epic-runs/local-secrets-20260716.json
- Reason: Parent-feature delivery coordination for #3843; no Product memory material created.

### DIFF
--- a/app/ops/host_secret_contract.py
+++ b/app/ops/host_secret_contract.py
@@ -8,6 +8,9 @@ from pathlib import Path
 
 
 DEFAULT_CONTRACT_PATH = Path("config/secrets/host_secret_contract.json")
+_CONTRACT_FIELDS = frozenset({"version", "keychain_service", "keychain_account_template", "consumers"})
+_CONSUMER_FIELDS = frozenset({"channel", "consumer", "secrets"})
+_KEYCHAIN_ACCOUNT_TEMPLATE = "{channel}:{consumer}:{secret}"
 
 
 class UndeclaredSecretConsumerError(ValueError):
@@ -17,6 +20,7 @@ class UndeclaredSecretConsumerError(ValueError):
 @dataclass(frozen=True)
 class HostSecretContract:
     keychain_service: str
+    keychain_account_template: str
     allowed: frozenset[tuple[str, str, str]]
 
     def require_declared(self, *, channel: str, consumer: str, secret: str) -> None:
@@ -25,14 +29,30 @@ class HostSecretContract:
                 f"undeclared host secret request: channel={channel!r}, consumer={consumer!r}, secret={secret!r}"
             )
 
+    def keychain_account(self, *, channel: str, consumer: str, secret: str) -> str:
+        """Return the declared, channel-scoped Keychain account identifier."""
+        self.require_declared(channel=channel, consumer=consumer, secret=secret)
+        return self.keychain_account_template.format(
+            channel=channel, consumer=consumer, secret=secret
+        )
+
 
 def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretContract:
     payload = json.loads(path.read_text(encoding="utf-8"))
-    if payload.get("version") != 1 or not isinstance(payload.get("keychain_service"), str):
+    if not isinstance(payload, dict) or set(payload) != _CONTRACT_FIELDS:
+        raise ValueError("invalid host secret contract")
+    if (
+        payload["version"] != 1
+        or not isinstance(payload["keychain_service"], str)
+        or payload["keychain_account_template"] != _KEYCHAIN_ACCOUNT_TEMPLATE
+        or not isinstance(payload["consumers"], list)
+    ):
         raise ValueError("invalid host secret contract")
     allowed: set[tuple[str, str, str]] = set()
-    for item in payload.get("consumers", []):
-        channel, consumer, secrets = item.get("channel"), item.get("consumer"), item.get("secrets")
+    for item in payload["consumers"]:
+        if not isinstance(item, dict) or set(item) != _CONSUMER_FIELDS:
+            raise ValueError("invalid host secret consumer declaration")
+        channel, consumer, secrets = item["channel"], item["consumer"], item["secrets"]
         if not isinstance(channel, str) or not isinstance(consumer, str) or not isinstance(secrets, list):
             raise ValueError("invalid host secret consumer declaration")
         for secret in secrets:
@@ -41,4 +61,8 @@ def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretC
             allowed.add((channel, consumer, secret))
     if not allowed:
         raise ValueError("host secret contract declares no consumers")
-    return HostSecretContract(keychain_service=payload["keychain_service"], allowed=frozenset(allowed))
+    return HostSecretContract(
+        keychain_service=payload["keychain_service"],
+        keychain_account_template=payload["keychain_account_template"],
+        allowed=frozenset(allowed),
+    )
--- a/app/ops/host_secret_contract.py
+++ b/app/ops/host_secret_contract.py
@@ -12,6 +12,7 @@ DEFAULT_CONTRACT_PATH = Path("config/secrets/host_secret_contract.json")
 _CONTRACT_FIELDS = frozenset({"version", "keychain_service", "keychain_account_template", "consumers"})
 _CONSUMER_FIELDS = frozenset({"channel", "consumer", "secrets"})
 _KEYCHAIN_ACCOUNT_TEMPLATE = "{channel}:{consumer}:{secret}"
+_KEYCHAIN_SERVICE = "yggdrasil.host-secrets"
 _INITIAL_CHANNELS = frozenset({"dev", "test", "prod"})
 _INITIAL_CONSUMER = "heimdal-capture-watch"
 _INITIAL_SECRET = "heimdal.raw-store-key"
@@ -48,8 +49,9 @@ def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretC
     if not isinstance(payload, dict) or set(payload) != _CONTRACT_FIELDS:
         raise ValueError("invalid host secret contract")
     if (
-        payload["version"] != 1
-        or not isinstance(payload["keychain_service"], str)
+        type(payload["version"]) is not int
+        or payload["version"] != 1
+        or payload["keychain_service"] != _KEYCHAIN_SERVICE
         or payload["keychain_account_template"] != _KEYCHAIN_ACCOUNT_TEMPLATE
         or not isinstance(payload["consumers"], list)
     ):
--- a/app/ops/host_secret_contract.py
+++ b/app/ops/host_secret_contract.py
@@ -1,0 +1,44 @@
+"""Value-free contract for host-local Keychain secret consumers."""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+from pathlib import Path
+
+
+DEFAULT_CONTRACT_PATH = Path("config/secrets/host_secret_contract.json")
+
+
+class UndeclaredSecretConsumerError(ValueError):
+    """Raised without secret material when a consumer requests an undeclared key."""
+
+
+@dataclass(frozen=True)
+class HostSecretContract:
+    keychain_service: str
+    allowed: frozenset[tuple[str, str, str]]
+
+    def require_declared(self, *, channel: str, consumer: str, secret: str) -> None:
+        if (channel, consumer, secret) not in self.allowed:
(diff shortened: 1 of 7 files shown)


config/secrets/host_secret_contract.json
```diff
@@ -0,0 +1,9 @@
+{
+  "version": 1,
+  "keychain_service": "yggdrasil.host-secrets",
+  "consumers": [
+    {"channel": "dev", "consumer": "heimdal-capture-watch", "secrets": ["heimdal.raw-store-key"]},
+    {"channel": "test", "consumer": "heimdal-capture-watch", "secrets": ["heimdal.raw-store-key"]},
+    {"channel": "prod", "consumer": "heimdal-capture-watch", "secrets": ["heimdal.raw-store-key"]}
```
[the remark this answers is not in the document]
RasmusTho: Fixed by 5f97f9a4: the contract now declares and validates the deterministic `{channel}:{consumer}:{secret}` account mapping, and tests prove distinct dev/test/prod accounts.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,44 @@
...
+            )
+
+
+def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretContract:
+    payload = json.loads(path.read_text(encoding="utf-8"))
+    if payload.get("version") != 1 or not isinstance(payload.get("keychain_service"), str):
+        raise ValueError("invalid host secret contract")
+    allowed: set[tuple[str, str, str]] = set()
+    for item in payload.get("consumers", []):
+        channel, consumer, secrets = item.get("channel"), item.get("consumer"), item.get("secrets")
```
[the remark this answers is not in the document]
RasmusTho: Fixed by 5f97f9a4: loader validation now rejects unknown top-level and consumer-declaration fields; bounded tests cover both fail-closed paths.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,68 @@
...
+    keychain_account_template: str
+    allowed: frozenset[tuple[str, str, str]]
+
+    def require_declared(self, *, channel: str, consumer: str, secret: str) -> None:
+        if (channel, consumer, secret) not in self.allowed:
+            raise UndeclaredSecretConsumerError(
+                f"undeclared host secret request: channel={channel!r}, consumer={consumer!r}, secret={secret!r}"
+            )
+
+    def keychain_account(self, *, channel: str, consumer: str, secret: str) -> str:
```
RasmusTho, line 44: The colon-join is not an injective tuple mapping for declarations the loader accepts. For example, `(channel="dev:x", consumer="y", secret="z")` and `(channel="dev", consumer="x:y", secret="z")` both resolve to `dev:x:y:z`; empty strings and `:` are currently valid in every component. A future contract expansion could therefore make HSP-02 retrieve the wrong declared Keychain item while passing `require_declared`. Reject reserved delimiters/empty identifiers (with a collision test), or encode the components injectively before deriving the account.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,71 @@
...
+        raise ValueError("invalid host secret contract")
+    allowed: set[tuple[str, str, str]] = set()
+    for item in payload["consumers"]:
+        if not isinstance(item, dict) or set(item) != _CONSUMER_FIELDS:
+            raise ValueError("invalid host secret consumer declaration")
+        channel, consumer, secrets = item["channel"], item["consumer"], item["secrets"]
+        if not isinstance(channel, str) or not isinstance(consumer, str) or not isinstance(secrets, list):
+            raise ValueError("invalid host secret consumer declaration")
+        for secret in secrets:
+            if not isinstance(secret, str):
```
RasmusTho: **<sub><sub>P2 Badge</sub></sub>  Enforce value-free identifiers, not merely field names**

The exact-key checks close the `raw_key`-field bypass, but the loader still accepts arbitrary strings in `secrets` (and in `channel`/`consumer`). A tracked edit such as `"secrets": ["<actual-key-material>"]` is therefore a valid contract and passes `test_contract_is_value_free`, which only rejects two literal substrings. That leaves the value-free invariant unenforced on the fields that can actually carry the key. Since HSP-01 has only the fixed initial identifier, validate the declared identifiers against an explicit allowed set (or an equally strict value-free schema) and add a regression test with secret-shaped material.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+
+def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretContract:
+    payload = json.loads(path.read_text(encoding="utf-8"))
+    if not isinstance(payload, dict) or set(payload) != _CONTRACT_FIELDS:
+        raise ValueError("invalid host secret contract")
+    if (
+        payload["version"] != 1
+        or not isinstance(payload["keychain_service"], str)
```
RasmusTho: **<sub><sub>P1 Badge</sub></sub>  Pin the Keychain service namespace to the v1 constant**

`keychain_service` is still accepted as any string, so replacing the tracked value with actual key material (for example, `"keychain_service": "<actual-key-material>"`) loads successfully. That bypasses the claimed value-free/strict-identifier enforcement even though the docs call this a stable non-secret namespace and say other identifier-bearing strings are rejected. Require the fixed `yggdrasil.host-secrets` value and add a regression test that a secret-shaped service value is rejected.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,80 @@
...
+            secret=quote(secret, safe=""),
+        )
+
+
+def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretContract:
+    payload = json.loads(path.read_text(encoding="utf-8"))
+    if not isinstance(payload, dict) or set(payload) != _CONTRACT_FIELDS:
+        raise ValueError("invalid host secret contract")
+    if (
+        payload["version"] != 1
```
RasmusTho: **<sub><sub>P2 Badge</sub></sub>  Reject a boolean schema version explicitly**

Python considers `True == 1`, so a tracked payload with `"version": true` passes this v1 check and is loaded as a valid contract. The loader otherwise enforces a strict JSON schema; require an actual integer version (excluding `bool`) before comparing it to `1`, with a malformed-contract regression test.

app/ops/host_secret_contract.py
RasmusTho: Fixed by 4d9ed6e8: each declared component is percent-encoded before account derivation; the regression test proves delimiter-containing tuples cannot collide.

app/ops/host_secret_contract.py
RasmusTho: Fixed by 712daea4: v1 now accepts only its explicit channels, consumer, and logical identifier; regression tests reject value-bearing alternatives.

app/ops/host_secret_contract.py
RasmusTho: Fixed by 1ad301f1: the loader now requires the canonical `yggdrasil.host-secrets` service and rejects a value-bearing replacement.

app/ops/host_secret_contract.py
RasmusTho: Fixed by 1ad301f1: the loader now requires `type(version) is int` before accepting v1; regression coverage rejects `true`.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,82 @@
...
+        self.require_declared(channel=channel, consumer=consumer, secret=secret)
+        return self.keychain_account_template.format(
+            channel=quote(channel, safe=""),
+            consumer=quote(consumer, safe=""),
+            secret=quote(secret, safe=""),
+        )
+
+
+def load_host_secret_contract(path: Path = DEFAULT_CONTRACT_PATH) -> HostSecretContract:
+    payload = json.loads(path.read_text(encoding="utf-8"))
```
RasmusTho: **<sub><sub>P1 Badge</sub></sub>  Reject duplicate JSON object keys before schema validation**

`json.loads` silently keeps the last duplicate key. A tracked contract can therefore contain raw secret material in an earlier duplicate field while the final value remains canonical, e.g. two `keychain_service` keys with the first set to a key and the second to `yggdrasil.host-secrets`; this loader accepts it and the raw-text value-free test also misses it. Parse with an `object_pairs_hook` that rejects duplicate keys (at every object level) and add this regression case, otherwise the repository-wide no-plaintext invariant is bypassable.

docs/LOCAL_SECRET_PROVISIONING/DEFINE_HOST_SECRET_CONTRACT.md
```diff
@@ -26,6 +26,15 @@ accidental secret manager.
...
    explicit non-rotation rule for `HEIMDAL_RAW_STORE_KEY`.
 
+## Operator runbook
+
+Create or inspect a Keychain item only through the logical identifier and its declared channel and
+consumer; its account is deterministically produced by percent-encoding each
+`{channel}:{consumer}:{secret}` component before colon-joining them. Never paste a value into a shell
+history, repository file, issue, or receipt. HSP-02 will own non-interactive retrieval and temporary
+runtime material. A raw-store key is generated only when the governed preflight proves no encrypted
+records require an existing key; rotation is a separate migration, never a bootstrap retry.
```
RasmusTho, line 36: **<sub><sub>P2 Badge</sub></sub>  Mark HSP-01 delivered in its source-anchored spec**

This PR implements the contract described here, but `DEFINE_HOST_SECRET_CONTRACT.md` still declares `State: Authored task specification (future-state; child issue not yet filed)`. Issue #3845 exists and this PR closes it, so after merge the owner doc would simultaneously describe the slice as unfiled/future and deliver its implementation. Update the task state/task-order wording to record HSP-01 as delivered by #3845 / PR #3888 while leaving HSP-02 as future work.

app/ops/host_secret_contract.py
RasmusTho: Fixed by 99af070b: `object_pairs_hook` now rejects duplicate keys at every JSON object, with a regression test for hidden key material.

docs/LOCAL_SECRET_PROVISIONING/DEFINE_HOST_SECRET_CONTRACT.md
RasmusTho: Fixed by 153375c8: HSP-01 is marked implemented/delivered by #3845 / PR #3888 and its completed criteria are checked; HSP-02 remains future work.

app/ops/host_secret_contract.py
```diff
@@ -0,0 +1,91 @@
...
+@dataclass(frozen=True)
+class HostSecretContract:
+    keychain_service: str
+    keychain_account_template: str
+    allowed: frozenset[tuple[str, str, str]]
+
+    def require_declared(self, *, channel: str, consumer: str, secret: str) -> None:
+        if (channel, consumer, secret) not in self.allowed:
+            raise UndeclaredSecretConsumerError(
+                f"undeclared host secret request: channel={channel!r}, consumer={consumer!r}, secret={secret!r}"
```
RasmusTho: **[P1] Do not echo caller-supplied values in the undeclared-request error**

`require_declared` accepts arbitrary tuple strings, but its rejection embeds all three verbatim in the exception. A future HSP-02 failure path that logs this exception can therefore write attempted key material to command output or a receipt (for example, a misconfigured caller passes a raw key as `secret`), contradicting the non-disclosure invariant and this class’s “without secret material” contract. Emit only a fixed/redacted message (or known-safe declared identifiers) and add a regression asserting rejected caller values never appear in the exception.

RasmusTho: Fixed in 7f7e4b92: the rejection now uses a fixed redacted message, with a regression test asserting all caller-supplied identifiers are absent.

PR #3931, opened 2026-07-16
YSS-01: durable YouTube source registry + settings model

RasmusTho:
## Change Lane
- [x] Implementation lane
- [ ] Docs authoring lane
- [ ] Governance lane

Governing-Issue: #3916

## Linked Issue
Closes #3916

## Summary
- Guards every static vault-shared legacy scaffold before file I/O, including non-YSS seeds such as `settings/paths.md`; safe mode leaves the missing file absent and healthy mode permits the scaffold.
- Routes direct edits of runtime-gating settings through the single WriteGuard + durable `SettingsWriteReceipt` seam, including `youtubeSync.enabled`; first observation/state loss is evaluated from the registered default baseline instead of trusting disk.
- Enforces owner-file authority for runtime gates: `youtubeSync.enabled` is accepted only from `youtube.md`, while a misplaced `local.md` override is ignored loudly with no success receipt. Non-gating documented vault-local overrides remain unchanged.
- Validates and canonicalizes every present `account_binding_id` as UUID text before either memory or PostgreSQL, explicitly reserves `__none__`, and runs the shared UUID assertion against both backends.
- Validates provenance recursively as strict portable JSON before backend selection, rejecting Python-only values, non-string object keys, non-finite numbers, invalid timestamps, and unsupported fields without inserting a row.
- Pins `youtubeSync.subscriptionDefaultPolicy` to `VALID_ACQUISITION_MODES` with a production-constant parity test, avoiding runtime import-layer coupling.
- Makes the exact AC8 target create an isolated Alembic-migrated PostgreSQL database with autocreate disabled, run the shared backend contract, and assert Alembic downgrade raises for the forward-only migration.
- Retains the prior YSS-01 review repairs and truthful owner-doc slice state; YSS-02..YSS-11 and parent operator/live acceptance remain pending.

## Owner-Doc Writeback
- [x] `docs/YOUTUBE_SOURCE_SYNC/ESTABLISH_SOURCE_REGISTRY_AND_SETTINGS.md` — delivered repository-verifiable slice state; live acceptance pending.
- [x] `docs/YOUTUBE_SOURCE_SYNC/README.md` — YSS-01 delivered state and remaining child/parent boundary.
- [x] `docs/YOUTUBE_SOURCE_SYNC/SOURCE_SYNC_CONTRACT.md` — runtime-gating owner-file authority and rejected cross-file override semantics.
- [x] `docs/KNOWLEDGE_ACQUISITION/README.md` — Phase 4 row distinguishes delivered YSS-01 from pending YSS-02..YSS-11.
- [x] `companion-ui/docs/UI_RUNTIME_BOUNDARIES.md` and `docs/COMPANION_UI_PRODUCT_SPEC.md` — file-origin runtime gates truthfully describe watcher WriteGuard/receipt enforcement.
- `docs/DOCS_INDEX.md` was not changed: its ownership/index mapping remains accurate; no index entry moved or changed authority.

## Validation
- [x] After integrating `origin/main` `c8be43b1`: `pytest -q tests/vault/test_settings_service.py tests/settings/test_scoped_settings_service.py tests/vault/test_youtube_sync_settings.py tests/watcher/test_settings_delta_receipts.py tests/knowledge_acquisition/test_source_registry.py` — 37 passed.
- [x] Isolated disposable `pgvector/pgvector:pg16`: `pytest -q -m pg tests/knowledge_acquisition/test_source_registry_pg.py::test_pg_backend_contract tests/migrations/test_yss01_source_registry_schema_parity.py` — 2 passed; the AC target runs Alembic upgrade, the shared PG contract, and raising forward-only downgrade.
- [x] Raw `pytest -q -m "not pg"` — collection blocked only by the pre-existing missing `companion_ui` package root in `tests/api/test_companion_vault_settings.py`.
- [x] `PYTHONPATH=companion-ui/companion-app pytest -q tests/api/test_companion_vault_settings.py` — 10 passed.
- [x] `pytest -q -m "not pg" --ignore=tests/api/test_companion_vault_settings.py` — 9,738 passed, 46 skipped, 225 deselected, 18 xfailed.
- [x] `ruff check app tests` — All checks passed.
- [x] `mypy app` — Success: no issues found in 769 source files after main integration.
- [x] `git diff --check` — clean.
- [x] `DATABASE_URL=postgresql://app:app@127.0.0.1:1/app PYTHONPATH=. alembic upgrade head --sql` — offline SQL emission succeeds; `bd79f3044759` is the sole head with `down_revision=e1d2c3b4a5f6`.
- [x] `DOCS_GUARD_ALLOW_TEMPORAL_SKIP=1 python scripts/docs_guard.py` — OK.

## BuilderOps Routing
- Records/projections/receipts: LearningSignal `lrn_20260717100858_ad8e9a20`; epic run-state `yss-delivery-20260717` (coordinator-held).
- Reason: review-repair learning and coordinated YSS parent/child delivery state are routed through the existing BuilderOps records; this PR contains the associated repo-governed code, test, and owner-doc writeback.

## Residual Risk
- The raw local full-suite command still needs the repository Companion package root for one baseline-identical test; the bounded single-module run and single-file ignore preserve coverage without attributing that collection defect to YSS-01.
- The UUID CHECK is intentionally fail-loud for any non-contract historical account id. This table has not shipped before this PR; bootstrap, fresh migration, and legacy-repair producers are parity-tested.
- No production/runtime channel or deployment action was performed.


app/vault/manager.py
```diff
@@ -744,6 +744,24 @@ def _initial_settings_files(
             },
             "# Companion UI Settings\nShared Companion UI defaults for this vault.\nLocal UI preferences may override these in local settings.\n",
         ),
+        (
+            "youtube.md",
+            {
+                "schema": "design-handoff.youtube-sync.v1",
+                "scope": "vault-shared",
+                "youtubeSync.enabled": False,
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 988ba544. The static vault-shared initializer seeds now live in one source (`_static_shared_settings_seeds` in `app/vault/manager.py`), and `SettingsService.update_setting` scaffolds a missing static shared settings file from that seed before writing, so vaults initialized before YSS-01 can enable sync through the production seam. WriteGuard still decides before any file IO (a blocked gated write creates nothing), and `vault.md`/`local.md` (parameterized seeds) keep failing loudly. Regression coverage: `tests/vault/test_youtube_sync_settings.py::test_update_setting_scaffolds_missing_youtube_settings_file`.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,876 @@
...
+    if unknown:
+        raise SourceRegistryValidationError(f"acquisition_policy.media has unknown keys: {sorted(unknown)}")
+    if "enabled" in media and not isinstance(media["enabled"], bool):
+        raise SourceRegistryValidationError("acquisition_policy.media.enabled must be a boolean")
+    for key in ("max_quality", "format", "storage_binding"):
+        if key in media and media[key] is not None and not isinstance(media[key], str):
+            raise SourceRegistryValidationError(f"acquisition_policy.media.{key} must be a string or null")
+    for key in ("min_free_gb", "retention_days"):
+        if key in media and media[key] is not None and not isinstance(media[key], int | float):
+            raise SourceRegistryValidationError(f"acquisition_policy.media.{key} must be a number or null")
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 988ba544. `_validate_media_policy` now rejects bool and non-finite numbers (NaN/±Infinity) for `min_free_gb`/`retention_days` with `SourceRegistryValidationError`, keeping both backends identical and fail-loud. Covered for both backends via `tests/knowledge_acquisition/_source_registry_contract.py` (memory + pg suites).

RasmusTho (commented):
Independent 8-angle review fleet ran against the parallel (now-closed) #3934 implementation of this same contract; these are the findings that verified against **this PR's diff** (head 1a918777). Posting per the local review gate; from session d849a0e5 (the other #3916 claimant — claim reconciled, this PR governs).

**1. [enforcement gap] `youtubeSync.enabled` gating is documented but unenforced for direct `youtube.md` edits.**
This PR adds `youtubeSync.enabled` (vault-shared, `youtube.md`) to `RUNTIME_GATING_SETTINGS`, and `companion-ui/docs/UI_RUNTIME_BOUNDARIES.md` now lists it as "WriteGuard + `SettingsWriteReceipt`". That holds on the `update_setting` (API/CLI/UI) path — but the watcher's governed seam for *direct file edits*, `app/watcher/settings_delta.py::handle_settings_local_delta`, hard-returns for any `rel_path != settings/local.md` and is not extended by this PR. For the two pre-existing gating keys (local.md) an out-of-band file edit routes through WriteGuard + receipt; for the first vault-shared gating key it does not. Failure scenario: an Obsidian edit or git-synced change flips `youtubeSync.enabled=true` in `youtube.md` while WriteGuard has writes blocked — the sync master switch is reconfigured with no health-gate assertion and no receipt, so the documented control-action audit trail is silently absent exactly where the register claims it exists. Suggest: extend the delta seam to youtube.md gating keys in this PR, or scope the UI_RUNTIME_BOUNDARIES row to the API write path and file a bounded follow-up (YSS-06-adjacent, since the runner consumes this gate) — either way the doc and the enforcement should agree at merge.

**2. [drift risk] Acquisition-policy mode enum is duplicated across two enforcement points.**
The four modes are defined in `app/knowledge_acquisition/source_registry.py` (the frozenset near the top of the module) and re-listed literally as `allowed_values` on `youtubeSync.subscriptionDefaultPolicy` in `app/vault/settings_service.py`. A fifth mode (or rename) applied to one list splits the normative set (`SOURCE_SYNC_CONTRACT.md :: Acquisition policy`): settings accept a default policy the registry rejects, or vice versa. Import direction is awkward (vault↛KA), so a cheap governance test asserting the two sequences are equal would pin them together without new coupling.

**3. [validation] Number bounds accept non-integral floats.**
`_validate_value` accepts `int | float` then range-checks, so `youtubeSync.maxConcurrentAcquisitions: 2.5` or `inboxPollSeconds: 90.5` validate. Downstream consumers (concurrency width, poll sleep in YSS-06) expect ints; a fractional value either TypeErrors at runtime or silently mis-sizes. Suggest rejecting non-integral values for these keys (an `integer` flag on `SettingDefinition`, or `float(value).is_integer()` when bounds are ints).

**4. [minor, test-only backend] Memory store returns shared mutable references.**
`_Memory*.get()` returns the stored binding object; its nested `cursor`/`acquisition_policy`/`provenance` dicts are caller-mutable in place without the lock (pg deserializes fresh dicts per fetch), so the two backends diverge under caller mutation. The store is test-only, but the module's identical-behavior promise is the point of the dual backend; copying nested dicts on insert/return closes it cheaply.

No findings against: migration reversibility marker + head chaining, WL/HL refusal, single-inbox swap semantics, guard-census updates (db-import allowlist, write-seam machinery, vault_manager expected-set, settings-delta filter), INV-YSS-5/9 (no secrets/personal identifiers in fixtures).

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,876 @@
...
+                    and row.collection_ref == binding.collection_ref
+                    and row.account_binding_id == binding.account_binding_id
+                ):
+                    raise DuplicateBindingError(_duplicate_message(binding))
+            self._rows[binding.binding_id] = binding
+            return binding
+
+    def get(self, binding_id: str) -> SourceBinding | None:
+        with self._lock:
+            return self._rows.get(binding_id)
```
RasmusTho: **[P2] Do not expose the memory backend's stored mutable JSON by reference.**

`SourceBinding` is only shallowly frozen: callers can mutate `get(...).cursor`, `.acquisition_policy`, or `.provenance`, which mutates the memory backend's stored row outside the lock and bypasses validation. The Postgres backend deserializes fresh JSON on fetch, so the declared identical-backend contract diverges. Deep-copy nested JSON on insert and on every memory-backend return/list path, with a shared-contract regression test.

app/vault/manager.py
```diff
@@ -744,6 +744,24 @@ def _initial_settings_files(
             },
             "# Companion UI Settings\nShared Companion UI defaults for this vault.\nLocal UI preferences may override these in local settings.\n",
         ),
+        (
+            "youtube.md",
```
RasmusTho, line 783: **[P1] Restore the existing-vault scaffold before enabling this setting.**

This seed is only used during new-vault initialization. An already-initialized vault normally has no `settings/youtube.md`; on the current head, a healthy production `update_setting(..., "youtubeSync.enabled", True)` still raises `SettingsWriteError` because the destination file is absent, so setup cannot enable sync. The existing thread says the static-seed/ensure repair and a regression test were added, but neither is present at `9ebf931`. After the WriteGuard check (so blocked writes create nothing), seed this one static shared file or add an equivalent existing-vault migration, and cover that production call path.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,876 @@
...
+    unknown = set(media) - _ALLOWED_MEDIA_KEYS
+    if unknown:
+        raise SourceRegistryValidationError(f"acquisition_policy.media has unknown keys: {sorted(unknown)}")
+    if "enabled" in media and not isinstance(media["enabled"], bool):
+        raise SourceRegistryValidationError("acquisition_policy.media.enabled must be a boolean")
+    for key in ("max_quality", "format", "storage_binding"):
+        if key in media and media[key] is not None and not isinstance(media[key], str):
+            raise SourceRegistryValidationError(f"acquisition_policy.media.{key} must be a string or null")
+    for key in ("min_free_gb", "retention_days"):
+        if key in media and media[key] is not None and not isinstance(media[key], int | float):
```
RasmusTho: **[P2] Reject non-finite and boolean media numbers to preserve backend parity.**

`bool` is an `int`, and `float("nan")`/infinity satisfy this check. The memory backend therefore stores those policies, while the Postgres path serializes JSON that `jsonb` rejects, exposing a raw database failure instead of the promised `SourceRegistryValidationError`. This is also the previously reported P2 whose claimed repair is absent from `9ebf931`. Exclude booleans and require `math.isfinite()` for both fields in the shared validation suite.

app/vault/settings_service.py
```diff
@@ -567,6 +711,10 @@ def _validate_value(definition: SettingDefinition, value: Any) -> _ValidationRes
         return False, f"{definition.key} must be a boolean"
     if definition.type == "number":
         if isinstance(value, int | float) and not isinstance(value, bool):
```
RasmusTho: **[P2] Make bounded YouTube settings finite integers, not arbitrary floats.**

The new bounds accept `90.5` for poll seconds and `2.5` for `maxConcurrentAcquisitions`; `NaN` also passes both comparisons because neither `<` nor `>` is true. Those keys model seconds/days/concurrency and the source contract requires integer intervals, so the value later reaches scheduler/concurrency consumers outside their contract rather than degrading to the documented default with `SettingsValidationError`. Add an integer/finite constraint to these definitions (or an equivalent typed validator) and exercise fractional plus `NaN` values through resolution and `update_setting`.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in d425be7d (PR head d425be7d). The memory backend now deep-copies nested cursor, last_error, acquisition_policy, and provenance at insert plus every get/list/update return, so caller mutation cannot alter stored state. Shared contract coverage exercises insert/get/list/update isolation on memory and PG.

app/vault/manager.py
RasmusTho: Fixed in d425be7d (PR head d425be7d). After the production WriteGuard check, SettingsService scaffolds a missing static shared seed such as settings/youtube.md for an existing vault; blocked gated writes create nothing. Parameterized vault.md and local.md remain fail-loud. Covered by test_update_setting_scaffolds_missing_youtube_settings_file.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in d425be7d (PR head d425be7d). Media min_free_gb and retention_days now reject bool, NaN, and ±Infinity at shared validation time with SourceRegistryValidationError, before either backend persists JSON. The shared memory/PG contract suite covers every invalid value.

app/vault/settings_service.py
RasmusTho: Fixed in d425be7d (PR head d425be7d). All bounded youtubeSync numeric settings now require finite Python integers (never bool, NaN/∞, or fractions) at both SettingsService.resolve and the production update_setting seam. Parameterized production-call-site tests cover every bounded key.

app/vault/settings_service.py
```diff
@@ -382,6 +533,24 @@ def update_setting(
...
+
+        This runs only after runtime-gating WriteGuard evaluation. Files with
+        vault- or machine-specific seeds stay absent and fail loudly.
+        """
+        seed = shared_settings_file_seed(filename)
+        if seed is None:
+            raise SettingsWriteError(f"settings file does not exist: {path}") from cause
+        seed_frontmatter, seed_body = seed
+        try:
+            self.markdown_store.write_frontmatter(path, seed_frontmatter, body=seed_body)
```
RasmusTho, line 567: [P1] This scaffold is not actually guarded for every path it can take. `update_setting()` only calls `DEFAULT_WRITE_GUARD.assert_writes_allowed(...)` when the key is in `RUNTIME_GATING_SETTINGS`; a write of non-gating `youtubeSync.captionsEnabled` reaches this helper with no assertion (and the new test intentionally proves that it creates `youtube.md`). Thus a durable vault write can still happen in safe mode, while the new guard-census entry says this site is guarded. Gate the scaffold itself (or otherwise make the census/contract truthful) and add the blocked non-gating-scaffold case.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,905 @@
...
+def _assert_pg_schema(conn: Any) -> None:
+    cur = conn.cursor()
+    cur.execute("SELECT to_regclass(%s)", (_TABLE,))
+    row = cur.fetchone()
+    oid = row[0] if row else None
+    if not oid:
+        raise SourceRegistrySchemaMissingError(f"Missing table '{_TABLE}'. {_MIGRATION_HINT}")
+
+
+def _bootstrap_pg(conn: Any) -> None:
```
RasmusTho, line 649: [P1] Keep the test bootstrap schema identical to the migration schema. With `STORE_SCHEMA_AUTOCREATE=1`, this producer creates only the table plus two unique indexes; it omits the migration’s four CHECK constraints and `acquisition_source_registry_account_idx`. The pg contract test therefore exercises a less constrained schema than a migrated runtime, and there is no migration/bootstrap parity test like the established store/outbox tests. Reuse the migration DDL or add the missing constraints/index and a parity test.

app/vault/settings_service.py
RasmusTho: Fixed in ab052556d403df61fbda373f3c3f6f5b24dae529. Missing settings/youtube.md scaffolding now performs the same WriteGuard check for every youtubeSync.* key, including captionsEnabled; the product call-site test proves blocked gated and non-gated writes leave the file absent, a healthy guard scaffolds it, and an existing non-gating file still updates in safe mode.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in ab052556d403df61fbda373f3c3f6f5b24dae529. Bootstrap now declares the four named CHECK constraints and account index. The migration also repairs an existing old bootstrap table, and the new isolated-PG parity test compares fresh migration vs bootstrap plus that legacy-repair path.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,924 @@
...
+        version = override["policy_version"]
+        if not isinstance(version, int) or isinstance(version, bool) or version < 1:
+            raise SourceRegistryValidationError("acquisition_policy.policy_version must be a positive int")
+    if "extractor_ids" in override:
+        extractor_ids = override["extractor_ids"]
+        if not isinstance(extractor_ids, list) or not all(isinstance(x, str) for x in extractor_ids):
+            raise SourceRegistryValidationError("acquisition_policy.extractor_ids must be a list of strings")
+    if "captions" in override and not isinstance(override["captions"], bool):
+        raise SourceRegistryValidationError("acquisition_policy.captions must be a boolean")
+    if "media" in override and override["media"] is not None:
```
RasmusTho: **[P2] Reject an explicit null media policy instead of silently defaulting it.**

`media` is the contract’s required object-shaped policy field, but this guard skips validation for `{"media": null}`. `_build_acquisition_policy()` then silently retains the default media object, so an explicitly invalid write is accepted instead of raising `SourceRegistryValidationError`, contrary to YSS-01’s fail-loud policy-validation requirement. Require `media` to be an object whenever the key is present, and add the case to the shared memory/PG contract suite.

RasmusTho: Fixed in cadfd068. An explicitly present `acquisition_policy.media` now always flows through the object validator, so `null` fails loud while an omitted `media` key still receives the default. The shared memory/PG contract assertion covers both null and a non-object list through `SourceRegistry.register`.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,924 @@
...
+            return cls(_PgSourceRegistryBackend())
+        return cls(_MEMORY_REGISTRY)
+
+    def register(
+        self,
+        *,
+        collection_kind: str,
+        collection_ref: str,
+        title: str,
+        account_binding_id: str | None = None,
```
RasmusTho, line 973: **[P2] Reject account-less authenticated playlist bindings.**

The source-registry contract says `account_binding_id` is nullable only because RSS/public sources need none. But `register()` defaults it to `None` for every kind, so an `inbox_playlist` can be registered and activated via `set_inbox(None, binding_id)`. That creates an enabled inbox with no OAuth-account identity for YSS-02 to bind or for later per-account state to isolate. Reject a missing account binding for `inbox_playlist`, `owned_playlist`, and `liked_videos` (and cover the rule in the shared memory/PG contract suite); keep the nullable path only for the unauthenticated kinds.

RasmusTho: Fixed in 8c429582. SourceRegistry.register now rejects None for inbox_playlist, owned_playlist, and liked_videos, while public_playlist and subscription_feed remain account-less. The same shared contract runs through memory and PG. Because this cross-field invariant is directly expressible, the named CHECK is also present in bootstrap and Alembic existing-table repair; the migration/bootstrap/legacy parity test verifies all producers.

app/vault/settings_service.py
```diff
@@ -382,6 +533,36 @@ def update_setting(
...
+    ) -> MarkdownSettingsDocument:
+        """Seed a static shared settings file missing from an older vault.
+
+        Any ``youtubeSync.*`` scaffold is itself a vault-writing control
+        action. It therefore checks WriteGuard even when the requested key is
+        not normally runtime-gated; an existing non-gating setting file still
+        follows the ordinary non-gated update path. Files with vault- or
+        machine-specific seeds stay absent and fail loudly.
+        """
+        seed = shared_settings_file_seed(filename)
```
RasmusTho, line 556: **[P1] Keep the existing-vault scaffold scoped or guard every seed.** `shared_settings_file_seed(filename)` now returns seeds for `paths.md`, `workflow.md`, `design-handoff.md`, and `companion-ui.md` as well as `youtube.md`, but the guard below runs only when `key.startswith("youtubeSync.")`. On this exact head I exercised `update_setting(..., "handoffFolder", "Changed")` against a selected older vault with no `paths.md` while WriteGuard reported `safe_mode`; it created `settings/paths.md` and returned a non-gating receipt. That is an out-of-scope new scaffold-on-write path which bypasses safe mode, while the closed census classifies line 561 as guarded. Restrict this helper to the YSS-owned `youtube.md` migration, or assert WriteGuard for every seed this call site can write and add a blocked non-YSS regression/census classification.

app/vault/settings_service.py
```diff
@@ -30,16 +31,21 @@
...
-# the watcher/indexing runtime runs (registry.py:734, config.py:92).
+# the watcher/indexing runtime runs (registry.py:734, config.py:92), or
+# whether the YouTube Source Sync capability/runner is active
+# (docs/YOUTUBE_SOURCE_SYNC/SOURCE_SYNC_CONTRACT.md :: Settings model).
 # These must route through the governed write seam (WriteGuard + actor receipt).
 # See docs/COMPANION_UI_PRODUCT_SPEC.md :: Runtime control actions and
 # companion-ui/docs/UI_RUNTIME_BOUNDARIES.md :: Control-action register.
-RUNTIME_GATING_SETTINGS: frozenset[str] = frozenset({"enableVaultWatcher", "enableAutoIndexing"})
+RUNTIME_GATING_SETTINGS: frozenset[str] = frozenset(
+    {"enableVaultWatcher", "enableAutoIndexing", "youtubeSync.enabled", "youtubeSync.runnerEnabled"}
```
RasmusTho, line 47: **[P1] Route the vault-shared master switch through the file-origin gate.** Adding `youtubeSync.enabled` here makes the service/API path guarded, but `app/watcher/settings_delta.py::handle_settings_local_delta` returns immediately for every path except `settings/local.md`. A direct Obsidian/Git edit to the owning `settings/youtube.md` therefore becomes the effective master switch with neither WriteGuard evaluation nor `SettingsWriteReceipt`; the existing test only calls `SettingsService.update_setting` and cannot catch this production-origin gap. Extend the delta seam/state tracking to `settings/youtube.md` (or make the supported-origin/docs contract explicitly exclude that editable path) and cover a safe-mode file delta.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,938 @@
...
+        """
+        if not isinstance(collection_ref, str) or not collection_ref.strip():
+            raise SourceRegistryValidationError("collection_ref must be a non-empty string")
+        if collection_ref in UNSUPPORTED_COLLECTION_REFS:
+            raise SourceUnsupportedError(collection_ref)
+        if collection_kind not in VALID_COLLECTION_KINDS:
+            raise SourceRegistryValidationError(
+                f"collection_kind must be one of {sorted(VALID_COLLECTION_KINDS)}, got {collection_kind!r}"
+            )
+        if collection_kind in _ACCOUNT_BOUND_COLLECTION_KINDS and account_binding_id is None:
```
RasmusTho: **[P2] Validate the account-binding UUID before the NULL sentinel reaches Postgres.** This checks only `None`, so authenticated rows accept `""`, `"not-a-uuid"`, and the index sentinel `"__none__"`, despite the normative field being a UUID. That also breaks the promised memory/PG parity: memory accepts a public binding `(kind, ref, None)` and another `(kind, ref, "__none__")`, while the PG `COALESCE(account_binding_id, "__none__")` unique index makes the second collide. Validate every non-null account id as the contract UUID (and use valid UUID fixtures), or replace the sentinel expression with `NULLS NOT DISTINCT`/an equivalent collision-free schema; add the case to the shared backend suite.

tests/knowledge_acquisition/test_source_registry_pg.py
```diff
@@ -0,0 +1,36 @@
...
+from app.knowledge_acquisition.source_registry import SourceRegistry
+from tests.knowledge_acquisition._source_registry_contract import ALL_CONTRACT_ASSERTIONS
+
+
+def _make_pg_registry() -> SourceRegistry:
+    return SourceRegistry.for_runtime()
+
+
+@pytest.mark.pg
+def test_pg_backend_contract() -> None:
```
RasmusTho: **[P2] Exercise the forward-only migration in the declared AC target.** Issue #3916 maps “Alembic migration creates the table forward-only (downgrade raises)” to this test, but this function only enables `STORE_SCHEMA_AUTOCREATE` and runs service assertions; it never invokes Alembic or `downgrade()`. The added schema-parity test covers upgrade/bootstrap shape, but it also never asserts the required raising downgrade. Add a concrete downgrade/fail-forward assertion (and point the Issue `Verify:` marker at the migration test if split) so the behavioral AC resolves on this exact head rather than relying on inspection.

app/vault/settings_service.py
RasmusTho: Fixed at head d5d146a2. `_scaffold_missing_settings_file` now asserts the WriteGuard for every static shared seed before any `write_frontmatter`, not only `youtubeSync.*`; `test_static_shared_settings_scaffold_is_guarded_for_legacy_paths_file` covers blocked/no-file and healthy/allowed `paths.md`, and the closed properties census is repinned to the guarded seam.

app/vault/settings_service.py
RasmusTho: Fixed at head d5d146a2. The watcher delta seam now owns both `settings/local.md` and `settings/youtube.md`, derives each file’s gating keys from `SETTING_DEFINITIONS` + `RUNTIME_GATING_SETTINGS`, retains the last accepted value on WriteGuard denial, and emits the durable file-origin receipt when allowed. `youtube.md` is also explicitly scoped so the new main settings compiler cannot reload around the blocked seam. Covered by `test_youtube_master_switch_file_delta_is_guarded_and_receipted` plus the control-path classifier test.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed at head d5d146a2. Every present `account_binding_id` is validated and canonicalized with `uuid.UUID` before memory/PG selection; `__none__` is explicitly reserved. The canonical UUID CHECK is present in bootstrap, fresh Alembic DDL, and existing-table repair, and schema parity drops/rebuilds it. The shared memory+PG contract now covers invalid text, the sentinel, normalized valid UUID, required authenticated ids, and allowed `None` source kinds.

tests/knowledge_acquisition/test_source_registry_pg.py
RasmusTho: Fixed at head d5d146a2. The exact AC target `test_pg_backend_contract` now creates an isolated scratch PostgreSQL database, runs Alembic upgrade to `bd79f3044759` with schema autocreate disabled, executes the shared PG service contract, and calls Alembic downgrade to `e1d2c3b4a5f6` under `pytest.raises(RuntimeError, match="forward-only")`. It passed together with migration/bootstrap/legacy schema parity against an isolated `pgvector:pg16` container (2 passed).

app/watcher/settings_delta.py
```diff
@@ -10,12 +10,23 @@
...
 )
 
 SETTINGS_LOCAL_REL = Path("settings/local.md")
+SETTINGS_YOUTUBE_REL = Path("settings/youtube.md")
+
+_RUNTIME_GATING_KEYS_BY_FILE: dict[Path, frozenset[str]] = {
+    rel_path: frozenset(
+        definition.key
+        for definition in SETTING_DEFINITIONS
+        if definition.key in RUNTIME_GATING_SETTINGS and definition.file == rel_path.name
```
RasmusTho, line 29: **[P1] Classify vault-local overrides of shared gating keys.** `SettingsService._source_can_set_definition` intentionally lets a vault-local source override vault-shared definitions, but this map includes a key only when its owning `definition.file` matches. Consequently, adding `youtubeSync.enabled: true` to `settings/local.md` is accepted by `SettingsService.resolve()` as a vault-local effective value while this handler sees no changed key, calls no WriteGuard, and emits no receipt. I reproduced that path under `safe_mode`: the delta returned zero receipts/errors and resolution returned `True` from `local.md`. Either derive the file-origin key set from every scope the file can legally supply, or forbid this cross-scope override for the gating key, and cover the production watcher path.

app/watcher/settings_delta.py
```diff
@@ -144,7 +163,7 @@ def handle_settings_local_delta(
 
     current_values = {
         key: document.frontmatter[key]
-        for key in sorted(RUNTIME_GATING_SETTINGS)
+        for key in sorted(gating_keys)
         if key in document.frontmatter
     }
     if previous_values is None:
```
RasmusTho, line 192: **[P1] Do not trust first-seen gating values without the gate.** On a first watcher scan (or lost/corrupt state), `previous_values` is `None`, so this returns the file value before vault validation, WriteGuard, or receipt emission. A pre-start direct edit to `youtubeSync.enabled: true` under `safe_mode` therefore becomes the saved accepted baseline with zero receipts/errors; I reproduced exactly that result. Treat first observation as a transition from a safe/default baseline and route it through the same governed seam so watcher restart/state loss cannot bypass the file-origin contract.

app/vault/settings_service.py
```diff
@@ -232,6 +241,148 @@ def get(self, key: str) -> SettingDefinition | None:
...
+        "enum",
+        "discover_only",
+        "Conservative default acquisition mode for newly discovered subscriptions.",
+        "vault-shared",
+        "commit",
+        True,
+        True,
+        True,
+        "youtube.md",
+        allowed_values=(
```
RasmusTho, line 348: **[P2] Pin the settings policy enum to the registry contract.** This remains the transferred prior-review finding from issue #3916: the four acquisition modes are duplicated here and in `source_registry.VALID_ACQUISITION_MODES`, with no test tying them together. A later add/rename on only one side lets settings accept a default policy that registry writes reject (or vice versa). Add a neutral/shared vocabulary or, without introducing the awkward vault→KA import, a contract test that asserts the two sets are identical.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,966 @@
...
+    if override is None:
+        return {"origin": "manual_add", "at": now, "detail": {}}
+    if not isinstance(override, dict):
+        raise SourceRegistryValidationError(f"provenance must be an object, got {override!r}")
+    origin = override.get("origin", "manual_add")
+    if origin not in VALID_PROVENANCE_ORIGINS:
+        raise SourceRegistryValidationError(
+            f"provenance.origin must be one of {sorted(VALID_PROVENANCE_ORIGINS)}, got {origin!r}"
+        )
+    detail = override.get("detail", {})
```
RasmusTho, line 442: **[P2] Validate provenance as strict portable JSON before backend selection.** Only the outer `detail` type is checked and `at` is passed through untyped. The memory backend accepts values PostgreSQL cannot persist; I reproduced successful memory registration with `detail={"nested": object()}` and with `NaN`, whereas the PG path fails in `json.dumps`/`jsonb` instead of raising `SourceRegistryValidationError`. Validate/canonicalize the complete provenance payload with strict JSON semantics (`allow_nan=False`) at the service boundary and add the case to the shared memory/PG contract suite.

tests/knowledge_acquisition/test_source_registry.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_single_enabled_inbox_enforced_and_swap_atomic() -> None:
+    assert_single_enabled_inbox_and_swap(_make_registry)
+
+
+def test_duplicate_binding_refused() -> None:
+    assert_duplicate_binding_refused(_make_registry)
+
+
+def test_account_binding_nullability_matches_collection_kind() -> None:
```
RasmusTho, line 64: **[P2] Run the UUID/sentinel contract against memory too.** The new `assert_account_binding_uuid_contract` is in `ALL_CONTRACT_ASSERTIONS`, so the PG target executes it, but this memory module neither imports nor invokes it. That contradicts the helper's identical-suite claim and the PR body's statement that invalid text, `__none__`, and canonicalization are covered on both memory and PG. Add the memory invocation (or parameterize this module over `ALL_CONTRACT_ASSERTIONS`) so the strongest repair has the promised two-backend evidence.

RasmusTho (commented):
Local review gate — round A on exact head `d5d146a21e2494dc8c57ad87bc532870f9cc98a8` vs base `9cfc5333accbad78692b366776869e9b9be9c806`: **blocking findings; not clean**. Five findings are inline. Lifecycle audit also found that all 14 prior threads have replies, but 10/14 remain `isResolved=false`; those must be resolved before closure. No clean-streak credit.

app/watcher/settings_delta.py
RasmusTho: Fixed in 895f2864. Runtime-gating definitions now require their registered owner file during SettingsService resolution even when vault-local precedence would otherwise permit a shared-key override. The watcher filters stale cross-file state, surfaces the ownership validation error, keeps youtubeSync.enabled at false, and emits no success receipt. Production coverage: tests/watcher/test_settings_delta_receipts.py::test_local_file_cannot_override_youtube_master_switch.

app/watcher/settings_delta.py
RasmusTho: Fixed in 895f2864. A missing watcher state now establishes accepted previous values from each registered default and then runs the ordinary changed-key path, so a first-seen true activation must pass WriteGuard and emit the durable file receipt. An invalid/unselected vault retains that baseline rather than seeding state from disk. Production coverage: tests/watcher/test_settings_delta_receipts.py::test_first_seen_youtube_activation_is_guarded_and_receipted proves both safe-mode refusal/no receipt and healthy acceptance/receipt.

app/vault/settings_service.py
RasmusTho: Fixed in 895f2864 with the review-suggested contract-parity route, avoiding a vault-to-knowledge-acquisition runtime dependency. tests/vault/test_youtube_sync_settings.py::test_subscription_default_policy_matches_registry_contract imports the production SettingDefinition and VALID_ACQUISITION_MODES and pins exact set equality; existing production validators then enforce that shared vocabulary on each side.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in 895f2864. SourceRegistry now validates provenance recursively before backend selection, rejects Python-only containers/objects, non-string object keys, non-finite numbers, circular values, unsupported top-level fields, and invalid at values, then JSON-round-trips to a detached portable value with allow_nan=False. assert_provenance_is_strict_portable_json is part of ALL_CONTRACT_ASSERTIONS and proves identical refusal/no-row behavior for memory and PostgreSQL.

tests/knowledge_acquisition/test_source_registry.py
RasmusTho: Fixed in 895f2864. The memory suite now has the explicit test_account_binding_uuid_contract_memory target and invokes the same assert_account_binding_uuid_contract helper used by PostgreSQL through ALL_CONTRACT_ASSERTIONS. The focused memory suite and isolated Alembic-migrated PostgreSQL suite both pass.

app/watcher/settings_delta.py
```diff
@@ -130,9 +142,16 @@ def handle_settings_local_delta(
     settings_service: SettingsService | None = None,
     markdown_store: MarkdownSettingsStore | None = None,
 ) -> SettingsDeltaResult:
-    """Route watcher-detected runtime-gating settings deltas through the governed seam."""
+    """Route watcher-detected runtime-gating file deltas through the governed seam.
+
+    The historical function name is retained for callers, but the seam covers
+    both the vault-local runtime controls and the vault-shared YouTube master
+    switch. A blocked edit remains on disk as human-authored input while the
+    returned accepted runtime values stay at the last guarded state.
```
RasmusTho: **[P1] Keep denied file-origin gates out of effective resolution.** On a safe-mode edit from `youtubeSync.enabled: false` to `true`, this function correctly returns `{"youtubeSync.enabled": false}`, no receipt, and a blocked error, but it intentionally leaves `youtube.md` at `true`. The accepted map is only persisted in watcher state; `SettingsService.resolve(context)` rereads `youtube.md` and immediately returns the denied value as effective `true`. That means this gate does not control the settings source of truth described by the contract, and any consumer that resolves settings can bypass the denial. Wire the last accepted guarded value into effective resolution/runtime consumption (or restore/quarantine the denied disk value), and add a production-path assertion that resolve after a blocked file delta remains `false`.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,1019 @@
...
+    if not isinstance(detail, dict):
+        raise SourceRegistryValidationError("provenance.detail must be an object")
+    at = override.get("at", now)
+    if not isinstance(at, str) or not at.strip():
```
RasmusTho: **[P2] Validate `provenance.at` as the claimed timestamp.** This guard accepts every non-empty string, so `SourceRegistry.register(..., provenance={"at": "not-a-timestamp"})` succeeds in memory and PostgreSQL-facing validation even though `at` is the provenance timestamp and the PR body says invalid timestamps are rejected. Temporal lineage cannot be ordered or audited reliably with this value. Parse and canonicalize a supported RFC 3339/ISO-8601 timestamp (including the intended timezone rule) before backend selection, and add the invalid-string case to the shared memory/PG contract suite.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,1019 @@
...
+    for key in ("max_quality", "format", "storage_binding"):
+        if key in media and media[key] is not None and not isinstance(media[key], str):
+            raise SourceRegistryValidationError(f"acquisition_policy.media.{key} must be a string or null")
+    for key in ("min_free_gb", "retention_days"):
+        if key in media and media[key] is not None:
+            number = media[key]
+            # ``bool`` passes ``isinstance(..., int)`` and NaN/infinity are
+            # accepted by memory but rejected by Postgres jsonb. Refuse all
+            # three before either backend sees the policy.
+            if isinstance(number, bool) or not isinstance(number, int | float) or not math.isfinite(number):
```
RasmusTho, line 268: **[P2] Reject negative media-policy bounds.** The finite-number check still accepts `min_free_gb=-1` and `retention_days=-1`; I reproduced both passing shared validation. These values are out-of-bounds for free-space protection and retention duration: a negative minimum makes the future guard vacuous, while negative retention has no valid lifecycle meaning. AC6 requires invalid/out-of-bounds policy to fail loudly before persistence. Enforce the contract lower bounds here and cover both fields in the shared memory/PG contract suite.

app/knowledge_acquisition/source_registry.py
```diff
@@ -0,0 +1,1019 @@
...
+        merged_media.update(override["media"])
+        merged["media"] = merged_media
+    return merged
+
+
+def _portable_json_copy(value: Any, *, field: str) -> Any:
+    """Reject Python-only/non-finite values and return a detached JSON value."""
+
+    def _validate(item: Any, path: str, active: set[int]) -> None:
+        if item is None or isinstance(item, str | bool | int):
```
RasmusTho: **[P2] Reject JSON strings that PostgreSQL `jsonb` cannot store.** The new portability validator treats every Python string as portable. For example, `detail={"text": "\\u0000"}` and an unpaired surrogate both pass this function and the memory backend, but PostgreSQL `jsonb` rejects Unicode NUL (and invalid surrogate escapes), exposing a raw backend failure and breaking the identical memory/PG contract. Validate JSON string contents at this shared boundary, then add these values to the backend-shared refusal/no-row assertions.

app/vault/manager.py
```diff
@@ -791,6 +779,56 @@ def _initial_settings_files(
...
```
RasmusTho, line 783: **[P2] Update the owning initial-vault-files contract for `youtube.md`.** Adding this seed changes the scaffold created for every new vault, but `docs/CONCEPTS/VAULT_AND_SETTINGS_CONTEXT.md` still gives the authoritative `<vault>/settings/` file list without `youtube.md`. That makes the owner document false immediately after this change and violates the same-change docs rule for behavior/contracts. Add `youtube.md` to that initial file list (and keep the settings-spine/YSS cross-link consistent).

app/watcher/settings_delta.py
RasmusTho: Fixed in c7eca778. Denied runtime-gating file edits are now reverted on disk to the last accepted state (`_revert_denied_gating_keys`): set transitions, delete transitions, and the first-seen baseline all restore the owner file, so `SettingsService.resolve` can never surface a denied value as effective — the gate now controls the settings source of truth. The revert is registered in the WriteGuard site census as the gate's enforcement of an already-denied write, the seam docstring documents the semantics, and `tests/watcher/test_settings_delta_receipts.py` covers both directions plus the resolve-visible file state.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in c7eca778. `provenance.at` is parsed as RFC 3339/ISO-8601 via `_canonicalize_provenance_at`: arbitrary strings and naive (offset-less) timestamps are refused with `SourceRegistryValidationError`, and accepted values are canonicalized to UTC (+00:00) before backend selection. Shared contract suite covers not-a-timestamp, naive, empty, and canonicalization from +02:00 and Z forms on both backends.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in c7eca778. Contract lower bounds enforced at the shared boundary: `min_free_gb < 0` and `retention_days < 1` are refused (`min_free_gb=0` = no floor stays valid; zero-day retention is refused — 'unset' is null). Suite covers -1/-0.5 on both fields, retention_days=0, and the inclusive boundary values on both backends.

app/knowledge_acquisition/source_registry.py
RasmusTho: Fixed in c7eca778. `_check_portable_string` refuses NUL (\u0000) and unpaired surrogates in every JSON string — values AND object keys — inside `_portable_json_copy`, so PostgreSQL-jsonb-unstorable strings fail identically on both backends at validation time with no row created. Suite covers NUL value, unpaired surrogate value, and NUL object key.

app/vault/manager.py
RasmusTho: Fixed in c7eca778. `docs/CONCEPTS/VAULT_AND_SETTINGS_CONTEXT.md :: Initial Vault Files` now lists `youtube.md` in the scaffold and documents the not-required posture (resolution falls back to built-in defaults; scaffold-on-write from the initializer seed) with the cross-link to `SOURCE_SYNC_CONTRACT :: Settings model`.

app/watcher/settings_delta.py
```diff
@@ -130,9 +154,26 @@ def handle_settings_local_delta(
     settings_service: SettingsService | None = None,
     markdown_store: MarkdownSettingsStore | None = None,
 ) -> SettingsDeltaResult:
-    """Route watcher-detected runtime-gating settings deltas through the governed seam."""
+    """Route watcher-detected runtime-gating file deltas through the governed seam.
+
+    The historical function name is retained for callers, but the seam covers
+    both the vault-local runtime controls and the vault-shared YouTube master
+    switch. A blocked edit remains on disk as human-authored input (the
```
RasmusTho, line 161: Round-B deferred finding (tracked in #3964, gates YSS-06/YSS-10): deleting the whole owner file (`settings/youtube.md`) still bypasses this seam — the watcher deletion paths only route `is_settings_source_path` files, so an accepted `enabled: true` silently falls back to default False with no receipt. Fail-safe direction today (deletion can only disarm, and the keys have no runtime consumer in this branch), which is why it defers rather than blocks. The same issue also covers the cross-machine replay semantics (spurious `actor='human'` receipts for git-pulled changes) and resolution-layer enforcement.

tests/migrations/test_yss01_source_registry_schema_parity.py
```diff
@@ -0,0 +1,158 @@
...
+from __future__ import annotations
+
+import json
+import uuid
+from pathlib import Path
+
+import psycopg
+import pytest
+
+pytestmark = pytest.mark.pg
```
RasmusTho, line 12: Round-B deferred finding (tracked in #3965): this parity test is pg-marked and currently runs in NO automated lane — PR lanes filter `not pg`, and the nightly pg lane pins three unrelated files. Until #3965 adds it to the nightly pg lane, bootstrap/migration DDL drift is only caught by a developer running pg-marked tests manually.

PR #3956, opened 2026-07-17
fix: make index rebuild and strict doctor converge

RasmusTho:
Governing-Issue: #3955

Fixes #3955

## Change Lane
- [x] Implementation
- [ ] Docs authoring lane
- [ ] Governance lane

## SBS Impact
- Primary subsystem: DRI
- Secondary subsystem(s): PDM, RCA, OEF, Builder System
- Write class: derived
- Authority impact: none
- Persistence impact: rebuildable Postgres vector-index diagnostics and explicit cleanup of a vector whose present source became non-indexable
- Derived/rebuildable impact: aligns rebuild, reconcile, retrieval aliases, and doctor on indexability and canonical source text
- Human knowledge impact: none
- Memory impact: none
- Retrieval/context impact: prevents false stale or missing-vector diagnostics after a complete rebuild
- Sync/deployment impact: path-scoped PR Postgres verification; runtime promotion remains governed separately
- External boundary impact: none
- New or changed contract: explicit reconcile purges only the derived vector for a present canonically non-indexable source; absent-source fallback remains
- Owner-doc impact: `docs/DB_SCHEMA.md :: store_vector_index`; `docs/EMBEDDINGS.md :: Indexing behavior and invariants`; `docs/OPERATIONS.md :: CLI commands and runtime verification`
- Transition debt impact: reduces
- Fitness rule impact: strengthens
- Boundary risk: medium; Product data-path behavior and its path-scoped Builder System PG gate ship together

## Owner-Doc Writeback
`docs/DB_SCHEMA.md :: store_vector_index` and `docs/EMBEDDINGS.md :: Indexing behavior and invariants` now record canonical exact-byte production and bounded atomic derived-vector purge semantics. `docs/OPERATIONS.md :: CLI commands and runtime verification` records the operator-facing effect while preserving source immutability and absent-source fallback.

## Summary
- Centralize indexable text selection and provenance hashing so rebuild, reconcile, and doctor use `content` → `text` → `raw_text` plus the same fixed-point AI-panel canonicalization, including notes that mix fenced and legacy panels.
- Refresh the derived retrieval payload from authoritative source data during reconcile and keep its text aliases bound to the exact embedded bytes.
- Exclude genuinely contentless and AI-panel-only durable objects from the aggregate unembedded-stall diagnosis while preserving fail-loud behavior for every canonically text-bearing object.
- Treat a whitespace-only remainder after fixed-point panel removal as non-indexable while preserving exact whitespace on meaningful text.
- Canonicalize once before every runtime provider call, bind vector `content`/`text` and provenance to those exact bytes, preserve raw authoritative source payloads, and reject noncanonical legacy precomputed vectors fail-closed.
- Select and purge a derived vector whenever its present authoritative source is non-indexable, even if the vector already has the primary identity and empty canonical hash; never mutate the source or republish old retrieval bytes, and preserve the derived-payload fallback for an absent source.
- Reclassify source presence/indexability under a Postgres row lock and conditionally purge in that same transaction, so concurrent source updates are embedded and concurrent deletions retain fallback.
- Add a path-scoped PR Postgres gate that owns the panel canonicalizer, real Postgres regressions, direct CI ownership, and keep the store-payload sink census aligned with the moved call sites.
- Register all three runtime vector-producer seams with the memory-retrieval test selector so Unit tests (not pg) fail closed only on genuinely unowned paths.

## Validation
- `ruff check app tests` — passed.
- `mypy app` — passed (768 source files).
- Focused non-Postgres suite — 52 passed, 8 deselected on integrated `main`.
- Focused real-Postgres suite — 10 passed, 4 deselected, including deterministic source-update and source-delete purge races.
- `pytest -q -m "not pg"` — 10,014 passed, 38 skipped, 242 deselected, 18 xfailed on integrated `main`.
- Selector/producer regression bundle — 112 passed, 8 deselected.
- `python scripts/docs_guard.py` — passed.

## Coordination
- Pickup mode: `github-label-only-fallback` because the isolated repository had no dispatcher DB.
- Claim evidence: GitHub comment `5002349878`; agent `codex-root`; branch `codex/issue-3955-index-doctor-convergence`.

## BuilderOps Routing
- Records/projections/receipts: LearningSignal `lrn_20260717111149_fa737940`; pickup receipt comment `5002349878`.
- Reason: Captures the BGE-M3 runbook divergence where a zero-error rebuild could not converge strict doctor until producer/verifier semantics were repaired.

## Notes
PROD and release-channel pins are not changed by this PR.

### DIFF
--- a/app/cli/index_rebuild.py
+++ b/app/cli/index_rebuild.py
@@ -625,8 +625,10 @@ def reconcile(
         if not text:
             if authoritative_source_present:
                 try:
-                    purged = index_store.purge_vectors(
-                        UUID(str(object_id)), view=DEFAULT_EMBEDDING_VIEW
+                    classification = (
+                        index_store.purge_vector_if_present_source_non_indexable(
+                            UUID(str(object_id)), view=DEFAULT_EMBEDDING_VIEW
+                        )
                     )
                 except Exception as purge_exc:
                     _record_failure(
@@ -640,12 +642,35 @@ def reconcile(
                         False,
                     )
                     continue
-                summary["purged_non_indexable"] = int(
-                    summary["purged_non_indexable"]
-                ) + int(purged)
-            else:
+
+                if classification.source_present and not classification.source_indexable:
+                    summary["purged_non_indexable"] = int(
+                        summary["purged_non_indexable"]
+                    ) + int(classification.purged)
+                    continue
+
+                if classification.source_present:
+                    # The source became indexable after the candidate read.
+                    # Reclassify and embed the locked-read payload instead of
+                    # deleting its derived row from stale evidence.
+                    source_payload = dict(classification.source_payload or {})
+                else:
+                    # The source disappeared after the candidate read.  The
+                    # present-source purge contract no longer applies; retain
+                    # and, where possible, reconcile from vector fallback.
+                    source_payload = vector_payload
+                domain_obj = domain_objects.DomainObject(
+                    uuid=str(object_id),
+                    kind=kind,
+                    source_ref=source_ref,
+                    payload=source_payload,
+                    created_at=datetime.now(timezone.utc),
+                )
+                text = _extract_text(source_payload)
+
+            if not text:
                 summary["skipped"] = int(summary["skipped"]) + 1
-            continue
+                continue
 
         # ``store_vector_index`` is a derived retrieval projection.  Keep both
         # compatibility aliases bound to the exact producer-selected source
--- a/app/cli/index_rebuild.py
+++ b/app/cli/index_rebuild.py
@@ -11,7 +11,11 @@ from uuid import UUID
 import click
 
 from app.components.embeddings import EmbeddingIdentity, get_embedding_client
-from app.index.artifact_metadata import build_indexed_unit_payload, compute_payload_content_hash, extract_indexable_text
+from app.index.artifact_metadata import (
+    build_indexed_unit_payload,
+    canonicalize_indexable_text,
+    compute_payload_content_hash,
+)
 from app.llm.embed_queue import EmbedDeadLetterError, embed_with_retry
 from app import objects as domain_objects
 from app.stores import get_vector_index, resolve_store_backend
@@ -38,7 +42,7 @@ _RETRYABLE_NAMES = {"SerializationFailure", "DeadlockDetected", "TooManyRequests
 
 
 def _extract_text(payload: dict | None) -> str:
-    return extract_indexable_text(payload)
+    return canonicalize_indexable_text(payload)
 
 
 def _load_objects(limit: int | None) -> Tuple[List[domain_objects.DomainObject], str]:
@@ -267,6 +271,10 @@ def rebuild(
             summary["skipped"] = int(summary["skipped"]) + 1
             continue
 
+        indexed_payload = dict(domain_obj.payload or {})
+        indexed_payload["content"] = text
+        indexed_payload["text"] = text
+
         embedding: list | None = None
         try:
             embedding = embed_with_retry(
@@ -315,7 +323,7 @@ def rebuild(
                     object_id=UUID(str(domain_obj.uuid)),
                     kind=str(domain_obj.kind or "note"),
                     source_ref=str(domain_obj.source_ref or ""),
-                    payload=dict(domain_obj.payload or {}),
+                    payload=indexed_payload,
                     text=text,
                     embedding_identity=identity,
                 ),
@@ -432,12 +440,12 @@ def _stale_content_hash_rows(limit: int | None) -> List[dict]:
     return stale
 
 
-def _reconcile_object_text(object_id, vector_payload: dict | None) -> str:
-    """Resolve the source text to re-embed for a row.
+def _reconcile_object_payload(object_id, vector_payload: dict | None) -> dict:
+    """Resolve the authoritative payload to re-embed for a row.
 
     Prefer the authoritative ``store_objects`` payload (the durable object of
-    record); fall back to the text carried in the vector-index row payload when the
-    object row is absent, so an orphaned index row can still be reconciled.
+    record); fall back to the vector-index row payload when the object row is
+    absent, so an orphaned index row can still be reconciled.
     """
     from app.stores.pg import _connect  # local import: pg backend is optional
 
@@ -449,10 +457,10 @@ def _reconcile_object_text(object_id, vector_payload: dict | None) -> str:
             )
             row = cur.fetchone()
     if row and row.get("payload"):
-        text = _extract_text(dict(row["payload"]))
-        if text:
-            return text
-    return _extract_text(vector_payload or {})
+        payload = dict(row["payload"])
+        if _extract_text(payload):
+            return payload
+    return dict(vector_payload or {})
 
 
 @index.command("reconcile", help="Re-embed non-primary-identity vectors under the primary identity (converge a mixed index).")
@@ -555,18 +563,27 @@ def reconcile(
         kind = str(row.get("kind") or "note")
         source_ref = str(row.get("source_ref") or "")
         vector_payload = dict(row.get("payload") or {})
+        source_payload = _reconcile_object_payload(object_id, vector_payload)
         domain_obj = domain_objects.DomainObject(
             uuid=str(object_id),
             kind=kind,
             source_ref=source_ref,
-            payload=vector_payload,
+            payload=source_payload,
             created_at=datetime.now(timezone.utc),
         )
 
-        text = _reconcile_object_text(object_id, vector_payload)
+        text = _extract_text(source_payload)
         if not text:
             summary["skipped"] = int(summary["skipped"]) + 1
             continue
+
+        # ``store_vector_index`` is a derived retrieval projection.  Keep both
+        # compatibility aliases bound to the exact producer-selected source
+        # text so a successful reconcile cannot leave retrieval serving a
+        # stale legacy ``text`` value while provenance already reports clean.
+        indexed_payload = dict(source_payload)
+        indexed_payload["content"] = text
+        indexed_payload["text"] = text
 
         embedding: list | None = None
         try:
@@ -588,7 +605,7 @@ def reconcile(
             _oid=object_id,
             _kind=kind,
             _source_ref=source_ref,
-            _payload=vector_payload,
+            _payload=indexed_payload,
             _text=text,
             _embedding=embedding,
         ) -> None:
--- a/app/cli/index_rebuild.py
+++ b/app/cli/index_rebuild.py
@@ -441,6 +441,42 @@ def _stale_content_hash_rows(limit: int | None) -> List[dict]:
     return stale
 
 
+def _present_non_indexable_rows(limit: int | None) -> List[dict]:
+    """Return derived rows whose present source now has no canonical text.
+
+    This candidate class is independent of identity and content hash.  Legacy
+    panel-bearing vectors can already carry the current primary identity and a
+    hash of the empty canonical body, so neither ordinary reconcile selector
+    can prove that their stale retrieval bytes must be purged.
+    """
+    from app.stores.pg import _connect  # local import: pg backend is optional
+
+    with _connect() as conn:
+        with conn.cursor() as cur:
+            cur.execute(
+                """
+                SELECT v.object_id AS object_id, v.kind AS kind,
+                       v.source_ref AS source_ref, v.payload AS payload,
+                       v.provider AS provider, v.model AS model,
+                       v.dim AS dim, v.normalize AS normalize,
+                       o.payload AS current_payload
+                FROM store_vector_index AS v
+                JOIN store_objects AS o ON o.object_id = v.object_id
+                ORDER BY v.object_id
+                """
+            )
+            candidates = list(cur.fetchall())
+
+    rows = [
+        row
+        for row in candidates
+        if not canonicalize_indexable_text(dict(row.get("current_payload") or {}))
+    ]
+    if limit is not None:
+        rows = rows[:limit]
+    return rows
+
+
 def _reconcile_object_payload(object_id, vector_payload: dict | None) -> Tuple[dict, bool]:
     """Resolve the authoritative payload to re-embed for a row.
 
@@ -505,6 +541,7 @@ def reconcile(
     summary: Dict[str, object] = {
         "backend": resolved_backend,
         "total_mismatched": 0,
+        "total_present_non_indexable": 0,
         "reconciled": 0,
         "purged_non_indexable": 0,
         "skipped": 0,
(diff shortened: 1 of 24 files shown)


app/index/artifact_metadata.py
```diff
@@ -49,6 +50,31 @@ def compute_content_hash(text: str) -> str:
...
+    for key in INDEXABLE_TEXT_KEYS:
+        value = data.get(key)
+        if value:
+            return str(value)
+    return ""
+
+
+def compute_indexed_content_hash(text: str) -> str:
+    """Hash text with the producer's AI-panel canonicalization."""
+    return compute_content_hash(strip_ai_panels(text or ""))
```
RasmusTho: [P1] Canonicalize the text before embedding, not only before hashing

`compute_indexed_content_hash()` removes AI panels, but both rebuild and reconcile still pass the unstripped value returned by `extract_indexable_text()` to `client.embed_text()`. A panel-bearing row therefore gets a vector for the full panel text and a provenance hash for different, panel-free text; later panel-only changes compare equal and doctor reports the now-stale vector clean. This contradicts `docs/DB_SCHEMA.md :: store_vector_index` (hash of the exact embedded text) and the Issue requirement that producer, reconcile, and doctor share canonicalization. Apply the same canonicalization to the actual rebuild/reconcile embedding input and add an assertion on the embedder input, not only on the stored hash.

scripts/select_pr_tests.py
```diff
@@ -316,6 +316,9 @@
             "app/memory/",
             "app/retrieval/",
             "app/index/",
+            # Index rebuild is the memory-retrieval CLI producer for the
+            # canonical vector rows exercised by the index suites.
+            "app/cli/index_rebuild.py",
```
RasmusTho, line 330: [P2] Route index-rebuild changes to their direct regression suites

This new ownership entry stops `app/cli/index_rebuild.py` from failing closed as unowned, but the `memory_retrieval` targets do not include either `tests/cli/test_index_rebuild_resilience.py` or `tests/index/test_provenance_stamp.py`. Running the selector with only this changed path produces `tests/agent_memory tests/retrieval tests/indexer tests/search`, so a future code-only rebuild/reconcile regression can now get a green Unit-tests check without either direct suite. Add the direct index/rebuild targets here and assert them in `test_index_rebuild_cli_change_selects_memory_retrieval_coverage`.

app/cli/index_rebuild.py
```diff
@@ -430,7 +424,8 @@ def _stale_content_hash_rows(limit: int | None) -> List[dict]:
     stale = [
         row
         for row in candidates
-        if compute_content_hash(row.get("current_text") or "") != row.get("stored_hash")
+        if compute_payload_content_hash(dict(row.get("current_payload") or {}))
+        != row.get("stored_hash")
```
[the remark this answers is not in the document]
RasmusTho: Fixed in 8f6eb5773e618c77a31c3d53e7774cc85d21e3ec. Reconcile now reloads the authoritative store_objects payload, projects both retrieval text aliases from the exact selected source text, and the real-Postgres regression asserts the refreshed payload. Focused PG: 5 passed; full not-pg: 9,650 passed.

scripts/select_pr_tests.py
RasmusTho: Fixed in 8f6eb5773e618c77a31c3d53e7774cc85d21e3ec. memory_retrieval CI ownership now includes tests/index plus the direct rebuild-resilience and mixed-identity doctor suites; selector regression is green.

app/index/artifact_metadata.py
RasmusTho: Fixed in 8f6eb5773e618c77a31c3d53e7774cc85d21e3ec. Rebuild and reconcile now send the AI-panel-stripped canonical bytes to the embedder, stamp that same text, and persist matching retrieval aliases. Dedicated embedding-input and real-Postgres regressions are green.

app/index/doctor.py
```diff
@@ -348,6 +348,11 @@ def inspect_unembedded_pg_objects(*, limit: int = 5) -> tuple[int, list[str]]:
                  AND v.embedding IS NOT NULL
                  AND array_length(v.embedding, 1) > 0
                 WHERE v.object_id IS NULL
+                  AND (
+                    COALESCE(o.payload->>'content', '') <> ''
```
RasmusTho: [P1] Reuse the canonical indexability predicate here

`index rebuild` now calls `canonicalize_indexable_text()` before deciding whether to skip a row, but this SQL only checks whether the raw JSON strings are non-empty. A durable object whose selected `content` consists only of an AI panel canonicalizes to `""` and is skipped by rebuild, while this query still counts it as unembedded forever; strict rebuild → doctor therefore still cannot converge. Please apply the same canonical predicate to both the count and sample paths (and cover an AI-panel-only payload in the PG regression).

scripts/select_pr_tests.py
```diff
@@ -316,14 +316,26 @@
...
             "tests/indexer/",
             "tests/e2e/test_index_rules_e2e.py",
             "tests/e2e/test_promotion_intent_to_index.py",
             "tests/e2e/test_reality_mvp_pipeline.py",
         ),
-        ("tests/agent_memory", "tests/retrieval", "tests/indexer", "tests/search", *E2E_TARGETS["memory_retrieval"]),
+        (
+            "tests/agent_memory",
+            "tests/retrieval",
+            "tests/index",
```
RasmusTho, line 348: [P1] Attach the Postgres acceptance tests to PR CI

Adding `tests/index` to this target list does not execute the two new PG acceptance tests: every selector result is still run under `PR_MARKER_EXPRESSION`, which begins with `not pg`, and the live `Unit tests (not pg)` check confirms that lane. `test_doctor_uses_rebuild_text_precedence_for_content_hash` and `test_reconcile_incremental_on_stale_only` are both marked `pg`; the scheduled nightly is not a PR gate and does not include this provenance test file. Please add a PR-triggered PG owner/lane for these exact Verify targets (and a selector/workflow assertion proving they run), rather than only selecting their directory into a lane that deselects them.

app/cli/index_rebuild.py
```diff
@@ -454,10 +457,10 @@ def _reconcile_object_text(object_id, vector_payload: dict | None) -> str:
             )
             row = cur.fetchone()
     if row and row.get("payload"):
-        text = _extract_text(dict(row["payload"]))
-        if text:
-            return text
-    return _extract_text(vector_payload or {})
+        payload = dict(row["payload"])
+        if _extract_text(payload):
```
RasmusTho: [P1] Do not replace an authoritative contentless payload with stale vector data

The fallback is documented for an absent `store_objects` row, but it also runs when the authoritative row exists and canonicalizes to no text. For a contentless source that still has an older or identity-mismatched vector, reconcile will therefore re-embed the stale vector payload and republish its old `content`/`text`, rather than honoring the current authoritative source; retrieval can keep serving bytes the source no longer contains. Distinguish “row absent” from “row present but contentless” and handle the latter without falling back to the derived payload, with a regression for this transition.

RasmusTho: Fixed in 98409a64f3908113f6b43a7fa4f45070e27c0b37. Reconcile now distinguishes a present authoritative source from an absent one. A present canonically non-indexable source causes only its derived vector to be purged; source payload remains byte-identical, while absent sources retain fallback. The new real-PG AC is green.

app/index/doctor.py
RasmusTho: Fixed in 98409a64f3908113f6b43a7fa4f45070e27c0b37. Doctor now applies the shared canonical indexability predicate after its coarse SQL filter; the real-PG regression covers AI-panel-only and genuinely contentless rows while keeping text-bearing stalls fail-loud. Focused PG: 6 passed.

scripts/select_pr_tests.py
RasmusTho: Fixed in 98409a64f3908113f6b43a7fa4f45070e27c0b37. CI Smoke now has a path-scoped Index PG contracts job backed by pgvector/Postgres that runs both exact PG acceptance files on PRs. The workflow contract regression and local 6-test PG run are green.

tests/index/test_provenance_stamp.py
```diff
@@ -168,6 +170,51 @@ def test_doctor_lists_stale_candidates(tmp_path, monkeypatch) -> None:
...
+    reset_diagnose_cache()
+    try:
+        from app.search import service as search_service
+
+        canonical_content = "producer-selected content"
+        oid, _dim = search_service.ingest_object(
+            kind="note",
+            source_ref="unit-test://doctor-content-precedence",
+            payload={"content": canonical_content, "text": "legacy shadow text"},
+            text=canonical_content,
```
RasmusTho: [P2] Exercise the rebuild producer in this Verify target

This test creates the vector with `search_service.ingest_object(..., text=canonical_content)`, so it bypasses `app.cli.index_rebuild._extract_text` and hands the expected bytes to the writer explicitly. It therefore still passes if rebuild regresses to `text`-first selection, even though the governing AC specifically requires a vector rebuilt from divergent `content`/`text` not to be reported stale. Seed only the authoritative object, invoke `index rebuild`, then run doctor (and ideally assert the stored aliases/hash); that makes this named Verify target actually cover the producer precedence it gates.

RasmusTho: Fixed in 1c16e3354a643dcaeaed22c945a725f2b0bca862. The named PG Verify target now seeds only store_objects with divergent content/text, invokes the real index rebuild CLI, and asserts the persisted retrieval aliases, provenance hash, processed count, and clean doctor result. The exact six-test PG acceptance surface is green locally.

app/index/artifact_metadata.py
```diff
@@ -49,6 +50,36 @@ def compute_content_hash(text: str) -> str:
...
+
+
+def canonicalize_indexable_text(payload: dict | None) -> str:
+    """Return the exact producer text used for embedding and provenance."""
+    return strip_ai_panels(extract_indexable_text(payload))
+
+
+def compute_indexed_content_hash(text: str) -> str:
+    """Hash text with the producer's AI-panel canonicalization."""
+    return compute_content_hash(strip_ai_panels(text or ""))
```
RasmusTho: [P1] Hash the exact canonical bytes without canonicalizing them twice

`rebuild` and `reconcile` now pass the already-canonical result of `canonicalize_indexable_text()` into `build_indexed_unit_payload()`, but this helper strips AI panels again. `strip_ai_panels()` is not idempotent when a note contains both supported encodings: if any fenced panel exists, the first pass removes only fenced spans and leaves a legacy `## AI-...` panel; the second pass then removes that legacy panel. The embedder and persisted aliases therefore receive the once-stripped bytes while `content_hash` covers twice-stripped bytes. Doctor hashes the authoritative payload once, reports the freshly rebuilt row stale, and reconcile repeats the mismatch forever. Hash the already-embedded text directly (or make canonicalization reach a fixed point) and add a mixed fenced+legacy regression asserting embed input, aliases, stored hash, and doctor convergence.

RasmusTho: Fixed in 33543963 (current head 5db31e3b). strip_ai_panels now iterates until canonical text reaches a fixed point, so removing fenced panels cannot leave a newly exposed legacy heading panel. Added a direct mixed-format idempotence regression and extended the real-Postgres rebuild/doctor contract with mixed panels. Verified on current main: 6 PG tests passed, 9730 non-PG tests passed, ruff, mypy, and docs guard are green.

app/services/indexer.py
```diff
@@ -94,6 +94,7 @@ def handle_ingest_object_created(obj: Dict[str, object]) -> None:
         "raw_text": obj_payload.get("raw_text"),
         "text": content,
     }
+    canonical_content = canonicalize_indexable_text(payload)
```
RasmusTho: [P1] Keep an explicitly empty canonical body from falling through to raw_text

The real vault-watcher caller passes `content = strip_ai_panels(body).strip()` and also carries the immutable full-file `payload.raw_text`. For a panel-only/whitespace-only note, `content` and the derived `text` here are `""`, so `canonicalize_indexable_text(payload)` skips both and selects `raw_text`; after panel stripping that still contains YAML frontmatter, making `canonical_content` non-empty and causing this handler to embed/upsert metadata instead of taking the purge branch. The added panel-only unit test does not exercise this production shape because it leaves `content` as the raw panel block, which wins selection before canonicalizing to empty. Preserve the explicitly supplied canonical body (including empty) for vault-change events, and add a regression with empty `content` plus frontmatter-bearing `raw_text` through this handler/watcher seam.

app/indexer/consumer.py
```diff
@@ -114,6 +125,10 @@ def process_event(evt: Dict[str, Any]) -> None:
     obj_uuid = UUID(str(obj.uuid))
     obj_payload = dict(obj.payload or {})
     text = _extract_text(obj_payload)
+    idx = get_vector_index()
+    if not text:
+        _purge_vectors(idx, obj_uuid)
```
RasmusTho: [P1] Do not acknowledge a panel-only event when its purge failed

This is the only operation that removes an existing vector after the authoritative object becomes canonically empty, but `_purge_vectors()` catches every exception and returns `0`; this branch then returns normally, so the outbox worker acknowledges the request even when Postgres was unavailable and the stale vector remains retrievable. The worker already classifies propagated psycopg/connection errors as transient and keeps the row pending, so this return-only purge path needs fail-loud semantics (and a purge-raises regression) rather than the best-effort helper used before a replacing upsert. The same issue affects rejection of a noncanonical legacy precomputed event at line 73.

scripts/select_pr_tests.py
```diff
@@ -325,14 +325,33 @@
...
             "app/index/",
+            # Index rebuild is the memory-retrieval CLI producer for the
+            # canonical vector rows exercised by the index suites.
+            "app/cli/index_rebuild.py",
+            # Runtime vector producers share the same canonical-byte and
+            # provenance contract as rebuild/reconcile. Keep changes to these
+            # exact seams on the index/search acceptance surface instead of
+            # failing closed as unowned after implementation has already run.
+            "app/indexer/consumer.py",
+            "app/search/service.py",
```
RasmusTho, line 336: [P2] Select the service producer’s direct regression

`app/services/indexer.py` now resolves to `memory_retrieval`, but this target tuple does not include `tests/services/test_indexer_worker.py`, which is the governing Issue’s named Verify target for that producer. The new selector test only asserts `tests/index`, `tests/indexer`, and `tests/search`, so a future PR changing only `app/services/indexer.py` will be green without running `test_handle_ingest_object_created_preserves_source_and_indexes_canonical_bytes` or the panel-only service tests. This PR happens to run them because the test file itself is changed. Add the direct service test target and assert it in the selector regression.

app/services/indexer.py
RasmusTho: Fixed in 6cf2ca52: vault-change events now preserve their explicitly selected body, including empty, instead of falling through to frontmatter-bearing raw_text. Added a production-shape regression.

app/indexer/consumer.py
RasmusTho: Fixed in 6cf2ca52: purge-only rejection paths now require purge support and propagate purge failures so the outbox remains retryable. Added failure regressions for both current panel-only and legacy-precomputed paths.

scripts/select_pr_tests.py
RasmusTho: Fixed in 6cf2ca52: memory_retrieval now selects tests/services/test_indexer_worker.py, asserted by the runtime-producer selector regression.

PR #4099, opened 2026-07-24
Fix bounded capture options for CKM overview

RasmusTho:
Governing-Issue: #4098
Final-Review-Rounds: 2

Fixes #4098

## Summary
Add explicit positive finite capture-bound options to `ckm overview` and thread them through HTML rendering to the projection store. Existing default bounds and typed refusal behavior remain unchanged.

## Validation
- `python3 -m pytest -q tests/builderops/ckm/test_overview_html.py -k 'capture_bounds or capture_limit'` (2 passed)
- `python3 -m pytest -q tests/builderops/ckm/test_overview_html.py tests/builderops/ckm/test_query_plans.py` (31 passed)
- `python3 -m pytest -q tests/builderops/ckm --ignore=tests/builderops/ckm/test_postgres_integration.py` (237 passed)
- `ruff check app tests` (passed)
- `mypy app` (passed)
- Exact-head canonical Demerzel replay passed against `8f7f1e696`; overview artifact SHA-256 `af3eeca11de323aab262d45d54f301934ce9fe1d1f2946a6be15982e98dbfd28`; terminal receipt on #4089 comment 5072782036.

## BuilderOps Routing
- Records/projections/receipts: Parent #4089 terminal replay receipt (issue comment 5072782036); no new BuilderOps record created.
- Reason: This bounded repair completed the already-authorized real-store replay; the parent receipt now records canonical success and independent output checks.

---

### DIFF
--- a/app/builderops/cli.py
+++ b/app/builderops/cli.py
@@ -694,13 +694,35 @@ def ckm_project(ctx: click.Context, projection_type: str, output_dir: Path) -> N
     type=click.Path(dir_okay=False, path_type=Path),
     required=True,
 )
+@click.option(
+    "--class-capture-limit",
+    type=click.IntRange(min=1),
+    default=None,
+    help="Finite per-class projection capture bound (defaults to the store policy).",
+)
+@click.option(
+    "--aggregate-capture-limit",
+    type=click.IntRange(min=1),
+    default=None,
+    help="Finite aggregate projection capture bound (defaults to the store policy).",
+)
 @click.pass_context
-def ckm_overview(ctx: click.Context, output_path: Path) -> None:
+def ckm_overview(
+    ctx: click.Context,
+    output_path: Path,
+    class_capture_limit: int | None,
+    aggregate_capture_limit: int | None,
+) -> None:
     try:
         store = _ckm_store(ctx)
         if not store.db_path.is_file():
             raise CkmValidationError(f"CKM database does not exist: {store.db_path}")
-        result = write_overview_html(store, output_path)
+        result = write_overview_html(
+            store,
+            output_path,
+            class_capture_limit=class_capture_limit,
+            aggregate_capture_limit=aggregate_capture_limit,
+        )
     except (CkmValidationError, OSError, sqlite3.Error) as exc:
         raise click.ClickException(f"ckm overview failed: {exc}") from exc
     click.echo(result)
--- a/tests/builderops/ckm/test_overview_html.py
+++ b/tests/builderops/ckm/test_overview_html.py
@@ -9,7 +9,7 @@ from click.testing import CliRunner
 from app.builderops.cli import builderops
 from app.builderops.ckm.models import MATURITY_DIMENSIONS
 from app.builderops.ckm.overview_html import render_overview_html
-from app.builderops.ckm.store import CkmProjectionCaptureError, CkmStore
+from app.builderops.ckm.store import CkmStore
 
 
 @pytest.fixture()
@@ -113,6 +113,19 @@ def overview_store(tmp_path: Path) -> CkmStore:
     )
     store.set_watermark("fixture", "two")
     return store
+
+
+def _make_snapshot_exceed_default(store: CkmStore) -> None:
+    """Add enough real fixture artifacts to exceed the default class bound."""
+    existing = len(store.list_artifacts())
+    for index in range(existing, 501):
+        store.upsert_artifact(
+            source_ref=f"fixture/oversized-{index}.md",
+            artifact_kind="document",
+            source="fixture",
+            watermark="one",
+            provenance=f'{{"source_ref":"fixture/oversized-{index}.md"}}',
+        )
 
 
 @pytest.fixture()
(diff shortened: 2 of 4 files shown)


tests/builderops/ckm/test_overview_html.py
```diff
@@ -486,6 +486,105 @@ def test_cli_writes_overview(overview_store: CkmStore, tmp_path: Path) -> None:
...
+    result = CliRunner().invoke(
+        builderops,
+        [
+            "--db-path",
+            str(overview_store.db_path),
+            "ckm",
+            "overview",
+            "--out",
+            str(output),
+        ],
```
RasmusTho, line 556: [P1] Verify an explicitly supplied insufficient bound

This invocation supplies no capture-bound option and the monkeypatch refuses every call, so it only covers the omitted/default case. It cannot catch a regression where a too-small positive override is ignored, not forwarded, or silently accepted, despite the acceptance criterion requiring `snapshot_too_large` with no partial output for a supplied insufficient bound. Add an invocation with an explicit too-small positive class or aggregate limit against known counts. The adjacent non-positive test should also include a negative value because the issue contract names both zero and negative inputs.

tests/builderops/ckm/test_overview_html.py
```diff
@@ -486,6 +486,105 @@ def test_cli_writes_overview(overview_store: CkmStore, tmp_path: Path) -> None:
...
+    overview_store: CkmStore,
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    observed: dict[str, int] = {}
+    original = CkmStore.load_projection_batch
+
+    def traced(self: CkmStore, **kwargs: int):
+        observed.update(kwargs)
+        if kwargs.get("class_capture_limit") == 500:
```
RasmusTho: [P1] Exercise a snapshot that actually exceeds the defaults

The `overview_store` fixture is far below both 500-per-class and 3,000 aggregate, and this branch is unreachable in the invocation because the CLI passes `class_capture_limit=501`. The test therefore proves only argument forwarding; it would still pass if the default path could not render any realistically oversized snapshot. Issue #4098 explicitly requires this Verify target to render a fixture that exceeds the defaults with sufficient explicit bounds. Populate enough rows to cross a real default (and show the default refusal) before asserting the explicit-bound render succeeds.

tests/builderops/ckm/test_overview_html.py
```diff
@@ -486,6 +486,113 @@ def test_cli_writes_overview(overview_store: CkmStore, tmp_path: Path) -> None:
...
+    overview_store: CkmStore,
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    observed: dict[str, int] = {}
+    original = CkmStore.load_projection_batch
+
+    def traced(self: CkmStore, **kwargs: int):
+        observed.update(kwargs)
+        if not kwargs:
```
RasmusTho: [P1] Exercise a real over-default snapshot in this Verify target\n\nThis exact-head repair still uses `overview_store`, whose object classes and aggregate are far below 500/3,000. The `if not kwargs` branch manufactures a default refusal, while the explicit-bound invocation succeeds against the same tiny store; this proves forwarding but not the issue AC that this named test render a fixture that actually exceeds the defaults. Add enough real rows to cross at least one default, assert the no-override call returns `snapshot_too_large` with no output, then assert sufficient explicit finite bounds render that same fixture.

tests/builderops/ckm/test_overview_html.py
```diff
@@ -486,6 +486,113 @@ def test_cli_writes_overview(overview_store: CkmStore, tmp_path: Path) -> None:
...
+        builderops,
+        [
+            "--db-path",
+            str(overview_store.db_path),
+            "ckm",
+            "overview",
+            "--out",
+            str(output),
+            "--class-capture-limit",
+            "0",
```
RasmusTho: [P1] Cover the negative half of the CLI-validation AC\n\nIssue #4098 explicitly requires this Verify target to prove that both zero and negative bounds are rejected before store materialization, but the test still invokes only `0`. Parameterize this case over `0` and a negative value (and preferably both option names) so a future range/parsing regression cannot satisfy the named AC with half its required input domain untested.

app/builderops/cli.py
```diff
@@ -694,13 +694,35 @@ def ckm_project(ctx: click.Context, projection_type: str, output_dir: Path) -> N
     type=click.Path(dir_okay=False, path_type=Path),
     required=True,
 )
+@click.option(
+    "--class-capture-limit",
```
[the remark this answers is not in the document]
RasmusTho: Owner doc now documents both explicit positive finite flags and preserves the default typed refusal semantics.

tests/builderops/ckm/test_overview_html.py
RasmusTho: Repaired in commit 6421f4946: the named test now runs both omitted defaults and explicit insufficient positive bounds, asserting snapshot_too_large and no partial output.

tests/builderops/ckm/test_overview_html.py
RasmusTho: Repaired in commit 6421f4946: the named test creates a real 501-artifact fixture, proves default refusal, then renders the same oversized snapshot with finite 501/3001 bounds.

tests/builderops/ckm/test_overview_html.py
RasmusTho: Repaired in commit 6421f4946: the named test creates a real 501-artifact fixture, proves default refusal, then renders the same oversized snapshot with finite 501/3001 bounds.

tests/builderops/ckm/test_overview_html.py
RasmusTho: Repaired in commit 6421f4946: the named test covers both zero and negative capture-bound values before store materialization.

PR #4158, opened 2026-07-27
fix: include selector contract tests for shared outbox producers

RasmusTho:
Final-Review-Rounds: 1

Governing-Issue: #4156

Fixes #4156

## Change Lane
- [ ] Implementation lane
- [ ] Docs authoring lane
- [x] Governance lane

## Summary
- Add the exact ObjectStore outbox-emission regression target to the `app/objects/__init__.py` selector path.
- Add the event-envelope contract suite to the `app/outbox/events.py` shared producer selector path and document the owner-doc contract.

## Validation
- `python3 scripts/validate_issue_readiness.py --body-file /var/folders/df/cmsr6vf5175313kjjbgqyhb00000gn/T/tmp.CrqrGLvhxR --label agent:ready` — ready_candidate
- `pytest -q tests/scripts/test_select_pr_tests.py::test_object_store_module_selects_store_ingest_regressions tests/scripts/test_select_pr_tests.py::test_outbox_embedding_events_select_shared_regressions` — 2 passed
- `python3 scripts/docs_guard.py` — Docs guard: OK
- `ruff check app tests` — All checks passed
- `git diff --check` — passed
- `python3 scripts/review_before_ci_gate.py --lane governance --review-gate-complete --risk-assessment-complete --changed-file docs/development/TEST_STRATEGY_HOT_PATH.md --changed-file scripts/select_pr_tests.py --changed-file tests/scripts/test_select_pr_tests.py` — satisfied

## BuilderOps Routing
- Records/projections/receipts: `lrn_20260727063457_f70207c2`
- Reason: Daily review-comment closure audit found that issue #4140 reached terminal state via PR #4141 while two actionable selector review comments remained unreplied and not fully satisfied.

## Review Thread Follow-Up
- Replied to #4141 `discussion_r3654213920` with #4156/#4158 repair evidence: `discussion_r3654898290`; GraphQL thread `PRRT_kwDOQEip6s6T7RKN` remains unresolved until this PR lands.
- Replied to #4141 `discussion_r3654213921` with #4156/#4158 repair evidence: `discussion_r3654898286`; GraphQL thread `PRRT_kwDOQEip6s6T7RKO` remains unresolved until this PR lands.

### DIFF
--- a/docs/development/TEST_STRATEGY_HOT_PATH.md
+++ b/docs/development/TEST_STRATEGY_HOT_PATH.md
@@ -31,8 +31,9 @@ The goal is to keep docs-only and governance/skill PRs cheap while preserving di
   the selected CI result as database-path evidence.
 - The exact shared producer `app/objects/__init__.py` uses the same `store_ingest` selection: its
   canonical object-store facade writes through the store provider seam and emits ingest lifecycle
-  events. The exact `app/outbox/events.py` producer belongs to both `outbox_worker` and
-  `memory_retrieval`, so its selection unions delivery-worker/event and indexer coverage. Other
+  events, and its selection includes the exact ObjectStore outbox-emission regression. The exact
+  `app/outbox/events.py` producer belongs to both `outbox_worker` and `memory_retrieval`, so its
+  selection unions delivery-worker/event, indexer, and event-envelope contract coverage. Other
   `app/objects/**` and `app/outbox/**` paths remain unowned unless explicitly mapped.
 - E2E tests under `tests/e2e/` run after merge and in the nightly suite, not on ordinary PRs. Opt-in classes (live LLM, browser, human UAT, eval) remain in their dedicated post-merge or nightly lanes.
 
--- a/scripts/select_pr_tests.py
+++ b/scripts/select_pr_tests.py
@@ -281,7 +281,12 @@ SUBSYSTEMS: tuple[tuple[str, tuple[str, ...], tuple[str, ...]], ...] = (
             "docs/DB_SCHEMA.md",
             "docs/RUNTIME_CORRECTNESS_KERNEL/",
         ),
-        ("tests/stores", "tests/ingest", "tests/architecture"),
+        (
+            "tests/stores",
+            "tests/ingest",
+            "tests/architecture",
+            "tests/services/test_outbox_idempotency.py::test_save_object_content_change_emits_new_event",
+        ),
     ),
     (
         "relevance",
@@ -377,6 +382,7 @@ SUBSYSTEMS: tuple[tuple[str, tuple[str, ...], tuple[str, ...]], ...] = (
             "tests/services/test_indexer_worker.py",
             "tests/cli/test_index_rebuild_resilience.py",
             "tests/cli/test_index_doctor_mixed_identity.py",
+            "tests/architecture/test_events_outbox_contracts.py",
             *E2E_TARGETS["memory_retrieval"],
         ),
     ),
--- a/tests/scripts/test_select_pr_tests.py
+++ b/tests/scripts/test_select_pr_tests.py
@@ -462,6 +462,10 @@ def test_store_ingest_change_selects_its_owned_contract_tests() -> None:
     assert "tests/stores" in selection.targets
     assert "tests/ingest" in selection.targets
     assert "tests/architecture" in selection.targets
+    assert (
+        "tests/services/test_outbox_idempotency.py::test_save_object_content_change_emits_new_event"
+        in selection.targets
+    )
 
 
 def test_object_store_module_selects_store_ingest_regressions() -> None:
@@ -474,6 +478,10 @@ def test_object_store_module_selects_store_ingest_regressions() -> None:
     assert "tests/stores" in selection.targets
     assert "tests/ingest" in selection.targets
     assert "tests/architecture" in selection.targets
+    assert (
+        "tests/services/test_outbox_idempotency.py::test_save_object_content_change_emits_new_event"
+        in selection.targets
+    )
 
 
 def test_outbox_worker_change_selects_worker_regressions() -> None:
@@ -504,6 +512,7 @@ def test_outbox_embedding_events_select_shared_regressions() -> None:
     assert "tests/indexer" in selection.targets
     assert "tests/index" in selection.targets
     assert "tests/services/test_indexer_worker.py" in selection.targets
+    assert "tests/architecture/test_events_outbox_contracts.py" in selection.targets
 
 
 def test_worker_metrics_module_change_has_a_ci_owner() -> None:


scripts/select_pr_tests.py
```diff
@@ -281,7 +281,12 @@
             "docs/DB_SCHEMA.md",
             "docs/RUNTIME_CORRECTNESS_KERNEL/",
         ),
-        ("tests/stores", "tests/ingest", "tests/architecture"),
+        (
+            "tests/stores",
+            "tests/ingest",
+            "tests/architecture",
+            "tests/services/test_outbox_idempotency.py::test_save_object_content_change_emits_new_event",
```
RasmusTho, line 288: **P2 — the first `::nodeid` target in `SUBSYSTEMS` silently bypasses `_existing_test_targets()`**

Independent review (I reached the same defect as `discussion_r3654912840` from the diff, and I concur that **P2 is the right severity — it is fail-loud, not false-green**; see the verification below for why it must not be escalated to P1, and why it also must not be waved through).

This is the only `::` target in the whole file (`grep -n "::" scripts/select_pr_tests.py` → line 288 only), so it establishes a new target class that the script's existing safety net does not cover:

```python
# scripts/select_pr_tests.py:838-842
def _existing_test_targets(targets: tuple[str, ...]) -> tuple[str, ...]:
    # Deleted/renamed test files must never reach pytest as a positional
    # target (pytest errors hard on a path that no longer exists). Directory
    # targets (never end in .py) always pass through unfiltered.
    return tuple(target for target in targets if not target.endswith(".py") or Path(target).is_file())
```

`"...test_outbox_idempotency.py::test_save_object_content_change_emits_new_event"` does not end in `.py`, so it takes the `not target.endswith(".py")` branch and is passed through **unfiltered** — and the comment's stated invariant ("Directory targets (never end in .py)") is now factually false, which will mislead the next maintainer of this function.

**Verified failure mode** (run at PR head in a clean worktree):

```
$ pytest -q -m "not pg" tests/ci tests/services/test_outbox_idempotency.py::test_RENAMED_does_not_exist
ERROR: not found: .../tests/services/test_outbox_idempotency.py::test_RENAMED_does_not_exist
(no match in any of [<Module test_outbox_idempotency.py>])
no tests ran in 0.44s
```

Consequences once the pinned test is renamed/deleted/moved:
- pytest exits `4` **before running anything**, so `tests/ci`, `tests/stores`, `tests/ingest`, `tests/architecture` are all skipped for that run.
- The CI step is `set -o pipefail` + `pytest ... | tee` (`.github/workflows/ci-smoke.yaml:668-669`), so the exit code does propagate → **red CI, no false green**. That is why this stays P2 and is not a protected-floor P0/P1.
- But the breakage lands on an **innocent third-party PR**: the PR that renames the test only matches `outbox_worker` (which owns `tests/services/test_outbox_idempotency.py` as a whole file, and *is* protected by the existence filter), so it goes green; the next PR touching `app/stores/**`, `app/ingest/**`, `app/objects/__init__.py`, `tests/stores/**`, `tests/ingest/**`, `docs/DB_SCHEMA.md` or `docs/RUNTIME_CORRECTNESS_KERNEL/**` eats an unrelated hard error with zero coverage executed.

**There is no other guard.** I checked: `tests/governance/test_select_pr_tests.py` has no static-target existence/collectability fitness test, and the two assertions added here (`tests/scripts/test_select_pr_tests.py:465-468`, `481-484`) only assert the literal *string* is in `selection.targets` — they would stay green against a dangling nodeid.

Suggested repair (both parts):

1. Make the existence filter nodeid-aware, e.g. check the path part:
   ```python
   file_part = target.split("::", 1)[0]
   return ... if not file_part.endswith(".py") or Path(file_part).is_file()
   ```
   and update the comment so it stops asserting the now-false "never end in .py" invariant.
2. Add a governance fitness test that every static `SUBSYSTEMS` / `*_TARGETS` entry resolves at HEAD — for a `::` entry, that the nodeid actually **collects** (`pytest --collect-only <nodeid>`). Step 1 alone still leaves the "file exists, test function renamed" window open, which is the more likely refactor.

scripts/select_pr_tests.py
```diff
@@ -281,7 +281,12 @@
             "docs/DB_SCHEMA.md",
             "docs/RUNTIME_CORRECTNESS_KERNEL/",
         ),
-        ("tests/stores", "tests/ingest", "tests/architecture"),
+        (
+            "tests/stores",
+            "tests/ingest",
+            "tests/architecture",
```
RasmusTho, line 287: **P3 — new exact target has no inline rationale, breaking this file's established convention**

Every other narrow/exact entry in `SUBSYSTEMS` carries an inline comment explaining *why* it is owned at that granularity and why it does not widen the subsystem — e.g. `tests/architecture/test_builderops_store_boundary.py` (L110-112), `tests/architecture/test_no_hardcoded_vault_layout.py` (L152-154), `app/objects/__init__.py` itself (L273-277), `app/outbox/events.py` (L509-511). The new target lands with none.

That matters more than usual here because this is the **first per-test `::nodeid` granularity in the file**. Without a comment, the next maintainer has no way to know whether the nodeid was deliberate (running the whole `tests/services/test_outbox_idempotency.py` would pull `outbox_worker`'s coverage into every store change) or an accident, and no warning that renaming the pinned test breaks `store_ingest` CI.

Suggest a short comment in the same style, e.g.:

```python
# The ObjectStore facade's outbox-emission regression (observation= content
# fingerprint, #2863) lives in outbox_worker's suite. Pin the single test
# rather than the whole file so a store change does not pull the delivery
# worker's coverage. NOTE: renaming this test breaks store_ingest selection.
```

docs/development/TEST_STRATEGY_HOT_PATH.md
```diff
@@ -31,8 +31,9 @@ The goal is to keep docs-only and governance/skill PRs cheap while preserving di
   the selected CI result as database-path evidence.
 - The exact shared producer `app/objects/__init__.py` uses the same `store_ingest` selection: its
   canonical object-store facade writes through the store provider seam and emits ingest lifecycle
-  events. The exact `app/outbox/events.py` producer belongs to both `outbox_worker` and
-  `memory_retrieval`, so its selection unions delivery-worker/event and indexer coverage. Other
+  events, and its selection includes the exact ObjectStore outbox-emission regression. The exact
+  `app/outbox/events.py` producer belongs to both `outbox_worker` and `memory_retrieval`, so its
+  selection unions delivery-worker/event, indexer, and event-envelope contract coverage. Other
```
RasmusTho, line 36: **P3 — owner-doc bullet does not name the exact targets it is supposed to protect**

Issue #4156 AC-3 is: *"The owner doc records the exact shared-producer selector contract **so future changes do not silently remove these targets**."* The prose here says "includes the exact ObjectStore outbox-emission regression" and "event-envelope contract coverage" without naming either target, so the doc itself cannot function as the anti-removal record.

This is also inconsistent with the two bullets immediately above in the same section, which are the file's established pattern and *do* name the exact paths:

> - Every `builder_system` match includes `tests/architecture/test_builderops_store_boundary.py` as an exact run target, not only as an ownership prefix. …
> - Every `vault` match includes `tests/architecture/test_no_hardcoded_vault_layout.py` as an exact run target, …

I am **not** rating this higher than P3 because the AC's protective intent is in fact met by the Verify-covered unit tests (`test_object_store_module_selects_store_ingest_regressions` / `test_outbox_embedding_events_select_shared_regressions` assert the literal target strings), and the AC-3 Verify target — "doc writeback at `docs/development/TEST_STRATEGY_HOT_PATH.md :: Current Protection Surface`" — is satisfied. Both Verify targets pass at head (`2 passed`), as does `python3 scripts/docs_guard.py`.

Suggest naming them, in the neighbours' style:
`tests/services/test_outbox_idempotency.py::test_save_object_content_change_emits_new_event` and `tests/architecture/test_events_outbox_contracts.py`.

scripts/select_pr_tests.py
```diff
@@ -377,6 +382,7 @@
             "tests/services/test_indexer_worker.py",
             "tests/cli/test_index_rebuild_resilience.py",
             "tests/cli/test_index_doctor_mixed_identity.py",
+            "tests/architecture/test_events_outbox_contracts.py",
```
RasmusTho, line 385: **P3 (follow-up, pre-existing) — the envelope contract suite is still unowned for its other two producers**

This addition is correct and strictly widening for `memory_retrieval`, and I verified it works end to end:

```
$ python3 scripts/select_pr_tests.py --changed-file app/outbox/events.py
subsystems=memory_retrieval,outbox_worker
pytest_args=... tests/architecture/test_events_outbox_contracts.py tests/workers tests/worker tests/services/test_outbox_idempotency.py ... tests/events
```

Noting for follow-up, not as a blocker on this PR: `tests/architecture/test_events_outbox_contracts.py` also imports and asserts on two other producers —

```python
from app.events.schema import make_outbox_event
from app.watcher.events import build_watcher_run_event, emit_watcher_run_event
```

`app/events/**` resolves to `events_receipts` (targets `tests/events`, `tests/receipts`, `tests/contracts`) and `app/watcher/**` to `watcher_sync`, and **neither selects this suite**. So a change to `make_outbox_event` — the shared envelope builder this file exists to pin — still does not run it. Same class of under-selection that #4156 was opened to repair, just for the other two producers.

Also verified as non-issues while checking union behaviour:
- `store_ingest`'s `tests/architecture` directory target + this exact-file target in the same run does **not** double-collect (`--collect-only` → 4 items either way).
- `outbox_worker`'s whole-file `tests/services/test_outbox_idempotency.py` + `store_ingest`'s new nodeid in the same run does **not** double-run (`19 passed`, same as the file alone).
- `tests/services/test_outbox_idempotency.py` carries `pytestmark = pytest.mark.not_pg`, so the pinned nodeid is not silently deselected by `PR_MARKER_EXPRESSION`.

scripts/select_pr_tests.py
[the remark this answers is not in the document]
RasmusTho: Confirmed and accepted as **P2 — deferred**, per `.codex/skills/verification-and-closure/SKILL.md :: Severity routing`.

The defect is real: `_existing_test_targets()` only filters targets ending in `.py`, so the new node-id target bypasses the deleted/renamed-test guard, and a future rename would strand unrelated `store_ingest` PRs on `ERROR: not found`. It does not meet a protected P0/P1 floor: pytest exits 4 and `set -o pipefail` in `.github/workflows/ci-smoke.yaml` propagates it to a red job, so it is fail-loud and can never produce a false-green check or receipt.

Durable defect intake: **#4206** — https://github.com/RasmusTho/agentic-pkm-mvp/issues/4206 (splits the file portion before the existence check, and adds a static target-collectability fitness test so a rename fails in the renaming PR rather than in an innocent downstream PR).

PR code left unchanged for this finding; merging PR #4158 at head `94f58dffe57bffeb6d863fcf9ad177061d3be298`.

RasmusTho: **P2 — deferred.** Durable defect intake: **#4206** — https://github.com/RasmusTho/agentic-pkm-mvp/issues/4206

#4206 carries both parts of the fix you named: split the file portion before the existence check, and add a static collectability fitness test over `SUBSYSTEMS` so the "file exists, function renamed" window is also closed and the failure lands in the renaming PR rather than in an innocent `store_ingest` PR.

Severity re-tested against the protected floors and it does not escalate: pytest exit 4 propagates through `set -o pipefail` to a red job, so no false-green CI/receipt/closure evidence is reachable. PR code left unchanged for this finding; merging at head `94f58dffe57bffeb6d863fcf9ad177061d3be298`.

PR #4210, opened 2026-07-27
Resolve the Signboard projection root to one source and give stale cards a repair path

RasmusTho:
Governing-Issue: #4198

Fixes #4198

Final-Review-Rounds: 1

## Summary
`SIGNBOARD_ROOT` was resolved three different ways — an active-vault-relative default in `app/dispatcher/signboard.py`, an independent `~/BuilderOpsVault/agent-delivery` literal in the API route and `start_full_system.sh`, and a shared-vault README claim that the vault's own `agent-delivery/` was the same board. This single-sources the root, adds the missing repair path for cards whose task id has left the dispatcher store, and turns the empty-board false-green into a visible error state.

## Changes
- `app/dispatcher/signboard.py` is the one source of the board-root default. `app/api/routes/signboard.py`, `scripts/start_full_system.sh`, and `scripts/export_runtime_env.sh` derive from `default_signboard_root()` via the new `scripts/lib/signboard_root.sh` helper and carry no board-root literal of their own. With no vault selected nothing invents a location: the variable stays unset and the BuilderOps bootstrap reports `signboard_root_missing`.
- `export-signboard --prune-absent` removes generated cards whose task id is absent from the dispatcher store — the `stale_card` drift `signboard-validate` reported and no command could clear (467 of 481 cards on the Demerzel board). Cards carrying human-authored `## Notes` content are kept and returned under `retained_with_notes` instead of deleted; malformed cards and non-generated files are never touched. The prune is opt-in, so the startup bootstrap and the `/signboard` refresh route still never delete. `signboard-validate` now names the repair command in its own output.
- `/api/signboard/board` reports `status: "error"` with an explicit message when the root is missing, is not a directory, or cannot be resolved at all, instead of returning six empty columns and HTTP 200.
- `scripts/export_runtime_env.sh` resolves and forwards the root itself rather than only passing through whatever the caller shell exported. That is why the channel deploy path started the API container with `SIGNBOARD_ROOT` empty while `start_full_system.sh` worked.
- Docs: `docs/AGENT_ISSUE_DISPATCHER.md` documents the single-sourced resolution, the prune, the error state, and a new subsection stating that the projection root and the shared BuilderOps vault's `agent-delivery/` are different trees with different schemas; `docs/builderops/BUILDEROPS_VAULT_STORE.md` no longer describes them as the same board.

Signboard stays projection-only per ADR-0010 — no claim, lease, or lock write path is added.

## BuilderOps Routing
- Reason: Repo-governed code/docs change only; the shared-vault README correction was applied in the BuilderOps vault itself and needs no repo record.

## Notes
Validation:
- `pytest -q tests/dispatcher/ tests/api/test_signboard_api.py tests/governance/test_project_pickup_deprecation.py tests/ops/test_builderops_startup_bootstrap.py` — 1122 passed
- Full `pytest -q -m "not pg"` under the host lease — 11061 passed, 9 failed; all 9 reproduce identically on clean `origin/main` in a separate worktree (missing `click` / `pydantic_settings` in the sandboxed subprocess env these tests build) and are unrelated to this change
- `ruff check app tests companion-ui/companion-app` — clean
- `python3 scripts/docs_guard.py` — OK; `python3 scripts/public_seam_lint.py --mode gate` — clean
---

### DIFF
--- a/app/api/routes/signboard.py
+++ b/app/api/routes/signboard.py
@@ -20,7 +20,13 @@ from app.dispatcher.config import load_paths
 from app.dispatcher.events import JsonlEventWriter
 from app.dispatcher.queue import block, complete
 from app.dispatcher.services import move_task
-from app.dispatcher.signboard import STATUS_COLUMNS, canonical_status, export_signboard
+from app.dispatcher.signboard import (
+    STATUS_COLUMNS,
+    NoActiveVaultError,
+    canonical_status,
+    default_signboard_root,
+    export_signboard,
+)
 from app.dispatcher.store import SqliteStore
 
 router = APIRouter(prefix="/signboard", tags=["signboard"])
@@ -62,9 +68,18 @@ def signboard_root() -> Path:
 
     No request controls this path. Resolving it once prevents a symlinked root
     from turning a card read into arbitrary traversal later on.
+
+    The default is not spelled here: it comes from ``default_signboard_root``
+    in the dispatcher projection module, which is the single source for the
+    Signboard root (#4198). ``SIGNBOARD_ROOT`` is the launcher-forwarded
+    override; when it is absent and no vault is selected this raises
+    :class:`NoActiveVaultError` so the board can render a visible error state
+    rather than an empty, healthy-looking one.
     """
-    raw = os.getenv("SIGNBOARD_ROOT", "~/BuilderOpsVault/agent-delivery")
-    return Path(raw).expanduser().resolve()
+    raw = os.getenv("SIGNBOARD_ROOT", "")
+    if raw.strip():
+        return Path(raw).expanduser().resolve()
+    return default_signboard_root()
 
 
 def _store() -> SqliteStore:
@@ -133,6 +148,12 @@ def read_signboard(root: Path) -> tuple[dict[str, list[dict[str, Any]]], list[st
     board: dict[str, list[dict[str, Any]]] = {column: [] for column in COLUMNS}
     errors: list[str] = []
     if not root.exists():
+        # An unreadable root used to render as six empty columns, which reads
+        # as "no work" instead of "misconfigured" (#4198). Report it.
+        errors.append(f"signboard root does not exist: {root}")
+        return board, errors
+    if not root.is_dir():
+        errors.append(f"signboard root is not a directory: {root}")
         return board, errors
     for column in COLUMNS:
         directory = root / column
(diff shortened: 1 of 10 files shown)


RasmusTho (commented):
## Independent review gate (local `code-review`)

**Verdict: pass — one P2 finding was fixed in `af414b4` and re-verified; no unresolved P0/P1.**

### What it does
Collapses three independent resolutions of the Signboard root into one (`default_signboard_root`), adds `export-signboard --prune-absent` for cards whose task id has left the dispatcher store, and turns a missing/unresolvable root from an empty 200 board into an explicit error state.

### Correctness
- `signboard_root()` no longer carries a default literal; `SIGNBOARD_ROOT` remains the launcher override and the fallback delegates to the dispatcher module. `test_signboard_root_is_single_sourced` pins that the route and the shell launchers stay literal-free, so the divergence cannot silently reappear.
- Dropping the `~/BuilderOpsVault` fallback means a no-vault host now has *no* board root rather than a wrong one. That is the right call for the no-vault idle posture: `builderops_startup` already had a `signboard_root_missing` degraded branch that was previously unreachable, and it now actually fires.
- `NoActiveVaultError` is handled at all three call sites. In `move_card` the inner `raise HTTPException` sits inside an outer `except ValueError`, and `NoActiveVaultError` is a `RuntimeError`, so neither the error nor the 503 is swallowed by the outer handler.
- The prune is correctly ordered after the write pass and keyed on the live task-id set, so a card for a task that merely moved column is untouched by it — the existing per-task glob still owns that case. Malformed generated cards are deliberately left for `signboard-validate` rather than deleted on an untrusted identity, which is the right conservative choice.
- Prune stays opt-in, so neither the startup bootstrap nor the `/signboard` refresh route can delete. That keeps the projection-only ADR-0010 posture intact — nothing here adds a claim, lease, or lock write path.

### Finding fixed during review

**P2 (fixed) — the prune protected `## Notes` but not `## Receipts`.**
`_render_task` emits *both* sections and always leaves `## Receipts` empty, and re-export never reaches a stale card — so text below that heading is human-authored and had survived precisely because nothing rewrote the card. The original prune would have deleted it. `_has_human_authored_content` now retains a stale card carrying either section, covered by `test_prune_retains_a_stale_card_carrying_human_receipts`, and the doc states both.

### Non-blocking observations

**P3 — `root` is `null` in the unresolved-root payload but a string otherwise.**
A consumer that formats `payload.root` unconditionally would render `"None"`. The bundled UI reads only `errors`, so nothing is affected today.

**P3 — the channel-deploy gap is documented, not closed.**
`scripts/deploy_channel.sh` reads the runtime env the last full-stack start produced rather than regenerating it, so a channel whose env predates the board-root line still comes up with `SIGNBOARD_ROOT` empty until a full-stack start regenerates it. Resolving it inside `export_runtime_env.sh` was tried and reverted: it put an `app.*` import ahead of that script's fail-loud settings-location import and masked it (caught by `tests/ops/test_export_runtime_env.py` in CI, not locally). Documenting the gap plus surfacing it as a visible error state is the honest resolution for this slice; closing it properly belongs to whatever owns runtime-env regeneration on the channel path.

### Tests
Good shape: the single-source assertion is a fitness test over the real files rather than a mock, the prune tests build genuine stale cards (task id absent from the store, so the exporter's own cleanup cannot reach them) rather than simulating absence, and both the retain-on-notes and retain-on-receipts paths are covered alongside the delete path. The API error-state tests cover both the missing-root and unresolvable-root branches, including the 503 on refresh.

### Security
No new request-controlled input: the board root is still resolved from environment/selection state only, and card reads keep their post-symlink containment check. The prune only unlinks files it has positively identified as generated cards inside the board root.

PR #4212, opened 2026-07-28
Stop materializing every evicted vault file to enforce SQLite confinement

RasmusTho:
Governing-Issue: #4199

Fixes #4199

Final-Review-Rounds: 1

## Summary
`builderops vault validate` opened every file in the shared vault to read a 16-byte SQLite header. On iCloud Drive most of those files are *dataless* — metadata local, bytes evicted — so each `open()` is a network round-trip and the documented operator check did not terminate on the real vault. The scan now stats first and materializes an evicted file only when its size could belong to a SQLite database image. The confinement invariant is unchanged.

## Changes
- `app/builderops/vault_queue.py`: `_sqlite_candidates` stats each entry before reading it. A file whose content is **local** is sniffed exactly as before — full header check, no narrowing, so an extension-less file that merely starts with the magic bytes is still caught. An **evicted** file (macOS `SF_DATALESS`, or zero allocated blocks against a non-zero size) is opened only when `size >= 512 and size % 512 == 0`. A SQLite image is always a whole number of pages and every legal page size is a multiple of 512, so that condition can never hide a real database — it is a property of the file format, not a heuristic about names or locations.
- Symlink rejection is untouched and still runs before any read. A `stat()` failure is skipped rather than raised, preserving the tolerance `Path.is_file()` provided for an entry that vanishes mid-walk.
- `builderops vault validate --progress` reports a stderr heartbeat plus a final `scanned N files, opened M, skipped K non-local` line; the JSON payload carries the same counts under `sqlite_scan`. That is the "slow vs hung" affordance the issue asked for.
- Doc writeback in `docs/builderops/BUILDEROPS_VAULT_STORE.md` (operator-check block: runtime cost, the new flag, `vault paths` stays instant) and `docs/BUILDEROPS_MODEL_INQUIRY/EXTERNAL_BUILDEROPS_VAULT_CONFIGURATION.md :: SQLite Confinement Invariant` (why the narrowing cannot weaken the invariant).

Nothing is written into the shared vault; no cache, no SQLite, no credentials.

## BuilderOps Routing
- Reason: Repo-governed code/docs change only; no operational record was produced.

## Notes
Validation:
- Real-vault before/after on the laptop against the shared BuilderOps vault (916 files). One file evicted with `brctl evict` to reproduce the dataless state (`st_flags 0x40000000`, `st_blocks 0`):
  - this branch: `0.13 s` scan, `{"files": 916, "opened": 915, "skipped_remote": 1}`, and the evicted file was **still dataless afterwards** — the scan did not download it.
  - `origin/main` on the same vault and the same single evicted file: `1.57 s` scan and the file came back **materialized**. One evicted file costs ~1.4 s, matching the ~1.0-1.2 s/file the issue measured; at ~900 evicted files that is the non-terminating cold run.
- `pytest -q tests/builderops/test_builderops_paths.py tests/builderops/test_builderops_claims.py` — 46 passed, including the unchanged extension-less and symlink-rejection confinement tests
- New `test_vault_validate_detects_extensionless_sqlite_without_full_scan` simulates a fully evicted vault: a real extension-less SQLite database is still rejected while 50 Markdown notes are never opened. Verified against `origin/main` in a separate worktree, where all 50 notes are opened.
- Full `pytest -q -m "not pg"` under the host lease — 11091 passed, 9 failed; all 9 reproduce identically on clean `origin/main` (missing `click` / `pydantic_settings` in the sandboxed subprocess env those tests construct on this host) and are unrelated to this change
- `ruff check app tests` — clean; `python3 scripts/docs_guard.py` — OK
---

### DIFF
--- a/app/builderops/cli.py
+++ b/app/builderops/cli.py
@@ -380,13 +380,43 @@ def vault_init(root: Path, as_json: bool) -> None:
 @vault.command("validate", help="Validate shared-vault separation and ticket status integrity.")
 @click.argument("root", type=click.Path(file_okay=False, path_type=Path))
 @click.option("--json", "as_json", is_flag=True)
+@click.option(
+    "--progress",
+    is_flag=True,
+    help=(
+        "Report scan progress on stderr while walking the vault, so a slow "
+        "synchronized vault is distinguishable from a hung command."
+    ),
+)
 @click.pass_context
-def vault_validate(ctx: click.Context, root: Path, as_json: bool) -> None:
+def vault_validate(ctx: click.Context, root: Path, as_json: bool, progress: bool) -> None:
     paths = _effective_paths(ctx)
-    payload = validate_vault(root, paths)
+    on_progress = _sqlite_scan_progress_reporter() if progress else None
+    payload = validate_vault(root, paths, on_progress=on_progress)
+    if progress:
+        scan = payload["sqlite_scan"]
+        click.echo(
+            f"scanned {scan['files']} files, opened {scan['opened']}, "
+            f"skipped {scan['skipped_remote']} non-local, {scan['elapsed_ms']} ms",
+            err=True,
+        )
     _emit(payload, as_json)
     if not payload["ok"]:
         raise click.ClickException("Builder Ops Vault validation failed")
+
+
+def _sqlite_scan_progress_reporter(every: int = 200) -> Callable[[dict[str, int]], None]:
+    """Emit a stderr heartbeat every ``every`` files so the scan looks alive."""
+
+    def report(stats: dict[str, int]) -> None:
+        if stats["files"] % every == 0:
+            click.echo(
+                f"scanning… {stats['files']} files, {stats['opened']} opened, "
+                f"{stats['skipped_remote']} skipped as non-local",
+                err=True,
+            )
+
+    return report
 
 
 @vault.command("claim", help="Write a shared advisory TTL claim for a Ready ticket.")
--- a/app/builderops/vault_queue.py
+++ b/app/builderops/vault_queue.py
@@ -5,7 +5,10 @@ from __future__ import annotations
 import json
 import os
 import re
+import stat
+import time
 import uuid
+from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


RasmusTho (commented):
## Independent review gate (local `code-review`)

**Verdict: pass — no P0/P1 findings.**

### What it does
Replaces the unconditional open-every-file header sniff in `_sqlite_candidates` with a stat-first scan. Local files are sniffed exactly as before; an evicted (iCloud dataless) file is materialized only when its size could be a SQLite database image. Adds `--progress` plus a `sqlite_scan` block to the payload.

### Correctness
- The narrowing predicate is sound for its purpose. A SQLite database image is `page_size * page_count` and every legal page size (512..65536, powers of two) is a multiple of 512, so `size >= 512 and size % 512 == 0` cannot exclude a well-formed database. The only thing it can skip is an *evicted* file that starts with the magic bytes without being a database image — and that case is still caught whenever the content is local, which is every file in a non-synchronized vault.
- Symlink rejection still runs before any `stat`/`open`, so the BMI-01 static-entry guarantee is intact.
- Swapping `Path.is_file()` for an explicit `stat()` + `S_ISREG` correctly preserves the old tolerance for an entry that disappears mid-walk. `is_file()` follows symlinks, but symlinks raise earlier, so the two are equivalent on the surviving entries.
- `_sqlite_candidates` keeps its old signature as a thin wrapper, so existing callers and confinement tests are untouched.

### Non-blocking observations

**P2 — `st_blocks == 0` can misclassify a local file as remote.**
`_content_is_local` treats zero allocated blocks against a non-zero size as "content not local". That is correct for iCloud dataless files, but APFS transparent compression (decmpfs, e.g. `afsctool`) also reports `st_blocks == 0` while the data is entirely local, and some network mounts report 0 for every file. On such a filesystem the scan would fall back to the size predicate for every file. The invariant still holds for any real database image (page-multiple size), so this cannot let an actual SQLite store through — but it does mean the magic-bytes-without-being-a-database case stops being caught there. The `sqlite_scan.skipped_remote` count makes the condition observable, which is the right mitigation for now; worth a follow-up only if a vault ever lands on such a mount.

**P3 — `elapsed_ms` is always `0` inside the progress callback.**
It is only computed after the walk, so every mid-scan callback carries a meaningless zero. Harmless, and the test pins the shape deliberately, but the field reads as if it were live.

**P3 — `sqlite_scan` stays all-zero when the scan raises.**
A symlink rejection leaves the initial zeroed dict in the payload alongside the error. Truthful enough (nothing was scanned) but could be read as "scanned a clean empty vault"; the `errors` list disambiguates.

**P3 — wrapped inline-code anchor in `BUILDEROPS_VAULT_STORE.md`.**
The `` `... :: SQLite Confinement\nInvariant` `` reference is split across a line break. Renders fine; just a little awkward to grep for.

### Tests
Coverage is proportionate and adversarial in the right place: the new test simulates a fully evicted vault and asserts both that a real extension-less database is still rejected *and* that the 50 decoy notes are never opened — the second half is what actually proves the fix, and it was verified to fail against `origin/main`. The unchanged extension-less and symlink confinement tests still pass untouched.

### Security
No new external input, no new write path, nothing written into the shared vault. The scan still reads at most 16 bytes per file.

PR #4393, opened 2026-07-29
MVR-01C: guard scalar rollback lineage and recovery

RasmusTho:
## Change Lane
- [ ] Docs authoring lane
- [ ] Governance lane

Final-Review-Rounds: 2

Governing-Issue: #3855

## Linked Issue
Closes #3855

## SBS Impact
- Primary subsystem: WSP
- Secondary subsystem(s): SFC,PDM,EBF
- Write class: mechanical durable instance-local state
- Persistence impact: versioned registry, authenticated scalar session, rollback projection, and host-global deployment authority
- Derived/rebuildable impact: database/cache projections remain rebuildable and non-authoritative
- New or changed contract: MVR-01C explicit scalar rollback and roll-forward lineage
- Owner-doc impact: updated deployment and release-channel owner docs
- Transition debt impact: retains only the specification-required scalar compatibility path
- Boundary risk: local concurrency, credential durability, deployment crash recovery, and selected-root isolation

## Owner-Doc Writeback

## Summary
- Deliver MVR-01C scalar rollback lineage with an explicit selected binding, authenticated gateway, minimum-runtime floor, and roll-forward merge.
- Serialize local scalar admission with deployment authority and make every crash/retry boundary fail closed on a single host.
- Route the shipped rollback command through the guarded overlay while preserving a capable current-image guard for retries.

## BuilderOps Routing
- Records/projections/receipts: LearningSignals lrn_20260729110215_fb19f361 and lrn_20260729192757_be6897ce
- Reason: Records mechanism-convergence divergences found while closing the scalar admission and retry state machine.

## Notes
Full sequential not-pg validation on the pre-base-drift reviewed head 20a4c1a5fe0daa066647863e888d2a23cff80dc9 produced 11714 passed, 40 skipped, 276 deselected, and 18 expected xfails; Ruff app/tests, mypy app (818 files), 5 import contracts, bash syntax, source anchors, docs guard with the governed runtime-PR temporal skip, and diff-check also passed. Final head b4b238ada63754dcd4e723e41d7d1c9c5ebf1960 preserves stable delivery patch ID 242c9335604dec9d226f5063035ea7850b67e9cc across a conflict-free, zero-path-overlap rebase. On that exact final head, all 8 mapped Issue selectors and 37 deploy/Linux inventory gates pass, Ruff and diff-check pass, and the fresh Sol/xhigh review is CLEAN with no P0/P1/P2 findings. The test-only correction isolates shared-runner process churn without changing the production fail-closed inventory helper. Current-SHA GitHub CI passed without rerun.


scripts/deploy_channel.sh
```diff
@@ -348,6 +395,13 @@ embedding_provider_preflight_gate() {
 
 recreate_channel_services() {
   local rc=0
+  if [ "${scalar_rollback}" = "1" ]; then
+    compose stop api worker watcher heimdal-capture-watch companion-ui \
+      scalar-rollback-gateway || return $?
+    compose up -d --force-recreate \
+      scalar-rollback-guard api scalar-rollback-gateway
```
RasmusTho, line 470: [P1] Retire scalar services before normal roll-forward

This branch starts `scalar-rollback-gateway`, but the normal branch below neither stops nor removes that service, and the non-scalar Compose model no longer names it. On scalar-to-current roll-forward the gateway therefore keeps the channel API port while the regular API tries to publish the same port, so recreation fails. Stop/remove the orphaned scalar services before current-mode startup and cover the scalar-to-current transition. This is a protected roll-forward state-machine and acceptance failure.

scripts/deploy_channel.sh
```diff
@@ -367,6 +421,37 @@ recreate_channel_services() {
...
+      docker inspect -f \
+        '{{if .State.Health}}{{.State.Health.Status}}{{else if .State.Running}}running{{else}}stopped{{end}}' \
+        "${cid}" 2>/dev/null
+    )" || rc=$?
+    if [ "${rc}" -ne 0 ]; then
+      echo "scalar rollback gate: ${service} state is unreadable" >&2
+      return "${rc}"
+    fi
+    case "${status}" in
+      healthy|running) ;;
```
RasmusTho, line 554: [P1] Require functional authenticated gateway proof

For `scalar-rollback-gateway`, which has no healthcheck, this accepts a merely running nginx process. The launcher only proves that the htpasswd path is a file, so an empty or unusable credential store can let nginx run and this gate record success while no authenticated request can succeed. Validate the credential artifact and prove an authenticated request (or an equivalent functional healthcheck) before writing success. Authentication and a failed governing acceptance target have a protected P1 floor.

docker-compose.scalar-rollback.yml
```diff
@@ -0,0 +1,140 @@
...
+          create_host_path: false
+      - type: bind
+        source: ${SCALAR_ROLLBACK_VAULT_ROOT:?validated selected root is required}
+        target: /app/selected-vault
+        bind:
+          create_host_path: false
+    environment:
+      INSTANCE_STATE_LEGACY_ROLLBACK: "1"
+      DESIGN_HANDOFF_APP_LOCAL_SETTINGS: /app/scalar-rollback/app-local.md
+      VAULT_ROOT: /app/selected-vault
```
RasmusTho, line 101: [P1] Clear environment-scoped vault selectors in scalar mode

Compose merges environment maps, so setting only `VAULT_ROOT` here does not remove earlier channel/vault selectors. The actual dev + full-host + scalar model retains `VAULT_ROOT_DEV` and `VAULT_ROOT_TEST` from the prior overlay while `!override` mounts only `/app/selected-vault`; `resolve_vault_root(environment="dev"|"test")` prefers those scoped values. The old runtime can therefore select a non-target path if it is accessible, or fail startup instead of using the validated binding. Explicitly reset/override the scoped selectors and validate the complete effective Compose model, not only the base and scalar files. This violates the selected-binding acceptance target and is P1.

RasmusTho (commented):
Final local /code-review gate for exact SHA 5ffe2f0f03826b7886138602e7de36547485d1ad against origin/main 1027150b614920e1e4a3946db486f310974e8d0c: BLOCKING. P0=0, P1=3, P2=0, P3=0. Current-SHA GitHub CI is green; all 8 governing acceptance selectors plus the protected import-boundary test pass locally (9 passed), and the latest three-file import correction is sound. The remaining P1s are protected contract/state-machine failures in scalar-to-current service cleanup, authenticated gateway proof, and effective selected-root isolation. Residual risk: scalar rollback can be recorded without usable authentication, dev/test can retain a non-selected vault selector, and roll-forward can fail because the old gateway still owns the API port. Multiple adjacent blockers remain in the same scalar deployment mechanism, so the mechanism-convergence gate must be clean again before another expensive proof cycle.

scripts/deploy_channel.sh
RasmusTho: Fixed in ebdd659c53af49e3e77558563407710acacdb060. Normal roll-forward now finds and force-removes only containers carrying the exact Compose project/service labels for the scalar gateway and guard before normal recreation. test_normal_roll_forward_retires_scalar_services_before_recreate proves lookup/removal ordering.

scripts/deploy_channel.sh
RasmusTho: Fixed in ebdd659c53af49e3e77558563407710acacdb060. Scalar preflight now validates a non-empty htpasswd plus a mode-0600 netrc whose loopback login exists in that htpasswd, then proves both invalid-password HTTP 401 and authenticated /healthz JSON ok=true with proxy bypass before writing the receipt. Negative tests preserve the prior receipt.

docker-compose.scalar-rollback.yml
RasmusTho: Fixed in ebdd659c53af49e3e77558563407710acacdb060. The scalar overlay now pins VAULT_ROOT, VAULT_ROOT_DEV, VAULT_ROOT_TEST, and WATCHER_VAULT_PATH to /app/selected-vault; the guard validates all four. The rendered dev + full-host + scalar test asserts the effective model and selected-only mount.

scripts/deploy_channel.sh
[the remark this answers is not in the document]
RasmusTho: Resolved on current head 20a4c1a5fe0daa066647863e888d2a23cff80dc9. Normal roll-forward now retires the exact labeled scalar gateway and guard before regular API recreation; test_normal_roll_forward_retires_scalar_services_before_recreate proves ordering. Current-SHA CI is green.

scripts/deploy_channel.sh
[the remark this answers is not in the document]
RasmusTho: Resolved on current head 20a4c1a5fe0daa066647863e888d2a23cff80dc9. Scalar admission validates the credential artifacts, proves invalid credentials receive 401, and proves an authenticated /healthz response before success is recorded. Negative tests preserve the prior receipt; current-SHA CI is green.

PR #4420, opened 2026-07-30
Serialize shared health-contract evaluation

RasmusTho:
## Change Lane
- [ ] Docs authoring lane
- [ ] Governance lane

Final-Review-Rounds: 2

Governing-Issue: #4322

## Linked Issue
Closes #4322

## SBS Impact
- Primary subsystem: OEF
- Secondary subsystem(s): PDM
- Write class: mechanical current-state correction
- Persistence impact: Serializes durable health transition history and incident accounting; no schema change
- Derived/rebuildable impact: none
- New or changed contract: Shared health-contract mutation and full-snapshot completion are ordered by evaluation admission sequence
- Owner-doc impact: none, as classified by the governing Issue
- Transition debt impact: Reduces concurrent health-state drift
- Boundary risk: High concurrency surface; overlapping readiness evaluations must not corrupt hysteresis, transition history, incident accounting, or dependency readiness

## Owner-Doc Writeback

## Summary
- Protect each HealthStateMachine transition with a short per-instance mutation lock while keeping diagnostics concurrent.
- Assign shared-contract evaluations monotonic admission sequences so older, slower diagnostics cannot overwrite newer state or move transition time backward.
- Publish full health snapshots through a short per-contract completion fence so a late older result inherits the newest completed dependency-down snapshot.
- Keep /readyz and /status off the event loop without making unrelated AnyIO workers queue behind a whole-evaluation mutex.

## BuilderOps Routing
- Reason: The governing Issue, dispatcher claim, PR, and validation receipts fully represent this bounded delivery; no separate learning, freshness, roadmap, or promotion material was produced.

## Notes
Dispatcher claim: verified lease lease-d03623fb for github-RasmusTho--agentic-pkm-mvp-issue-4322.

Low-convergence receipt: the first published whole-evaluation mutex passed CI but final review found shared-worker starvation and scheduler-dependent proof. Subsequent strongest-capability reviews exposed stale diagnostic completion and a split dependency/publication fence before additional expensive CI. The converged mechanism on SHA 74cca7c901a6950853f39ad7ba3d801cc1b90210 serializes only canonical mutation/publication, rejects stale state completions by admission sequence, invalidates pre-reset samples, and returns the newest fully completed snapshot when an older evaluation finishes late. The deterministic regression forces DB-down to pause after state invalidation, lets the older healthy evaluation observe the split, completes DB-down first, and proves the older result remains unhealthy/write-blocked. Fresh independent Sol/xhigh convergence review: PASS, no P0/P1 findings.

Base drift: rebased cleanly from a2ef71456cf72d2005e898508c8fa04666e58d6a to 6e9bc33787135883d294e62af430d9ed79c963bc; incoming commits touched no health-contract files.

Validation on SHA 74cca7c901a6950853f39ad7ba3d801cc1b90210: 54 affected health/readiness/incident tests passed; governing Verify targets passed; ruff check app tests companion-ui/companion-app passed; mypy app passed for 825 source files; git diff --check passed. Review-before-CI gate passed for risk:concurrency and risk:state-machine. Current-SHA CI and two consecutive final clean reviews remain merge gates.

### DIFF
--- a/app/health_contract.py
+++ b/app/health_contract.py
@@ -293,7 +293,7 @@ class HealthStateMachine:
         *,
         now: datetime | None = None,
     ) -> tuple[str, str, str]:
-        state, reason, since_ts, _, _ = self.update_with_transition(
+        state, reason, since_ts, _, _, _ = self.update_with_transition(
             age,
             thresholds,
             now=now,
@@ -307,14 +307,14 @@ class HealthStateMachine:
         *,
         now: datetime | None = None,
         sample_sequence: int | None = None,
-    ) -> tuple[str, str, str, bool, list[dict[str, str]]]:
+    ) -> tuple[str, str, str, bool, list[dict[str, str]], bool]:
         with self._lock:
             if sample_sequence is None:
                 self._next_sample_sequence += 1
                 sample_sequence = self._next_sample_sequence
             if sample_sequence <= self._last_applied_sample_sequence:
                 state, reason, since_ts, transition_history = self._snapshot_unlocked()
-                return state, reason, since_ts, False, transition_history
+                return state, reason, since_ts, False, transition_history, False
             previous_state = self.state
             state, reason, since_ts = self._update_unlocked(
                 age,
@@ -328,6 +328,7 @@ class HealthStateMachine:
                 since_ts,
                 state != previous_state,
                 list(self.transition_history),
+                True,
             )
 
     def reserve_sample_sequence(self) -> int:
@@ -436,6 +437,7 @@ class HealthContract:
         self._dependency_incident_lock = Lock()
         self._dependency_incident_reported = False
         self._dependency_incident_sequence = 0
+        self._dependency_down_reason: str | None = None
 
     def evaluate(self) -> dict[str, Any]:
         return self._evaluate()
@@ -483,7 +485,7 @@ class HealthContract:
             )
         self._should_capture_dependency_incident(
             sample_sequence=sample_sequence,
-            is_down=False,
+            down_reason=None,
             current_state=self.state_machine.snapshot()[0],
         )
 
@@ -497,7 +499,7 @@ class HealthContract:
         object_count, store_count_error = self._count_objects()
         latest_ts = self._latest_timestamp(records)
         age = self._compute_age(latest_ts, now)
-        state, reason, since_ts, transitioned, transition_history = (
+        state, reason, since_ts, transitioned, transition_history, sample_applied = (
             self.state_machine.update_with_transition(
                 age,
                 settings_result.settings.thresholds,
@@ -505,6 +507,12 @@ class HealthContract:
                 sample_sequence=sample_sequence,
             )
         )
+        if not sample_applied:
+            dependency_override = self._dependency_override_after(sample_sequence)
+            if dependency_override is not None:
+                state = "unhealthy"
+                reason = dependency_override
+                since_ts = now.isoformat()
         catch_up_progress = self._catch_up_progress(
             outbox_count,
             age,
@@ -678,7 +686,7 @@ class HealthContract:
         )
         capture_dependency_incident = self._should_capture_dependency_incident(
             sample_sequence=sample_sequence,
-            is_down=True,
+            down_reason=reason,
             current_state=current_state,
         )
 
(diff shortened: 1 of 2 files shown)


app/health_contract.py
```diff
@@ -352,8 +353,13 @@ def __init__(
         # readiness short-circuit without a real Postgres. Bounded by the caller's
         # connect_timeout so the health probe never hangs.
         self.db_ping_fn = db_ping_fn or ping_postgres
+        self._evaluation_lock = Lock()
 
     def evaluate(self) -> dict[str, Any]:
+        with self._evaluation_lock:
+            return self._evaluate()
```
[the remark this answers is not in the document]
RasmusTho: Independent final review round 1 confirms this as a blocking P1 on `151a3d2c7f60cfb5e9a47268feb8f395e78b223d`. Both routes call `run_in_threadpool(DEFAULT_CONTRACT.evaluate)` before `evaluate()` waits on the per-instance lock, so every concurrent waiter consumes the shared AnyIO worker capacity. Because one health evaluation includes bounded DB ping plus additional store/filesystem/index work, a burst can monopolize the shared pool for the serialized queue duration even though the asyncio loop itself remains responsive. Serialization/coalescing must happen before shared-pool admission, or health evaluation must use isolated worker capacity.

tests/api/test_health_contract_api.py
```diff
@@ -100,6 +102,111 @@ async def assert_nonblocking(endpoint) -> None:
     asyncio.run(assert_nonblocking(health_contract_route.health_status))
 
 
+def test_shared_health_contract_evaluation_is_serialized(monkeypatch) -> None:
+    evaluation_barrier = threading.Barrier(2, timeout=0.2)
```
RasmusTho: [P1] Make the governing serialization proof scheduler-deterministic

This `Barrier(2, timeout=0.2)` can break before the second thread starts on a loaded runner. In that schedule, even an implementation with no serialization enters `_evaluate` only once before timeout, then the late second call observes the already-broken barrier; `maximum_active_evaluations` remains 1 and this Issue’s explicit `Verify:` target false-passes. Coordinate worker admission with events/futures so the test first proves both route calls have reached the concurrency seam, then release the first evaluation and assert the second could not enter until it completed; do not use a fixed scheduling window as the correctness oracle.

RasmusTho (commented):
Final independent review round 1 found a second blocking P1 in the governing concurrency proof. This review covers exact head `151a3d2c7f60cfb5e9a47268feb8f395e78b223d`; it does not represent final review round 2.

app/health_contract.py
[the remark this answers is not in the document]
RasmusTho: Fixed in 0225c5e5ddc002fdc80f0d6e9fa89702a4e861eb. The whole-evaluation mutex is gone. Diagnostics now run concurrently; only the in-memory HealthStateMachine mutation is locked, and admission sequences reject stale completions. The mixed-consumer regression proves unrelated AnyIO worker work completes while one health diagnostic is blocked and the remaining shared tokens evaluate the same contract. Fresh Sol/xhigh convergence review passed.

tests/api/test_health_contract_api.py
RasmusTho: Fixed in 0225c5e5ddc002fdc80f0d6e9fa89702a4e861eb. The governing test now uses explicit events and an instrumented transition lock: it proves the second route reached the lock while the first update is held, then asserts only one mutable update was active. A separate production-path overtake test deterministically proves an older blocked diagnostic cannot overwrite a newer transition or move since/history backward.

app/health_contract.py
```diff
@@ -282,6 +292,77 @@ def update(
...
+        thresholds: HealthThresholds,
+        *,
+        now: datetime | None = None,
+        sample_sequence: int | None = None,
+    ) -> tuple[str, str, str, bool, list[dict[str, str]]]:
+        with self._lock:
+            if sample_sequence is None:
+                self._next_sample_sequence += 1
+                sample_sequence = self._next_sample_sequence
+            if sample_sequence <= self._last_applied_sample_sequence:
```
RasmusTho, line 316: [P1] Keep a newer DB-down result authoritative over stale healthy completions

When request A reserves an older sequence and passes its DB probe, then request B reserves a newer sequence and observes DB-down, B calls `advance_sample_sequence()` and returns `unhealthy`. If A later reaches this stale branch, the state machine still contains its outbox state because DB-down intentionally does not mutate hysteresis, so A receives `running` here and returns `writes_allowed: true` after the newer 503. I reproduced this on `0225c5e5ddc002fdc80f0d6e9fa89702a4e861eb` with A blocked immediately after a healthy ping and B returning DB-down first: B was `unhealthy`/false, then A completed `running`/true. Preserve the latest dependency override in the canonical completion fence (or otherwise make stale full evaluations retry/fail closed) so an older healthy request cannot restore false-green readiness.

RasmusTho: Fixed on the rebased head 74cca7c901a6950853f39ad7ba3d801cc1b90210. Normal and DB-down evaluations now publish full snapshots through a short per-contract completion fence keyed by admission sequence; if an older evaluation finishes after a newer completed DB-down evaluation, it returns a deep copy of that newer unhealthy/write-blocked snapshot. The regression deterministically pauses DB-down between state invalidation and dependency publication, lets the older evaluation pass the former split, completes DB-down first, then proves the older result cannot return green. Fresh Sol/xhigh convergence review passed.

PR #4424, opened 2026-07-30
feat(heimdal): CDLM-06 live transcript and default analysis projections

RasmusTho:
Governing-Issue: #4386

Fixes #4386

Final-Review-Rounds: 2

## Summary
Ships CDLM-06: per-segment ASR derived exactly once per content hash through the shared engine seam (`app.media.transcribe.run_asr`, one-Whisper rule), triggered from the production admission path; a sequence-ordered transcript projection with explicit gap markers and no person attribution; a `generic-default@1` analysis projection emitted as `derived_projection` blocks with `{revision, derived_from, template_id, engine}` provenance, convergent over identical admitted sets and cached by input-set identity; the template precedence seam (user selection > permitted metadata > default); the side-effect-free `GET /api/heimdal/meeting/{session_id}/projection` read; late admissions deriving a new revision; per-segment fail-loud ASR needs-attention isolated from other segments; and durable, rebuildable projection state (migration `b8d3f0a5c2e4` for Postgres/PDM, file-backed SQLite dev/test lane). The analysis engine in this slice is deterministic (no LLM), so block-level convergence is structural.

## Claim receipt
`pickup-claim-complete issue=4386 branch=feat/4386-cdlm-06-meeting-projections worktree=/Users/rasmus/worktrees/cdlm-06-4386 coordination_mode=dispatcher-backed fallback_reason=none task_id=github-RasmusTho--agentic-pkm-mvp-issue-4386 lease_id=lease-7dffaa0d holder=cdlm-hub-coordinator evidence=verified-dispatcher-lease`

## Risk gates
- TCD risk surfaces: migration, data. `scripts/review_before_ci_gate.py` satisfied at push SHA.
- Mechanism convergence packet (`heimdal-meeting-projection`) built; pre-CI independent review found two P1s (derivation store failures raising into acknowledged admissions; a revision insert race silently dropping an input set) and one P2 (failed-retry deleting the failure record before ASR) — all fixed in the second commit; post-fix independent convergence re-review returned CLEAN (one benign P3 TOCTOU noted: a concurrent identical input set can mint a duplicate — deterministic, identical blocks, no data loss).

## BuilderOps Routing
- Reason: no plan divergence with a repairable upstream artifact this slice; claim state carried by the lease receipt above.

## Notes
Validation:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -p pytest_asyncio.plugin -q tests/heimdal/test_meeting_projection.py` — 6 passed (all AC `Verify:` targets; ASR stubbed at the real shared seam via the module attribute the code calls).
- `pytest -q -m "not pg and not alpha_llm" -n auto tests` — 11815 passed; the only failures are the 8 pre-existing local-environment failures (`tests/ops/test_instance_state_volume_contract.py` ×7 and `tests/builderops/test_model_inquiry_cli.py`) previously proven to reproduce identically on clean `origin/main` on this host. Post-suite hardening commits re-verified with the focused heimdal suites (51 passed), lint, and mypy.
- `ruff check app tests companion-ui/companion-app` — clean. `mypy app` — clean (827 files). `docs_guard.py` — OK (STATUS.md/DOCS_INDEX.md/task-doc writeback included).
---


app/heimdal/meeting_projection.py
```diff
@@ -0,0 +1,1142 @@
...
+    store = _backend()
+    hashes = [seg["content_sha256"] for seg in report["segments"]]
+    derivations = store.derivations_for(hashes)
+
+    by_seq = {seg["seq"]: seg for seg in report["segments"]}
+    max_bound = 0
+    if report["closed"] and report["final_seq_count"] is not None:
+        max_bound = report["final_seq_count"]
+    elif report["received"]:
+        max_bound = max(report["received"]) + 1
```
RasmusTho: P2 (final review r1): transcript silently drops admitted segments with `seq >= final_seq_count`. When the session is closed, `max_bound` is clamped to `final_seq_count`, but `meeting_ledger.record_segment_admission` accepts any late seq with no upper bound against the declared count. A client that closes with an undercounted `final_seq_count` (or resends a segment numbered at/above it) gets that segment ledgered and derived — it appears in `analysis.derived_from` and in the ledger gap report `segments` — yet it is elided from the `transcript` array with no gap/needs-attention marker, violating INV-CDLM-9 (never elided) and making transcript and analysis mutually inconsistent. Fix is a one-liner: `max_bound = max(final_seq_count, max(received)+1 if received else 0)` when closed.

RasmusTho: Fixed in 1ee1df9bd8782699bb46188f1e661755822927e3: the transcript bound is now max(final_seq_count, received max + 1), so an over-count admission renders instead of being elided. Regression proof: tests/heimdal/test_meeting_projection.py::test_transcript_shows_segments_beyond_declared_count.

app/heimdal/meeting_projection.py
```diff
@@ -0,0 +1,1146 @@
...
+    content_sha256: str,
+    media_bytes: bytes,
+    kind: str,
+) -> Optional[AsrDerivation]:
+    if kind != "audio":
+        return None
+    store = _backend()
+    existing = store.get_derivation(content_sha256)
+    if existing is not None and existing.status == DERIVATION_OK:
+        return existing
```
RasmusTho, line 887: **P1 — idempotent replay skips analysis re-derivation, leaving the analysis permanently excluding an admitted+derived segment.**

The early return on an existing `ok` derivation row exits before `_rederive_analysis(session_id)` runs. Scenario (reproduced locally): admission derives the segment, `insert_derivation` succeeds, then the analysis re-derivation crashes (store error / process crash between the derivation insert and the revision insert). The client's idempotent resend hits this short-circuit and never re-derives the analysis — `build_projection` is side-effect-free by design, so if this was the session's last segment (e.g. session already closed), the analysis revision permanently lacks a segment the transcript shows. The docstring claims a failed re-derivation is "re-attempted on the next segment or read", but reads derive nothing, and there may be no next segment. Fix: on the `DERIVATION_OK` short-circuit (and after the not-created/stored-row-wins branch), still call `_rederive_analysis` (it is cheap when the input-set identity already has a revision — cache hit returns immediately), or reconcile at close time.

app/heimdal/meeting_projection.py
```diff
@@ -0,0 +1,1146 @@
...
+                    json.dumps(derivation.engine),
+                    _iso(derivation.derived_at),
+                ),
+            )
+            return cur.rowcount == 1
+
+    def derivations_for(self, hashes: List[str]) -> Dict[str, AsrDerivation]:
+        if not hashes:
+            return {}
+        placeholders = ",".join("?" for _ in hashes)
```
RasmusTho: **P2 — `derivations_for` builds one `?` placeholder per content hash; at ledger scale this exceeds SQLite's variable limit and 500s the projection.**

`MAX_SESSION_SEQ` is 100,000, so a session can legally hold ~100,001 segment rows, but `SQLITE_MAX_VARIABLE_NUMBER` defaults to 32,766 (999 on older builds). Beyond that, `build_projection` and every `_admitted_input_set` call raise `sqlite3.OperationalError: too many SQL variables`, turning the projection route into a persistent 500 and silently killing analysis re-derivation for the session (the derivation trigger swallows it). Dev/test lane only (the pg lane uses `= ANY(%s)`), but the dev lane is exactly where the restart criterion is proven. Chunk the IN-list (e.g. batches of 500) or use a temp table / json_each.

app/api/routes/heimdal_meeting.py
```diff
@@ -163,6 +163,27 @@ def close_session(
     return _session_response(session, replay=not newly_closed, trace_id=trace_id)
 
 
+@router.get("/{session_id}/projection")
+def projection(request: Request, session_id: str) -> Dict[str, Any]:
+    """The iPad's poll target (CDLM-06): transcript + analysis projections.
```
RasmusTho, line 168: **P2 — the projection poll response is unbounded: up to ~100k transcript rows plus O(retention x segments) revision provenance per poll.**

`build_projection` materializes one row per sequence up to `max(final_seq_count, max(received)+1)` — a single admitted (or maliciously claimed) segment at seq 100,000, or a close with `final_seq_count=100000`, yields ~100k `gap` dicts in every poll response. On top of that, the `revisions` array repeats each retained revision's full `derived_from` list (up to 20 revisions x n segments of `{seq, sha256}` pairs), so a long meeting's poll payload grows roughly O(20n) — easily tens of MB, serialized on every iPad poll of a side-effect-free endpoint. Consider run-length-encoding gap ranges (the gap report already exposes `missing` as a list of seqs), dropping `derived_from` bodies from the revisions summary (keep counts + input_set hash), and/or pagination.

RasmusTho (commented):
Final review round 2 (fresh adversarial pass): 1 P1 and 2 P2 findings inline.

PR #4428, opened 2026-07-30
feat(heimdal): CDLM-07 meeting block-ownership model and fail-closed guard

RasmusTho:
Governing-Issue: #4387

Fixes #4387

Final-Review-Rounds: 2

## Summary
Ships CDLM-07: a durable per-session block registry (`{block_id, owner, type, provenance, created_at, revised_at}`), one shared fail-closed block-write guard (`app/heimdal/meeting_blocks.py::apply_block_write`) through which every mutation passes — user edits, analysis revisions, reconciliation, template re-render, finalization — with `user_note` blocks writable only by the user's structural editor identity, derived writers confined to blocks their own provenance minted (no move/merge/reorder/renumber outside it), and unknown/conflicting ownership refusing the write, preserving content byte-for-byte, and recording a surfaceable refusal. `POST /api/heimdal/meeting/{session_id}/user-note` is idempotent by `(note_block_id, revision)` with client-monotonic revisions and CDLM-01 ack ordering (durable write + committed `heimdal.meeting.user_note.written` event before 2xx). The CDLM-06 transcript/analysis writers register their blocks through the same guard; refusals and the page composition surface on the projection read. Migration `c9e4a1b6d3f5` adds DB-level triggers (identity immutability, no deletes, append-only history/refusals, and a user_note provenance-claim backstop stated honestly as claim-not-authority).

## Claim receipt
`pickup-claim-complete issue=4387 branch=feat/4387-cdlm-07-block-ownership-guard worktree=/Users/rasmus/worktrees/cdlm-07-4387 coordination_mode=dispatcher-backed fallback_reason=none task_id=github-RasmusTho--agentic-pkm-mvp-issue-4387 lease_id=lease-658e9783 holder=cdlm-hub-coordinator evidence=verified-dispatcher-lease`

## Risk gates
- TCD risk surfaces: security, data, migration (protected-finding class). `scripts/review_before_ci_gate.py` satisfied at push SHA.
- Mechanism convergence packet (`heimdal-meeting-block-guard`) built; pre-CI independent review found two P1s (a route replay shortcut bypassing guard authorization; provenance_extra overriding structural provenance keys) and three P2s (unauthorized create-replays, false ack on reused revision with different text, no client-monotonic enforcement) — all fixed; post-fix convergence re-review returned CLEAN with one P2 (authorization probe could mutate under a revise race), closed by a `verify_only` guard mode that can never write, plus auth-before-disclosure ordering.

## BuilderOps Routing
- Reason: no plan divergence with a repairable upstream artifact this slice; claim state carried by the lease receipt above.

## Notes
Validation:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -p pytest_asyncio.plugin -q tests/heimdal/test_meeting_block_ownership.py` — 7 passed (all six AC `Verify:` targets plus the convergence-regression test), driving the production surfaces (real admission path mints transcript blocks; real analysis derivation mints analysis blocks; the guard function the production writers call is the seam under test).
- `pytest -q -m "not pg and not alpha_llm" -n auto tests` — 11829 passed; failures are the known pre-existing host-environment set (`tests/ops/test_instance_state_volume_contract.py`, `tests/builderops/test_model_inquiry_cli.py` — proven on clean `origin/main`) plus one parallel-only flake (`tests/ops/test_start_full_runtime_health.py`, passes serially).
- `ruff check app tests companion-ui/companion-app` — clean. `mypy app` — clean (829 files). `docs_guard.py` — OK (STATUS.md/DOCS_INDEX.md/task-doc/EVENTS.md writeback included).
---


app/heimdal/meeting_blocks.py
```diff
@@ -0,0 +1,1136 @@
...
+            session_id=session_id,
+            block_id=block_id,
+            writer=writer,
+            action=action,
+            reason=f"unknown writer kind {writer.kind!r}",
+        )
+
+    existing = store.get_block(block_id)
+
+    if action == ACTION_CREATE:
```
RasmusTho, line 858: P2 (final review r1): `verify_only=True` is honored only on the revise/retire/move branch (line 998); the `ACTION_CREATE` branch never checks it, so a caller probing authorization for a create (`verify_only=True, action=ACTION_CREATE`, unknown block_id) actually inserts the block. The docstring contract ("an authorization probe never writes") is violated for the create path. The current route never does this, but `apply_block_write` is the shared seam CDLM-08 will consume — guard the create branch (refuse or return without inserting when `verify_only`) before the first external caller trips it.

app/heimdal/meeting_projection.py
```diff
@@ -1037,6 +1064,50 @@ def _rederive_analysis(session_id: str) -> Optional[AnalysisRevision]:
 # ---------------------------------------------------------------------------
 
 
+def _register_analysis_blocks(session_id: str, revision: AnalysisRevision) -> None:
```
RasmusTho, line 1067: P2 (final review r1): unbounded refusal-row spam via derived-id squatting. The user-note route accepts any client-minted `note_block_id`, including the reserved derived namespace (`{session}:analysis:summary`, `{session}:transcript:{seq}`). If a user note squats `{session}:analysis:<type>`, every subsequent segment admission re-runs `_rederive_analysis` -> `_register_analysis_blocks`, whose revise on the squatted block is refused and durably recorded — one refusal row per analysis block type per segment, forever, growing `needs_attention` on every projection read (append-only table, no dedupe). The refusal itself is correct (fail-closed, user content preserved), but the loop is self-inflicted noise: either reject reserved block-id namespaces at the user-note route, or dedupe/skip registration once a target is known-squatted. Same squat also silently prevents the transcript/analysis block from ever being registered (log.warning only).

app/heimdal/meeting_blocks.py
```diff
@@ -0,0 +1,1171 @@
...
+                    return _refuse(
+                        store,
+                        session_id=session_id,
+                        block_id=block_id,
+                        writer=writer,
+                        action=action,
+                        reason="user_note blocks are writable only by the user's editor "
+                        "identity (INV-CDLM-6)",
+                    )
+            elif writer.kind != WRITER_DERIVED or not writer.engine:
```
RasmusTho, line 883: P2: The verify_only dry-create branch omits the `transcript_segment` role check that the real create path enforces. A `verify_only=True` CREATE probe for `block_type=transcript_segment` by a derived non-ASR writer (e.g. the analysis writer) returns `allowed=True`, while the real create refuses with "transcript_segment blocks are created only by the ASR derivation (CDLM-06)". Verified empirically on this head. No in-repo caller currently uses verify_only+CREATE (the user-note route probes only REVISE), but `apply_block_write` is documented as the one shared seam CDLM-08 consumes, and a probe that authorizes what the real write refuses breaks the fail-closed contract — a future replay-ack path built on the probe could acknowledge a create the guard would refuse. Fix: add `if block_type == TYPE_TRANSCRIPT_SEGMENT and writer.role != "asr": return _refuse(...)` to the dry branch (or restructure so both paths share one authorization function, as `_authorize_target` already does for existing blocks).

RasmusTho (commented):
Fresh-review round: one new P2 (inline).

app/heimdal/meeting_projection.py
```diff
@@ -935,6 +946,21 @@ def _derive_segment_inner(
         if stored is not None:
             derivation = stored
 
+    if derivation.status == DERIVATION_OK:
```
RasmusTho, line 949: P2 (final review round 2): transcript block registration is skipped on the derivation-replay short-circuit, so the block registry silently misses transcript_segment blocks in two reachable cases.

`_derive_segment_inner` registers the `{session}:transcript:{seq}` block only on the fresh-derivation path (this hunk). The early return at the top of the function (`existing.status == DERIVATION_OK` → `_rederive_analysis(); return existing`) never reaches this registration, so:

1. **Duplicate content in normal operation**: derivations are keyed by `content_sha256`. Two segments with identical bytes (e.g. two silent segments) share one derivation row; the second seq short-circuits on the stored row and its `{session}:transcript:{seq2}` block is never created — no crash required.
2. **Crash window is never healed**: a crash between `insert_derivation` and this `apply_block_write` leaves the block permanently unregistered; every client resend takes the short-circuit branch. Note the analysis-revision path explicitly heals this exact window on replay ("this is the only writer that can heal the missing revision") — the transcript block registration has no equivalent.

Impact today is registry completeness only (the projection transcript itself reads the derivation/ledger tables), but `page_blocks` under-reports transcript blocks, and CDLM-08 finalization is documented to consume this same seam. Suggested fix: mirror the healing pattern — perform the guard-registered block create on the replay branch too (it is already an idempotent replay in the guard when content matches).

RasmusTho: Dispositioned as deferred P2 on the Known Defects registry #4172 (entry KD-F9588AFBC165). Promotion trigger: CDLM-08 (#4388) consumes the block registry for finalization and will move the registration onto the replay path before relying on registry completeness. No code change in this PR for this finding.

PR #4431, opened 2026-07-30
feat(ops): provision HEIMDAL_RAW_STORE_KEY to the api process

RasmusTho:
Governing-Issue: #4422

Fixes #4422

Final-Review-Rounds: 2

## Summary
Makes the shipped media/screen ingress lanes able to admit in a running channel: declares the `heimdal-api-ingress` consumer for `heimdal.raw-store-key` in the host secret contract (dev/test/prod; capture-watch unchanged); the governed deploy wrapper adds the api secret layer for every `up` that includes the api service, **degrade-visibly** — an availability precheck proves the contract and the Keychain item resolve (value discarded, never printed) before any wrap is added, so an unprovisioned key skips the layer loudly instead of failing the deploy (the bootstrap mechanism stays untouched and fail-closed); the materialized handle is re-exported as `HOST_SECRET_RUNTIME_ENV_FILE_API` so it survives the nested capture-watch bootstrap's environment scrub, and the base `api` Compose service consumes it via a dedicated optional env-file layer. An api-lifespan startup preflight detects a missing `HEIMDAL_RAW_STORE_KEY` before first use, logs it loudly, and reports both ingress lanes `unavailable` on `/api/status` (named reasons only, never exception text) while every other API function keeps serving; the request-time `raw_store_key_unavailable` contract is unchanged. Both the precheck and the bootstrap run anchored to the deploy root so the contract and app package are this deploy's, never the caller's CWD. The single remaining operator step — placing the actual key material into `{channel}:heimdal-api-ingress:heimdal.raw-store-key` under service `yggdrasil.host-secrets` — is stated in `docs/STATUS.md`.

## Claim receipt
`pickup-claim-complete issue=4422 branch=feat/4422-rawkey-api-provisioning worktree=/Users/rasmus/worktrees/rawkey-4422 coordination_mode=dispatcher-backed fallback_reason=none task_id=github-RasmusTho--agentic-pkm-mvp-issue-4422 lease_id=lease-b6bde9a3 holder=cdlm-hub-coordinator evidence=verified-dispatcher-lease`

## Risk gates
- TCD risk surfaces: security, credential-durability. `scripts/review_before_ci_gate.py` satisfied at push SHA.
- Convergence packet (`heimdal-api-ingress-secret`) built; pre-CI independent review found one P0 (the mechanism's degrade flag is inert for kind `raw-store-key`, so a missing Keychain item would have failed the whole deploy) and one P1 (contract/app resolution from the caller's CWD — silent skip or cross-repo secret-path divergence) plus a P2 (textual-only deploy proof) — all fixed; post-fix convergence re-review CLEAN (P3 residuals recorded: precheck uses the mechanism's private lookup helper; keychain-locked conflates with unprovisioned under the degrade posture; millisecond precheck→bootstrap TOCTOU documented).

## BuilderOps Routing
- Reason: no plan divergence with a repairable upstream artifact; claim state carried by the lease receipt above.

## Notes
Validation:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -p pytest_asyncio.plugin -q tests/ops/test_host_secret_contract.py tests/deploy/test_deploy_channel_script.py tests/heimdal/test_media_ingress.py` — 104 passed, including the two new AC tests (the missing-key test drives the real api lifespan startup path via TestClient's context manager and asserts /api/status lane state, an unrelated route serving, and the unchanged request-time 500; the deploy test functionally proves the handle-rename shim survives a simulated inner-bootstrap scrub and probes the gating helper's call shapes). The existing 38 media-ingress tests stay green.
- Full `pytest -m "not pg and not alpha_llm" -n auto tests` — 11850 passed; failures were the known pre-existing host-environment set and the two parallel-only flakes (pass serially).
- `ruff` clean; `mypy app` clean (833 files); `docs_guard.py` OK; `bash -n` clean on the deploy wrapper.
---

### DIFF
--- a/scripts/lib/deploy_channel_compose.sh
+++ b/scripts/lib/deploy_channel_compose.sh
@@ -137,9 +137,10 @@ _deploy_channel_needs_dev_capture_secret() {
 # The api process is a declared consumer of heimdal.raw-store-key (#4422): the
 # governed media/screen ingress lanes encrypt through the raw store. Bootstrap
 # fires whenever `up` includes the api service (named explicitly, or implied by
-# an un-filtered `up`). Posture is degrade-visibly, never fail-deploy:
-# --run-on-credential-unavailable launches the stack without the binding, and
-# the api startup preflight reports the ingress lanes unavailable.
+# an un-filtered `up`). Posture is degrade-visibly, never fail-deploy: the
+# availability precheck proves the contract and Keychain item resolve before
+# any wrap is added, so an unprovisioned key skips the layer loudly and the
+# api startup preflight reports the ingress lanes unavailable.
 # The api ingress secret layer is additive and degrade-visibly: when the host
 # secret contract cannot be loaded or does not declare the api consumer in this
 # environment (e.g. a harness root without config/secrets), the deploy proceeds
--- a/scripts/lib/deploy_channel_compose.sh
+++ b/scripts/lib/deploy_channel_compose.sh
@@ -148,18 +148,40 @@ _deploy_channel_needs_dev_capture_secret() {
 # be prepared.
 _deploy_channel_api_ingress_bootstrap_available() {
   local channel="${1:?channel required}"
-  if "${PYTHON:-python3}" - "${channel}" <<'PY' 2>/dev/null
+  local root="${2:?repo root required}"
+  # Runs from ${root} with ${root} on PYTHONPATH: the contract path is
+  # repo-relative and the app package must be THIS deploy's, never whatever
+  # checkout the caller's shell happens to sit in.
+  # The precheck also proves the Keychain item is actually resolvable
+  # (value discarded, never printed): the bootstrap mechanism is fail-closed
+  # for kind raw-store-key, so wrapping compose while the item is missing
+  # would fail the whole deploy — the opposite of this slice's
+  # degrade-visibly posture. Missing item ⇒ skip the layer loudly; the api
+  # startup preflight then reports the ingress lanes unavailable.
+  if (
+    cd "${root}" \
+      && PYTHONPATH="${root}${PYTHONPATH:+:${PYTHONPATH}}" "${PYTHON:-python3}" - "${channel}" <<'PY' 2>/dev/null
 import sys
+
+from app.ops.host_secret_bootstrap import _security_keychain_lookup
 from app.ops.host_secret_contract import load_host_secret_contract
 
-load_host_secret_contract().require_declared(
+contract = load_host_secret_contract()
+contract.require_declared(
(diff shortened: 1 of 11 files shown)


scripts/lib/deploy_channel_compose.sh
```diff
@@ -269,6 +338,32 @@ deploy_channel_compose() {
...
+      # absolute or compose-file-relative, so the cd is inert for Compose.
+      # Never echoes or logs a secret value; the shim exports only a file
+      # path, and the file is bootstrap-owned and removed after Compose
+      # returns.
+      compose_command=(
+        sh -c 'cd "$1" && export PYTHONPATH="$1${PYTHONPATH:+:${PYTHONPATH}}" && shift && exec "$@"' _ "${root}"
+        "${PYTHON:-python3}" -m app.ops.host_secret_bootstrap
+        --channel "${channel}"
+        --consumer heimdal-api-ingress
+        -- sh -c 'export HOST_SECRET_RUNTIME_ENV_FILE_API="${HOST_SECRET_RUNTIME_ENV_FILE:-/dev/null}"; exec "$@"' _
```
RasmusTho: **P1 (final review r1): cross-consumer secret delivery via the un-unset shared handle.** The shim copies `HOST_SECRET_RUNTIME_ENV_FILE` to `HOST_SECRET_RUNTIME_ENV_FILE_API` but leaves the shared name exported. On a dev unfiltered `up -d`, the api gate fires but `_deploy_channel_needs_dev_capture_secret` does not (it requires `heimdal-capture-watch` named explicitly), so no inner bootstrap scrubs the shared name — and `docker-compose.dev.yml`'s `heimdal-capture-watch` env_file (`${HOST_SECRET_RUNTIME_ENV_FILE:-/dev/null}`) interpolates to the **heimdal-api-ingress consumer's materialized file**. Capture-watch previously got `/dev/null` in this scenario; now it receives the api consumer's layer — the accounts are per-consumer (`{channel}:{consumer}:{secret}`) so the values can differ. This breaches two issue constraints: the key must reach only its declared consumer path, and capture-watch provisioning must be unchanged. Fix: `unset HOST_SECRET_RUNTIME_ENV_FILE` in the shim after the rename (the deploy_channel_script test's scrub simulation already assumes the shared name ends scrubbed).

RasmusTho: Fixed in 91c9964111029b2242c8a45f6d2243a47ad69a8e: the rename shim now unsets HOST_SECRET_RUNTIME_ENV_FILE after exporting the api handle, so the shared name can never reach the capture-watch service's env_file chain on a dev unfiltered up. The functional test now proves both shapes: shim-only (dev unfiltered) scrubs the shared handle, and the nested-scrub shape still preserves the renamed handle.

PR #4499, opened 2026-08-01
Prove the Design Hub production acceptance matrix (CDH-06)

RasmusTho:
## Change Lane
- [ ] Docs authoring lane
- [ ] Governance lane

Final-Review-Rounds: 1

Governing-Issue: #4313

## Linked Issue
Closes #4313

## SBS Impact
- Primary subsystem: Builder System / CKM acceptance
- Secondary subsystem(s): BuilderOps design-run and model-inquiry regressions
- Write class: acceptance evidence and parent handoff receipts
- Persistence impact: none; tests write only to tmp_path, and no production module changed
- Derived/rebuildable impact: final cockpit HTML and the print evidence remain rebuildable from fixed inputs
- New or changed contract: production-path success/refusal acceptance matrix; no new runtime contract
- Owner-doc impact: deliberately none in this PR; independent parent acceptance on #4131 must authorize the later docs-only promotion
- Transition debt impact: closes unverified integration risk for the CKM Design Hub capability
- Boundary risk: child cannot close parent #4131 or claim supported owner truth; both spec State lines stay pre-acceptance

## Owner-Doc Writeback

## Summary
- Adds tests/builderops/test_design_hub_acceptance.py: the production-path success/refusal acceptance matrix CDH-06 requires, driven through the real DesignRunGovernance, the real DesignAgentAdapterRegistry, the shipped repo-governed policy and Yggdrasil token bytes, the real SQLite receipt store, and the real `ckm overview --cockpit --design-hub` CLI composition root.
- Extends tests/builderops/ckm/test_design_cockpit.py::test_design_hub_projection_preserves_direction_b_authority so the deterministic / JS-off / print-complete / inert-HTML contract is proven across every terminal and refusal state plus one excluded run, not only the pending and succeeded states CDH-05 exercised.
- Records the delivery state that actually exists in the two local CKM Design Hub spec State lines, and deliberately promotes no CKM or Builder System owner doc.

## BuilderOps Routing
- Records/projections/receipts: LearningSignal lrn_20260801011005_784605b1 (workflow_divergence); design-run BuilderOpsReceipts are exercised in tests only and never written to a production store
- Reason: The acceptance matrix surfaced a real contract-text divergence: the spec asks for one governed success per registered adapter, but INV-CDH-5A makes claude-design-via-claude-code structurally headless-incapable, so that phrasing and the delivered behaviour cannot both be read literally.

## Notes
See the PR discussion below for the full validation ledger, the deterministic print receipt, and the disclosed limitations.

## Acceptance criteria — `Verify:` results at head `2f92cc4c935b39d9ba31f67dcbe5de16bfdfd12e`

| AC | Verify target | Result |
|---|---|---|
| 1 | `tests/builderops/test_design_hub_acceptance.py::test_design_hub_production_matrix_is_fail_closed` | PASS |
| 2 | `tests/builderops/test_design_hub_acceptance.py::test_design_hub_outputs_remain_unaccepted_builder_material` | PASS |
| 3 | `tests/builderops/ckm/test_design_cockpit.py::test_design_hub_projection_preserves_direction_b_authority` | PASS |
| 4 | `tests/builderops/test_design_hub_acceptance.py::test_focused_regression_suites_are_bound_to_pr_head` | PASS |
| 5 | `runtime receipt: ckm_design_hub.cdh06_completion_handoff.v1` | Local spec State writeback in this diff; the child receipt and the complete conditional-acceptance ledger post on #4131 at merge |

## The acceptance matrix

Every refusal case asserts **zero** provider turns; every executed case asserts **exactly one** and no fallback to a second registered adapter.

| Case | Terminal state | Refusal code | Provider turns |
|---|---|---|---|
| no adapters (dormant `from_declared_sources`) | `unavailable` | `adapter_unavailable` | 0 |
| unknown adapter | no run created | `UnknownDesignAgentError` at the registry boundary | 0 |
| unavailable adapter (`claude-design-via-claude-code`) | `unavailable` | `adapter_unavailable` | 0 |
| policy denial (source refs > `max_source_refs`) | `denied` | `admission_denied` | 0 |
| approval pending | `approval_pending` | `approval_pending` | 0 |
| malformed request | `malformed` | `malformed_request` | 0 |
| missing Yggdrasil receipt (visual) | refused at contract construction | — | 0 |
| drifted Yggdrasil token hash | `denied` | `yggdrasil_token_drift` | 0 |
| stale Yggdrasil receipt | `denied` | `yggdrasil_gate_stale` | 0 |
| current Yggdrasil parity (control) | `approval_pending` | — | 0 |
| stale approval (recorded before admission) | refused | `DesignRunApprovalRequiredError` | 0 |
| revoked approval | refused | `DesignRunApprovalRequiredError` | 0 |
| non-causal approval (approved after named start) | refused | `DesignRunApprovalRequiredError` | 0 |
| mismatched approval at `execute_exact` | refused | `DesignRunApprovalRequiredError` | 0 |
| matching approval at `execute_exact` (control) | `succeeded` | — | 1 |
| governed success, `codex` | `succeeded` | — | 1 |
| governed success, `fable` | `succeeded` | — | 1 |
| timeout | `timed_out` | `timed_out` | 1 |
| provider failure | `failed` | `execution_failed` | 1 |
| malformed provider output (digest does not bind content) | `failed` | `execution_failed` | 1 |
| correctly digested handoff bound to a **foreign** run | `failed` | `execution_failed` | 1 |
| real CLI `ckm overview --cockpit --design-hub` | renders the dormant posture | — | 0 additional |

Seven distinct terminal/refusal states are asserted as a set: `unavailable`, `denied`, `approval_pending`, `malformed`, `timed_out`, `failed`, `succeeded`.

## Provider posture — nothing was provisioned

No `builderops-design-run` credential grant exists and none was created. The production registry at this head reports exactly:

```
claude-design-via-claude-code  available=False  interactive_subscription_only
codex                          available=False  model_access_unavailable
fable                          available=False  model_access_unavailable
```

Every refusal row above runs against that genuinely dormant registry. The governed-success rows run through the **one** doubled seam — the neutral ADR-0064 `ModelAccessResolver` port and the `ModelTurnAdapter` behind it — because with no grant, provisioning a provider would change the very posture under test. Everything else on the path is production: real `DesignRunGovernance`, real `DesignAgentAdapterRegistry`, the shipped `config/builderops/design_run_policy.json` and `companion-ui/companion-app/colors_and_type.css` **bytes**, real `SqliteBuilderOpsStore`, real `build_design_hub_projection`, real `render_overview_html`, and the real Click composition root. No credential is read; `_SubstrateResolver` never reaches the host secret contract.

**Disclosed limitation.** `claude-design-via-claude-code` is hard-refused as `interactive_subscription_only` inside `DesignAgentAdapterRegistry.resolve_group` before any adapter lookup, so it can never reach a governed success by construction (INV-CDH-5A). The matrix therefore proves one governed success per **headless-registered** adapter (`codex`, `fable`) and an exact, no-fallback refusal for the interactive-only route. The spec's "one governed success per registered adapter" phrasing and INV-CDH-5A cannot both be read literally; that divergence is filed as `LearningSignal lrn_20260801011005_784605b1` for upstream text repair, and it is not repaired here because a child slice does not own that acceptance text.

## Deterministic print evidence — method stated plainly

**No live browser render was performed.** `preview_start` timed out after 300s with no attached browser in this environment, the same limit CDH-05 hit. I did not take a screenshot and am not claiming one. The evidence below is the **deterministic print receipt** route the issue Scope allows.

Final cockpit HTML generated three times from fixed inputs through the real CLI (`ckm overview --cockpit --design-hub --repo-root`, generation time pinned):

- SHA-256 `cab8349df2c11eda0e2372fc441d3132062f0eb84a4fe7683e92bb6c806f78dc` — identical across all three runs
- 27,810 bytes; `<script` count 1 (the pre-existing Direction B filter script, unchanged)
- design-hub section 4,090 bytes, containing no `<form`, `<button`, `<select`, `<textarea`, `<script`, `<a `, ` :not(summary) { display:block !important; }`, `details::details-content { content-visibility:visible !important; }`, and `[hidden] { display:block !important; }`, and the design-hub section contains no `hidden` attribute and no `display:none`
- JS-off completeness: every run id, every causal receipt id, and every adapter id is asserted present in the raw server-rendered markup

## Yggdrasil receipt

`ckm_design_hub.yggdrasil_handoff.v1` re-verified, not re-derived: repo token source `companion-ui/companion-app/colors_and_type.css`, SHA-256 `7d8cdd49f59061f895959159a08e82348e7e02eb8b8ba7426020a50c7fa915b1` — **unchanged** from the value CDH-05 recorded, parity pass. This PR introduces no new visual token, primitive, or markup; the extended cockpit test only asserts against the renderer CDH-05 already shipped.

## Validation actually run at head `2f92cc4c9`

| Check | Result |
|---|---|
| `python3 -m pytest -q tests/builderops` | **1087 passed** in 177s |
| `python3 -m pytest -q tests/builderops/test_design_hub_acceptance.py tests/builderops/ckm/test_design_cockpit.py` | 9 passed |
| `python3 -m ruff check app tests companion-ui/companion-app` | All checks passed |
| `python3 -m mypy app` | Success, 848 source files |
| `python3 scripts/public_seam_lint.py --mode gate` | `secret_hits: 0`, `uncovered_hits: 0`, `stale_rows: 0`, `migration_pending: 0` |
| `python3 scripts/docs_guard.py` | OK |

## Proof-of-proof: mutation probes

An acceptance slice is worthless if its assertions are vacuous, so the matrix was mutation-tested against the unchanged production modules. 11 probes were run and reverted; **11 are now detected**:

approval fail-open (`_usable_approval`), interactive-only hard refusal removed, artifact digest not validated, exact-start-evidence not enforced, `max_source_refs` cap removed, `yggdrasil_token_drift` removed, `yggdrasil_gate_stale` removed, `malformed_request` check removed, design-hub run exclusion removed, `adapter_unavailable`/`unavailable` downgraded, and the returned-handoff binding equality guard removed.

The last one **survived** the first review round and was the P2 finding that prompted the new foreign-binding matrix case; it is now detected.

## Delivery path

Full path per `AGENTS.md :: Proportional delivery` — the TCD risk surfaces `external-api`, `credential-durability`, and `state-machine` are the subject under proof.

- Mechanism convergence packet built for key `builderops-design-run :: admission-approval-execution state machine and provider fencing` (invariants, valid/terminal/indeterminate/compensated states, transitions and writers, durability ordering at the accepted-start fence, producers/consumers/recovery, lock and lease ordering, and the full test map).
- Independent high-capability review round 1 at SHA `c7e2ff733`: **BLOCKING** — 2 P1, 1 P2, 5 P3.
  - P1: the spec `State:` lines certified a CDH-06 receipt on #4131 that does not exist. Fixed — they now state only the receipts that exist and say the CDH-06 receipt posts on merge.
  - P1: `FOCUSED_REGRESSION_HEAD_DIGEST` was a new permanent governance mechanism with no contract demand, no named replacement, and no review-by date, and was trivially satisfiable while the named suites failed. Removed. Head binding is now discharged by the mechanism that already exists — `scripts/select_pr_tests.py` expands this diff to the `builder_system` lane, which runs the whole `tests/builderops` tree, and `pr-evidence-pack` snapshots those check runs keyed on `pull_request.head.sha`.
  - P2: the returned-handoff binding guard was uncovered by this slice's matrix. Fixed with the foreign-binding case.
  - P3 taken: the two CLI guard invocations now assert the exact `ClickException` text.
- Independent review round 2 (pre-CI convergence) at SHA `c7e2ff733` -> `6a3736d61`: **CLEAN**, with the P1-2 replacement claims verified against `scripts/select_pr_tests.py` and `.github/workflows/pr-evidence-pack.yml` and the P2 fix mutation-verified.
- Final independent review gate, round 1, at SHA `6a3736d61`: **CLEAN** (no P0/P1), 3 P2 + 5 P3. Because this slice's whole deliverable is proof quality, the three P2s were fixed rather than deferred — most importantly a surviving mutant against INV-CDH-9/AC3, where deleting the two `sorted()` calls in `app/builderops/ckm/design_hub.py` left the suite green because determinism was only compared inside a single interpreter with a fixed hash seed. The total order of runs, excluded run ids, and adapters is now pinned explicitly.
- Final independent review gate, round 2, at SHA `2f92cc4c9`: **CLEAN**. All three P2 fixes independently mutation-verified, including the ordering mutant failing under `PYTHONHASHSEED` 0-5 against the new test and passing against the old one.
- `Final-Review-Rounds: 1` — this PR changes no runtime surface (the diff is tests plus two spec State lines), and `verification-and-closure` states that a test-enforcement change carrying a high-risk label alone does not qualify as a runtime surface. The convergence rounds above are the pre-expensive gate, not the final gate.

## Known residual, disclosed

The literal HTML SHA-256 above is reproducible **within** a run — three consecutive CLI invocations from fixed inputs produced byte-identical output, which is the INV-CDH-9 claim the tests assert — but it is not reproducible **across machines**, because the fixture's sandbox path and generated run identifiers are unpinned. It is a local receipt, not a portable one.

`_AUTHORITY_CROSSING_TOKENS` is a source-text grep and overclaims slightly: `design_agent_adapters.py` transitively reaches `load_runtime_secret_values` through `BuilderModelAccessResolver`. No matrix case invokes it — that symbol is reachable only from `credential_value` — but the assertion proves the absence of a direct textual reference, not of a transitive path.

Of the three ordering assertions added for the final-gate P2, only the runs assertion is load-bearing: at the fixture's cardinality the adapters and excluded-run-id sequences have one element each, so those two lines cannot currently fail. Adapter ordering is covered by `test_selector_uses_registered_adapter_availability`; `excluded_run_ids` ordering is not independently enforced. Recorded as P3, not repaired.

`test_focused_regression_suites_are_bound_to_pr_head` is named for a head binding it no longer performs itself; after the P1-2 fix it proves that the named focused-suite set is complete and that every acceptance `Verify:` target resolves at the head where it runs, and its docstring says exactly that. The name is the AC4 `Verify:` target in the governing spec, so renaming it would need a coordinated spec edit and is deliberately left to the parent acceptance audit.

## Explicitly not done

Parent #4131 is **not** closed, relabelled, or otherwise mutated by this PR — a child may not close its parent. No CKM or Builder System owner doc is promoted; `README.md :: Capability acceptance` stays fully unticked. No provider was provisioned, no credential read or requested, no provider ranking or fallback added, no GitHub mutation from the capability, no Product/Runtime write, no release, no promotion.

### DIFF
--- a/tests/builderops/ckm/test_design_cockpit.py
+++ b/tests/builderops/ckm/test_design_cockpit.py
@@ -8,6 +8,8 @@ from html.parser import HTMLParser
 from pathlib import Path
 from typing import Any, Mapping
 
+import pytest
+
 from app.builderops.ckm.design_hub import (
     DesignHubProjection,
     build_design_hub_projection,
@@ -30,6 +32,8 @@ from app.builderops.design_run_governance import (
     DESIGN_RUN_RECEIPT_EVENT_TYPE,
     DesignRunGovernance,
     DesignRunReceiptEvent,
+    DesignRunUnavailableError,
+    load_design_run_policy,
 )
 from app.builderops.model_access_resolver import BuilderModelAccessResolver
 from app.builderops.store import SqliteBuilderOpsStore
@@ -87,8 +91,15 @@ class _RecordingAdapter:
     artifact_content: str = "bounded, cited design-run output"
     provider: str = "fixture-provider"
     model: str = "fixture-model"
+    # Per-run terminal behaviour: "success" (default), "timeout", or "failure".
+    modes: dict[str, str] = field(default_factory=dict)
 
     def execute(self, request: Mapping[str, Any]) -> AdapterResult:
+        mode = self.modes.get(str(request["run_id"]), "success")
+        if mode == "timeout":
+            raise TimeoutError("fixture design agent timed out")
+        if mode == "failure":
+            raise RuntimeError("fixture design agent failed")
         content_hash = hashlib.sha256(
             self.artifact_content.encode("utf-8")
         ).hexdigest()
@@ -112,6 +123,8 @@ class _RecordingRegistry:
 
     adapter: _RecordingAdapter = field(default_factory=_RecordingAdapter)
     design_agent_id: str = "cockpit-fixture-agent"
+    # Run ids for which the exact requested route is unavailable at start time.
+    unavailable_run_ids: frozenset[str] = frozenset()
 
     def descriptors(
         self, *, run_id: str
@@ -141,6 +154,8 @@ class _RecordingRegistry:
     def select(
         self, design_agent_id: str, *, run_id: str
     ) -> ResolvedDesignAgentAdapter:
+        if run_id in self.unavailable_run_ids:
+            raise RuntimeError(f"design agent unavailable: {design_agent_id}")
         return ResolvedDesignAgentAdapter(
             design_agent_id=design_agent_id,
             descriptor=self.descriptors(run_id=run_id)[0],
@@ -166,18 +181,26 @@ def _brief(
     *,
     brief_id: str,
     deliverable: str = "content_review",
+    source_ref_count: int = 1,
 ) -> CuratedDesignBrief:
+    source_refs = tuple(
+        sorted(
+            (
+                DesignSourceRef(
+                    source_type="ckm_observation",
+                    source_id=f"ckm:observation:cockpit-fixture-{index:03d}",
+                    content_hash=SHA_A,
+                )
+                for index in range(source_ref_count)
+            ),
+            key=lambda ref: ref.identity,
+        )
+    )
     return CuratedDesignBrief(
         brief_id=brief_id,
         projection_id="ckm.projection.cockpit",
         requested_deliverable=deliverable,
-        source_refs=(
-            DesignSourceRef(
-                source_type="ckm_observation",
-                source_id="ckm:observation:cockpit-fixture",
-                content_hash=SHA_A,
-            ),
-        ),
+        source_refs=source_refs,
         constraints=("Use explicit evidence only.",),
         yggdrasil_gate_receipt=None,
         non_visual_exemption=True,
@@ -187,7 +210,7 @@ def _brief(
 def _repo_root(tmp_path: Path) -> Path:
     root = tmp_path / "repo"
     policy_path = root / "config/builderops/design_run_policy.json"
-    policy_path.parent.mkdir(parents=True)
+    policy_path.parent.mkdir(parents=True, exist_ok=True)
     source_policy = (
         Path(__file__).parents[3] / "config/builderops/design_run_policy.json"
     )
@@ -195,12 +218,17 @@ def _repo_root(tmp_path: Path) -> Path:
     return root
 
 
-def _governance(tmp_path: Path, *, db_name: str = "cockpit.sqlite3") -> DesignRunGovernance:
+def _governance(
+    tmp_path: Path,
+    *,
+    db_name: str = "cockpit.sqlite3",
+    registry: _RecordingRegistry | None = None,
+) -> DesignRunGovernance:
     store = SqliteBuilderOpsStore(tmp_path / db_name)
     store.initialize()
     return DesignRunGovernance(
         store=store,
-        registry=_RecordingRegistry(),
+        registry=registry or _RecordingRegistry(),
         repo_root=_repo_root(tmp_path),
         lease_ttl_seconds=30,
     )
@@ -213,12 +241,18 @@ def _submit(
     adapter: DesignAgentDescriptor | None = None,
     deliverable: str = "content_review",
     evaluated_at: str = T1,
+    source_ref_count: int = 1,
+    request_brief: CuratedDesignBrief | None = None,
 ) -> None:
-    brief = _brief(brief_id=f"{run_id}.brief", deliverable=deliverable)
+    brief = _brief(
+        brief_id=f"{run_id}.brief",
+        deliverable=deliverable,
+        source_ref_count=source_ref_count,
+    )
     resolved_adapter = adapter or _descriptor()
     request = governance.build_request(
         request_id=f"{run_id}.request",
-        brief=brief,
+        brief=request_brief or brief,
         adapter=resolved_adapter,
         requested_at=T0,
     )
@@ -493,16 +527,114 @@ def test_default_overview_is_unchanged_without_design_hub(tmp_path: Path) -> Non
     assert '<section class="design-hub"' in enabled_once
 
 
+def _every_terminal_state_governance(tmp_path: Path) -> DesignRunGovernance:
+    """One store carrying every distinct terminal/refusal state at once.
+
+    CDH-06 (issue #4313) requires the deterministic / JS-off / print /
+    inert-HTML contract to hold across the *whole* state space, not only the
+    pending and succeeded states CDH-05 exercised.
+    """
+
+    adapter = _RecordingAdapter(
+        modes={
+            "run.cockpit.timed-out": "timeout",
+            "run.cockpit.failed": "failure",
+        }
+    )
+    registry = _RecordingRegistry(
+        adapter=adapter,
+        unavailable_run_ids=frozenset({"run.cockpit.unavailable"}),
+    )
+    governance = _governance(tmp_path, registry=registry)
+
+    # approval_pending
+    _submit(governance, run_id="run.cockpit.pending")
+
+    # denied — the repo-governed policy caps explicit source refs.
+    policy = load_design_run_policy(governance.repo_root)
+    _submit(
+        governance,
+        run_id="run.cockpit.denied",
+        source_ref_count=policy.max_source_refs + 1,
+    )
+
+    # malformed — the request does not bind its exact submitted inputs.
+    _submit(
+        governance,
+        run_id="run.cockpit.malformed",
+        request_brief=_brief(brief_id="brief.cockpit.foreign"),
+    )
+
+    # unavailable — the exact requested route cannot be selected at start.
+    _submit(governance, run_id="run.cockpit.unavailable")
+    governance.approve(
+        run_id="run.cockpit.unavailable",
+        approval_id="run.cockpit.unavailable.approval",
+        approved_at=T2,
+    )
+    with pytest.raises(DesignRunUnavailableError):
+        governance.execute(
+            run_id="run.cockpit.unavailable", started_at=T3, completed_at=T4
+        )
+
+    # timed_out and failed — distinct typed terminal failures.
+    for run_id in ("run.cockpit.timed-out", "run.cockpit.failed"):
+        _submit(governance, run_id=run_id)
+        governance.approve(
+            run_id=run_id, approval_id=f"{run_id}.approval", approved_at=T2
+        )
+        governance.execute(run_id=run_id, started_at=T3, completed_at=T4)
+
+    # succeeded — one governed success with a validated handoff.
+    _run_to_succeeded(governance, run_id="run.cockpit.succeeded")
+
+    # excluded — an unverifiable causal chain is refused entirely.
+    _inject_dangling_receipt(governance, run_id="run.cockpit.excluded")
+    return governance
+
+
 def test_design_hub_projection_preserves_direction_b_authority(tmp_path: Path) -> None:
-    governance = _governance(tmp_path)
-    _submit(governance, run_id="run.cockpit.inert")
-    _run_to_succeeded(governance, run_id="run.cockpit.inert-succeeded")
+    governance = _every_terminal_state_governance(tmp_path)
     design_hub = build_design_hub_projection(governance=governance)
-    assert design_hub.runs  # non-trivial fixture
+
+    # Every distinct terminal/refusal state is present exactly once, plus one
+    # entirely excluded run.
+    states = {run.run_id: run.state for run in design_hub.runs}
+    assert states == {
+        "run.cockpit.pending": "approval_pending",
+        "run.cockpit.denied": "denied",
+        "run.cockpit.malformed": "malformed",
+        "run.cockpit.unavailable": "unavailable",
+        "run.cockpit.timed-out": "timed_out",
+        "run.cockpit.failed": "failed",
+        "run.cockpit.succeeded": "succeeded",
+    }
+    assert len(set(states.values())) == 7
+    assert design_hub.excluded_run_ids == ("run.cockpit.excluded",)
+
+    # INV-CDH-9 determinism depends on a total order that does not vary with
+    # set/dict iteration order, so pin it explicitly rather than inferring it
+    # from two renders inside one interpreter (which share a hash seed).
+    assert [run.run_id for run in design_hub.runs] == sorted(states)
+    assert list(design_hub.excluded_run_ids) == sorted(design_hub.excluded_run_ids)
+    assert [item.design_agent_id for item in design_hub.adapters] == sorted(
+        item.design_agent_id for item in design_hub.adapters
+    )
 
     ckm_store = _cockpit_store(tmp_path)
     without_design_hub = _render_with_design_hub(ckm_store, None, generated_at=T0)
     with_design_hub = _render_with_design_hub(ckm_store, design_hub, generated_at=T0)
+
+    # Determinism across the full state space: identical explicit generation
+    # time, CKM batch, and validated projection produce byte-identical HTML.
+    assert with_design_hub == _render_with_design_hub(
+        ckm_store, design_hub, generated_at=T0
+    )
+    assert with_design_hub == _render_with_design_hub(
+        ckm_store,
+        build_design_hub_projection(governance=governance),
+        generated_at=T0,
+    )
 
     safety_before = _parse_html_safety(without_design_hub)
     safety_after = _parse_html_safety(with_design_hub)
@@ -531,5 +663,37 @@ def test_design_hub_projection_preserves_direction_b_authority(tmp_path: Path) -
     for adapter in design_hub.adapters:
         assert f'data-design-hub-adapter="{adapter.design_agent_id}"' in with_design_hub
 
-    # No color-only meaning: every state/availability signal has a textual label.
-    assert "state: succeeded" in with_design_hub or "state: approval_pending" in with_design_hub
+    # No color-only meaning: every state has a textual label in the markup, and
+    # every typed refusal names its code and message as text.
+    for run in design_hub.runs:
+        assert f"state: {run.state}" in with_design_hub
+        if run.refusal_code is not None:
+            assert f"<code>{run.refusal_code}</code>" in with_design_hub
+            assert run.refusal_message is not None
+    assert {
+        run.refusal_code for run in design_hub.runs if run.refusal_code
+    } == {
+        "approval_pending",
+        "admission_denied",
+        "malformed_request",
+        "adapter_unavailable",
+        "timed_out",
+        "execution_failed",
+    }
+
+    # Print completeness: the only collapsible container the design hub uses is
+    # <details>, and the stylesheet forces its content open for print, so no
+    # projected evidence can be lost on paper.
+    assert "@media print" in with_design_hub
+    print_css = with_design_hub.split("@media print", 1)[1]
+    assert "details > :not(summary) { display:block !important; }" in print_css
+    assert "details::details-content { content-visibility:visible !important; }" in print_css
+    assert "[hidden] { display:block !important; }" in print_css
+    assert "hidden" not in design_hub_section
+    assert "display:none" not in design_hub_section
+
+    # Incapable of execution in every state: no script, handler, network, or
+    # storage capability is reachable from any rendered terminal state.
+    assert without_design_hub.count("<script") == with_design_hub.count("<script")
+    for token in ("<script", "javascript:", "srcdoc", "formaction", "http-equiv"):
+        assert token not in design_hub_section
--- a/tests/builderops/test_design_hub_acceptance.py
+++ b/tests/builderops/test_design_hub_acceptance.py
@@ -1,0 +1,1299 @@
+"""CDH-06 (issue #4313): production-path acceptance matrix for the Design Hub.
+
+`docs/CKM_DESIGN_AGENT_INTEGRATION/VALIDATE_DESIGN_HUB_END_TO_END.md ::
+Acceptance Criteria` requires proof on the *composition* path, not on isolated
+unit fixtures. These tests therefore drive the real
+:class:`~app.builderops.design_run_governance.DesignRunGovernance`, the real
+:class:`~app.builderops.design_agent_adapters.DesignAgentAdapterRegistry`, the
+real repo-governed policy and Yggdrasil token files, the real SQLite BuilderOps
+receipt store, and the real cockpit renderer.
+
+Exactly one seam is doubled: the neutral ADR-0064 ``ModelAccessResolver`` port
+and the ``ModelTurnAdapter`` behind it. That is deliberate and is itself part of
+the contract being proven. The shipped production posture has **no**
+``builderops-design-run`` credential grant: `codex` and `fable` resolve to
+``model_access_unavailable`` without reading any credential, and
+`claude-design-via-claude-code` is ``interactive_subscription_only`` and can
+never be headless-available (INV-CDH-5A). Provisioning a provider to obtain a
+"real" success would change the production posture under test, so the governed
+success path is proven through the declared substrate seam while every refusal
+path is proven against the genuinely dormant production registry.
+
+Every case asserts the exact number of provider turns. Refusal cases assert
+zero, governed execution asserts exactly one, and no case ever falls through to
+a second adapter (INV-CDH-5).
+"""
+
+from __future__ import annotations
+
+import ast
+import hashlib
+import json
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Literal
+
+import pytest
+from click.testing import CliRunner
+
+from app.builderops.ckm.design_hub import (
+    build_design_hub_projection,
+    design_hub_digest,
+)
+from app.builderops.ckm.overview_html import CockpitRenderContext, render_overview_html
+from app.builderops.ckm.store import CkmStore
+from app.builderops.cli import builderops
+from app.builderops.design_agent_adapters import (
+    DESIGN_AGENT_IDS,
+    DESIGN_AGENT_ROLE_PROFILES,
+    DesignAgentAdapterRegistry,
+    UnknownDesignAgentError,
+)
+from app.builderops.design_run_contract import (
+    CuratedDesignBrief,
+    DesignSourceRef,
+    DigestBoundAttachmentRef,
+    YggdrasilGateReceipt,
+)
+from app.builderops.design_run_governance import (
+    DESIGN_RUN_POLICY_PATH,
+    DESIGN_RUN_RECEIPT_EVENT_TYPE,
+    DesignRunApprovalRequiredError,
+    DesignRunGovernance,
+    DesignRunGovernanceError,
+    DesignRunUnavailableError,
+    load_design_run_policy,
+)
+from app.builderops.model_access_resolver import ModelAccessResolutionError
+from app.builderops.store import SqliteBuilderOpsStore
+from llm_contract import (
+    AdapterResult,
+    ModelCapabilities,
+    ModelResolutionRequest,
+    ResolvedModelAccess,
+    validate_resolved_group,
+)
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+YGGDRASIL_TOKEN_SOURCE = "companion-ui/companion-app/colors_and_type.css"
+
+SHA_A = "a" * 64
+SHA_B = "b" * 64
+
+T0 = "2026-08-01T09:00:00Z"
+T1 = "2026-08-01T09:01:00Z"
+T2 = "2026-08-01T09:02:00Z"
+T3 = "2026-08-01T09:03:00Z"
+T4 = "2026-08-01T09:04:00Z"
+
+# The three registered domain adapters and the substrate target each resolves
+# to. `claude-design-via-claude-code` is included on purpose: the matrix must
+# prove it is refused rather than silently omitted.
+SUBSTRATE_TARGETS: dict[str, tuple[str, str]] = {
+    "design.codex": ("openai", "gpt-5.6-sol"),
+    "design.claude": ("anthropic", "claude-fable-5"),
+    "design.fable": ("anthropic", "claude-fable-5"),
+}
+HEADLESS_CAPABLE_AGENT_IDS = ("codex", "fable")
+INTERACTIVE_ONLY_AGENT_IDS = ("claude-design-via-claude-code",)
+
+
+# ---------------------------------------------------------------------------
+# Test doubles limited to the one declared substrate seam.
+# ---------------------------------------------------------------------------
+
+
+class _TimeoutSentinel(Exception):
+    """Marker so a timeout double cannot be confused with a generic failure."""
+
+
+@dataclass
+class _CountingAdapter:
+    """Neutral ``ModelTurnAdapter`` that records every provider turn.
+
+    ``mode`` selects the terminal behaviour under test. It never reads a
+    credential, never opens a socket, and never spawns a process.
+    """
+
+    adapter_id: str
+    provider: str
+    model: str
+    mode: Literal[
+        "success",
+        "timeout",
+        "failure",
+        "malformed_output",
+        "foreign_binding",
+    ] = "success"
+    artifact_content: str = "bounded, cited design-run output"
+    calls: list[dict[str, Any]] = field(default_factory=list)
+
+    def execute(self, request: Mapping[str, Any]) -> AdapterResult:
+        self.calls.append(dict(request))
+        if self.mode == "timeout":
+            raise TimeoutError("design agent did not answer in time")
+        if self.mode == "failure":
+            raise _TimeoutSentinel("design agent transport failed")
+        if self.mode == "foreign_binding":
+            # A correctly digested artifact whose handoff binds a *different*
+            # run and evidence set. The digest alone is not enough: the handoff
+            # must bind the exact run it was started for.
+            content_hash = hashlib.sha256(
+                self.artifact_content.encode("utf-8")
+            ).hexdigest()
+            binding = dict(request["handoff_binding"])
+            binding["run_id"] = "run.acceptance.foreign"
+            binding["handoff_id"] = "run.acceptance.foreign.handoff"
+            binding["source_refs"] = [
+                DesignSourceRef(
+                    source_type="ckm_observation",
+                    source_id="ckm:observation:foreign",
+                    content_hash=SHA_B,
+                ).model_dump(mode="json")
+            ]
+            payload = {
+                "schema_version": request["handoff_output_schema_version"],
+                "artifact_content": self.artifact_content,
+                "handoff": {**binding, "content_hash": content_hash},
+            }
+            return AdapterResult(
+                response_text=json.dumps(payload, sort_keys=True),
+                provider_request_id="not-persisted",
+            )
+        if self.mode == "malformed_output":
+            # Structurally parseable but the digest does not bind the content,
+            # so the returned handoff cannot be validated.
+            payload = {
+                "schema_version": request["handoff_output_schema_version"],
+                "artifact_content": self.artifact_content,
+                "handoff": {
+                    **dict(request["handoff_binding"]),
+                    "content_hash": SHA_B,
+                },
+            }
+            return AdapterResult(
+                response_text=json.dumps(payload, sort_keys=True),
+                provider_request_id="not-persisted",
+            )
+        content_hash = hashlib.sha256(
+            self.artifact_content.encode("utf-8")
+        ).hexdigest()
+        payload = {
+            "schema_version": request["handoff_output_schema_version"],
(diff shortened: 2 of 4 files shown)


tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+
+# The three registered domain adapters and the substrate target each resolves
+# to. `claude-design-via-claude-code` is included on purpose: the matrix must
+# prove it is refused rather than silently omitted.
+SUBSTRATE_TARGETS: dict[str, tuple[str, str]] = {
+    "design.codex": ("openai", "gpt-5.6-sol"),
+    "design.claude": ("anthropic", "claude-fable-5"),
+    "design.fable": ("anthropic", "claude-fable-5"),
+}
+HEADLESS_CAPABLE_AGENT_IDS = ("codex", "fable")
```
RasmusTho, line 98: **P2 — the "one governed success per registered adapter" matrix can silently shrink to cover fewer adapters than the contract claims.**

`HEADLESS_CAPABLE_AGENT_IDS` and `INTERACTIVE_ONLY_AGENT_IDS` are hand-maintained literals, and nothing asserts that they *partition* `DESIGN_AGENT_IDS`. Line 593 binds the dormant descriptor list to `DESIGN_AGENT_IDS`, but the governed-success loop at line 916 iterates the literal, not the registry.

Failure scenario: a fourth design agent is registered in `app/builderops/design_agent_adapters.py` reusing an existing role profile (e.g. `design.codex`). `SUBSTRATE_TARGETS` still resolves, `_fresh_adapters()` still builds, every assertion still passes — and the matrix now proves a governed success for 2 of 4 registered adapters while `README.md :: Capability acceptance` still cites this exact test for "one governed success per registered adapter". The evidence quietly covers less than it claims.

This is the same defect *shape* the round-1 P1 was about, and the diff already guards against it for the focused-suite set (`assert discovered <= set(FOCUSED_REGRESSION_SUITES)`, line 1274). The asymmetry is the finding. `_DESIGN_SURFACE_MODULES` (line 1062) has the identical hardcoded-list drift risk.

One-line fix: `assert set(HEADLESS_CAPABLE_AGENT_IDS) | set(INTERACTIVE_ONLY_AGENT_IDS) == set(DESIGN_AGENT_IDS)`.

Not blocking: no adapter is currently uncovered, and a new *role profile* would `KeyError` loudly in `_SubstrateResolver`. Deferring to a bounded follow-up is fine.

tests/builderops/ckm/test_design_cockpit.py
```diff
@@ -493,17 +527,106 @@ def test_default_overview_is_unchanged_without_design_hub(tmp_path: Path) -> Non
...
+        "run.cockpit.succeeded": "succeeded",
+    }
+    assert len(set(states.values())) == 7
+    assert design_hub.excluded_run_ids == ("run.cockpit.excluded",)
 
     ckm_store = _cockpit_store(tmp_path)
     without_design_hub = _render_with_design_hub(ckm_store, None, generated_at=T0)
     with_design_hub = _render_with_design_hub(ckm_store, design_hub, generated_at=T0)
 
+    # Determinism across the full state space: identical explicit generation
```
RasmusTho, line 628: **P2 — surviving mutant: the INV-CDH-9 determinism proof is process-local and cannot detect hash-order nondeterminism.**

Probe (mine, not in the PR's list of 11): in `app/builderops/ckm/design_hub.py`, change `known_run_ids = sorted({...})` to `list({...})` and `runs=tuple(sorted(runs, key=...))` to `tuple(runs)`. **Both named suites still pass — 9 passed.** The same mutation also survives `test_design_hub_production_matrix_is_fail_closed`, whose CLI leg (`assert outputs[0] == outputs[1]`, line 526 of the acceptance module) invokes the CLI twice inside one Python process.

Why it survives: all three determinism assertions here compare renders produced in a *single* interpreter, where `set` iteration order is fixed for the process. Cross-process, it is not: I confirmed the same 8-element run-id set yields 4 different orderings under `PYTHONHASHSEED=1/2/3/4`, and nothing in `pytest.ini`/`pyproject.toml`/CI pins `PYTHONHASHSEED`.

I verified there is **no live bug**: with production code unchanged I rendered the real `ckm overview --cockpit --design-hub` HTML from one fixed store in four separate processes under different hash seeds and got the identical digest `d23dcf8cfa3c8e17677c6e35300940274c9860dd0e03dd76082f56d3b77a6f2b` every time. The production sorts are correct — they are just unguarded.

Since INV-CDH-9 and AC3 are precisely the "byte-identical cockpit HTML from fixed inputs" claim, and this slice's entire product is proof, an unguarded sort is a real gap. Cheapest fix: render the fixture in a subprocess under two different `PYTHONHASHSEED` values and compare digests, or assert `[r.run_id for r in design_hub.runs] == sorted(...)`.

Non-blocking: AC3's literal claim is proven and the shipped behaviour is correct.

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+
+# What binds "these regressions pass" to the exact PR head is CI, not this
+# module: `scripts/select_pr_tests.py` expands this diff to the `builder_system`
+# lane, which runs the whole `tests/builderops` tree — every suite named below —
+# and `.github/workflows/pr-evidence-pack.yml` snapshots those check runs keyed
+# on `pull_request.head.sha`. This test therefore does the one thing that
+# evidence cannot do for itself: prove the *named set is complete and its
+# targets resolve* at the head where it runs, so the head-bound pass evidence
+# cannot quietly cover a smaller set than the slice claims.
+FOCUSED_REGRESSION_SUITES: tuple[str, ...] = (
```
RasmusTho, line 1204: **P2 — `FOCUSED_REGRESSION_SUITES` is narrower than the AC it discharges, and the completeness guard is built not to notice.**

AC4 reads "All focused design-run, **CKM**, and model-inquiry regressions pass on the exact PR head," and #4313's own Suggested Validation runs `pytest -q tests/builderops/ckm tests/builderops/test_model_inquiry_*.py` — i.e. the whole CKM directory. The named set contains **2 of the 19** modules in `tests/builderops/ckm/` (`test_design_cockpit.py`, `test_overview_html.py`); `test_projections.py`, `test_store.py`, `test_query_service.py`, `test_assessment_engine.py`, `test_linkers.py`, `test_metrics.py` and 11 others are absent.

The completeness assertion at line 1274 cannot catch this, because the discovery filter at line 1269 only admits paths whose *filename* contains `design` or `model_inquiry`, plus one hardcoded `test_overview_html.py`. So the guard that exists to stop the named set drifting narrower than the claim is scoped to exactly the part that cannot drift.

Failure scenario: a reader (or the parent acceptance audit) reads AC4 as PASS and concludes the CKM regression surface named by this slice was enumerated; in fact 17 CKM modules were never named. No coverage is actually lost — I confirmed `select_pr_tests.py` expands this diff to `tests/builderops` and I ran the full tree locally — so this is an evidence-scope defect, not a false CI green.

Fix is one line: widen `discovered` to include `tests/builderops/ckm/**/test_*.py`, or narrow the AC/docstring wording to "design-run, design-hub and model-inquiry" (which is what the docstring at line 1266 already says, and which contradicts the AC it discharges).

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+        "tests/builderops/ckm/test_design_cockpit.py",
+        "test_design_hub_projection_preserves_direction_b_authority",
+    ),
+    (
+        "tests/builderops/test_design_hub_acceptance.py",
+        "test_focused_regression_suites_are_bound_to_pr_head",
+    ),
+)
+
+def test_focused_regression_suites_are_bound_to_pr_head() -> None:
```
RasmusTho, line 1245: **P3 — the AC4 `Verify:` target's name asserts a property the function no longer establishes.**

`test_focused_regression_suites_are_bound_to_pr_head` performs no head binding: it checks file existence, that four `Verify:` names resolve via `ast`, and a subset relation. The PR body discloses this honestly under "Known residual" and correctly identifies that the real binding is the `builder_system` CI lane plus `pr-evidence-pack` keyed on `pull_request.head.sha` — I verified both claims and they hold.

Flagging as informational only so the parent acceptance audit does not have to rediscover it: a reader who sees "AC4: PASS" against a test named `..._bound_to_pr_head` will reasonably assume in-test head binding. Renaming needs a coordinated edit to `VALIDATE_DESIGN_HUB_END_TO_END.md :: Acceptance Criteria`, so leaving it to the parent audit is the right call — please just make sure that hand-off actually names it.

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+        "succeeded": "succeeded",
+    }
+    assert len(set(observed.values())) == len(observed)
+
+
+# ---------------------------------------------------------------------------
+# AC2 — outputs stay unaccepted Builder material with validated lineage.
+# ---------------------------------------------------------------------------
+
+_AUTHORITY_CROSSING_TOKENS = (
```
RasmusTho, line 1052: **P3 — `_AUTHORITY_CROSSING_TOKENS` is a source-text grep whose comment overclaims, and it is brittle.**

The comment at line 1159 says "No design surface owns a mutation transport at all," but the check only proves those 11 literals are absent from 4 files. It is blind to transitive reach: `app/builderops/design_agent_adapters.py` imports `BuilderModelAccessResolver`, whose `credential_value()` calls `load_runtime_secret_values` — the exact token the list is watching for. The *conclusion* is still correct for these paths (no exercised case reaches `credential_value`, which I verified), but the assertion does not prove it.

Brittleness: `subprocess` appears verbatim in INV-CDH-5A's own text ("HTTP/subprocess transport"). The day someone quotes that invariant in a docstring in one of these four modules, this passes-today assertion fails for a purely cosmetic reason.

The substantive proofs of AC2 — the repo-tree digest at line 1148, `object_types == {"BuilderOpsReceipt"}`, and the `authority_surface == "builderops"` target-ref check — are the ones doing real work and I mutation-confirmed they bite. Consider softening the comment to what the grep actually asserts.

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+        expected_approval_id=mismatch_projection.approval.approval_id,
+        expected_approval_hash=mismatch_projection.approval.content_hash,
+    )
+    assert governance.projection("run.acceptance.mismatch").state == "succeeded"
+    assert _total_calls(adapters) == 1
+
+    # -- Case: one governed success per headless-registered adapter ----------
+    for index, agent_id in enumerate(HEADLESS_CAPABLE_AGENT_IDS):
+        success_adapters = _fresh_adapters()
+        success_governance = _substrate_governance(
```
RasmusTho, line 918: **P3 — the governed-success half of AC1 does not run through the production composition root.**

Every success/timeout/failure/malformed-output/foreign-binding case is built with the direct `DesignRunGovernance(...)` constructor via `_substrate_governance` and `channel="test"`, not `DesignRunGovernance.from_declared_sources(...)` with `channel="dev"`. Only the dormant read path (`_assert_cli_composition_is_fail_closed`) reaches the real Click composition root.

I checked this is largely unavoidable: `from_declared_sources` constructs `BuilderModelAccessResolver` internally with no injection seam for the ADR-0064 port, so proving a governed success without provisioning a provider requires the direct constructor. Doubling the `ModelAccessResolver` port and the `ModelTurnAdapter` behind it is honest — the real `DesignAgentAdapterRegistry`, real `DesignRunGovernance`, shipped policy/token bytes, real SQLite store and real renderer are all live, and my probes A–E confirm the governance seam under proof is not stubbed away.

Worth recording that the residue is: for the success half, the composition-root wiring (which resolver, which channel) is asserted by nothing. AC1's phrase "production composition" is therefore slightly stronger than what runs.

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+    assert foreign.result.handoff is None
+    assert foreign.refusal is not None
+    assert foreign.refusal.code == "execution_failed"
+    assert _total_calls(foreign_adapters) == 1
+
+    # -- The real CLI composition root renders the same fail-closed truth ----
+    _assert_cli_composition_is_fail_closed(tmp_path, monkeypatch)
+
+    # -- Distinctness of every terminal/refusal state ------------------------
+    assert set(observed) == {
```
RasmusTho, line 1027: **P3 — two of these three assertions are vacuous.**

Line 1036's `assert observed == {literal dict}` fully determines both `set(observed)` (line 1027) and `len(set(observed.values())) == len(observed)` (line 1045). Once the dict equality holds, neither of the other two can ever fail. They read as three independent checks but are one.

Cosmetic only — the dict equality itself is meaningful (`observed` is populated from real `governance.projection(...).state` reads) and I confirmed it is not vacuous. Just noise in a module whose whole job is to make assertions legible as proof.

RasmusTho (commented):
## Independent final review gate — verdict: **CLEAN (non-blocking)**

Fresh reviewer, no knowledge of how this PR was written. Scope: the PR diff only, reviewed against issue #4313, `docs/CKM_DESIGN_AGENT_INTEGRATION/VALIDATE_DESIGN_HUB_END_TO_END.md`, `README.md :: Cross-Task Invariants / Interaction Safety` and `:: Capability acceptance`, and `AGENTS.md :: Proportional delivery`.

**No P0 or P1 findings. 3 x P2 (real defects, non-blocking, follow-up), 5 x P3.**

### Gate commands at head `6a3736d61`

| Command | Result |
|---|---|
| `python3 -m pytest -q tests/builderops/test_design_hub_acceptance.py tests/builderops/ckm/test_design_cockpit.py` | **9 passed** in 0.95s |
| `python3 -m ruff check app tests companion-ui/companion-app` | **All checks passed** |
| `python3 -m mypy app` | **Success, 848 source files** |
| `python3 scripts/docs_guard.py` | **OK** |

### Independent proof-of-proof (my own mutation probes, not the author's)

Ran 5 fresh probes against unchanged production modules and reverted each; **5 of 5 detected** by the tests in this diff:

| Probe | Mutation | Detected |
|---|---|---|
| A | `design_run_governance.py`: handoff `source_refs` no longer bound to the brief | yes |
| B | `ckm/design_hub.py`: `previous_receipt_id` dropped from the projected chain | yes |
| C | `design_agent_adapters.py`: provider identity leaked on an unavailable descriptor (INV-CDH-7) | yes |
| D | `design_run_contract.py` + governance: handoff relabelled `accepted_truth` (INV-CDH-8) | yes |
| E | `ckm/design_hub.py`: refusal `public_message` dropped | yes |
| H | `ckm/overview_html.py`: `details > :not(summary) { display:block !important; }` removed from `@media print` | yes |
| I | `ckm/design_hub.py`: adapter ordering unsorted | yes |

One probe **survived** — see P2-2.

### Factual spot-checks of the PR body — all accurate

- Yggdrasil token `companion-ui/companion-app/colors_and_type.css` SHA-256 = `7d8cdd49f59061f895959159a08e82348e7e02eb8b8ba7426020a50c7fa915b1` — **confirmed** by independent `shasum -a 256`.
- Production registry posture — **confirmed** by running `DesignAgentAdapterRegistry.from_declared_sources(channel="dev")` outside the tests: all three agents `available=False`, `provider_identity=None`, limitation codes exactly `interactive_subscription_only` / `model_access_unavailable` / `model_access_unavailable`. No `builderops-design-run` grant exists in the repo and none was created by this diff.
- "`scripts/select_pr_tests.py` expands this diff to the `builder_system` lane" — **confirmed**; running the selector on the four changed paths yields `subsystems=builder_system` and `pytest_args=... tests/builderops ...`, so the whole tree does run at the head SHA.
- "CDH-01 through CDH-05 carry terminal delivery receipts on #4131" — **confirmed**; #4308–#4312 are all CLOSED and each has a child validation receipt comment on #4131.
- No credential is read on any exercised path: `load_runtime_secret_values` is reached only from `BuilderModelAccessResolver.credential_value`, which no case in this matrix invokes.

### Authority-boundary check — clean

The diff is 2 test files + 2 spec `State:` lines. Parent #4131 is not closed, relabelled, or mutated. `README.md :: Capability acceptance` remains fully unticked. No CKM or Builder System owner doc is promoted, no provider provisioned, no GitHub mutation. Both new `State:` lines are factually accurate against live GitHub.

### Proportionality

A 1281-line acceptance module is proportional here: issue #4313's Scope enumerates ~20 required cases and the module covers essentially all of them. I found no simpler mechanism that satisfies the contract, and no new permanent governance gate (the round-1 `FOCUSED_REGRESSION_HEAD_DIGEST` removal was the right call). Not over-built.

Findings follow as inline comments.

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+                str(repo_root),
+                "--out",
+                str(out_path),
+            ],
+        )
+        assert result.exit_code == 0, result.output
+        outputs.append(out_path.read_text(encoding="utf-8"))
+
+    # Byte-identical final cockpit HTML from fixed inputs (INV-CDH-9).
+    assert outputs[0] == outputs[1]
```
RasmusTho, line 526: **P3 — the PR body's deterministic print receipt is not independently reproducible.**

The body cites SHA-256 `cab8349df2c11eda0e2372fc441d3132062f0eb84a4fe7683e92bb6c806f78dc` at 27,810 bytes for the final cockpit HTML "generated three times from fixed inputs through the real CLI." I reconstructed that invocation from this helper and got a different digest and 27,806 bytes — because the CKM store's `epoch_<uuid>` and the sandbox path are per-run values that the body does not pin. The digest is therefore a narrative claim about one machine, not a receipt a third party can re-derive from the repo.

Issue #4313's Scope offers "a reviewed print artifact **or** deterministic print receipt" as the evidence, so this matters for what the parent acceptance audit can independently confirm. The durable, reproducible part is what the test asserts — `outputs[0] == outputs[1]`, the structural `@media print` rules, and the JS-off completeness checks — and those I did verify and mutation-confirm (probe H). Recommend the ledger on #4131 cite the test rather than the bare digest, or state the exact pinned inputs.

RasmusTho (commented):
## Final review gate — re-run at head `2f92cc4c935b39d9ba31f67dcbe5de16bfdfd12e`

**Verdict: CLEAN** (1 P3 new, 2 P3 informational). Nothing blocking.

Fresh independent review, scoped primarily to the delta `6a3736d61..2f92cc4c9` (32 lines across two test modules), with the prior round's cleared categories re-confirmed at the current head.

---

### P2-2 — deterministic ordering mutant — **FIXED, mutation-verified**

Reproduced the mutant directly: removed both `sorted()` calls in `app/builderops/ckm/design_hub.py` (`known_run_ids`, and `runs=tuple(sorted(runs, key=...))`) and ran `test_design_hub_projection_preserves_direction_b_authority` under `PYTHONHASHSEED` 0–5.

| Seed | 0 | 1 | 2 | 3 | 4 | 5 |
|---|---|---|---|---|---|---|
| Result | FAIL | FAIL | FAIL | FAIL | FAIL | FAIL |

Every failure lands on the new assertion, not incidentally:

```
>       assert [run.run_id for run in design_hub.runs] == sorted(states)
E       AssertionError: At index 0 diff: 'run.cockpit.failed' != 'run.cockpit.denied'
```

**Control run**: with the same mutant applied, I checked out the *previous* round's `tests/builderops/ckm/test_design_cockpit.py` (at `6a3736d61`) and it **passed** on seeds 0–3. So the prior finding was real and the delta genuinely closes it.

**Not tautological.** `states` is a dict built from `design_hub.runs`, but it is independently pinned to a 7-key literal two lines above, and `sorted(states)` normalizes key order — so the comparison is projection-order vs. a genuinely independent total order. Confirmed by the mutant killing it. Production file reverted.

### P2-1 — matrix partition staleness — **FIXED, verified against three drift scenarios**

Mutated `app/builderops/design_agent_adapters.py` three ways; all three are caught:

| Scenario | Caught by | Line |
|---|---|---|
| 4th agent `zephyr` with a **new** role profile `design.zephyr` | exhaustiveness union | 1264 |
| 4th agent `zephyr` **reusing** an existing role profile `design.fable` | exhaustiveness union | 1264 |
| Existing agent's role profile renamed (`design.fable` → `design.fable.v2`) | `set(SUBSTRATE_TARGETS) == set(DESIGN_AGENT_ROLE_PROFILES.values())` | 1268 |

The reused-profile case is the one that matters, and it is caught: `SUBSTRATE_TARGETS` parity alone would not have noticed it, but the `HEADLESS ∪ INTERACTIVE == DESIGN_AGENT_IDS` assertion does. The two partitions are now bound to `DESIGN_AGENT_IDS`, not hand-maintained in isolation. Production file reverted.

### P2-3 — evidence scope — **FIXED, prefix matching correct and non-vacuous**

- Discovery is **not** empty: 13 modules discovered, 12 matched exactly, 1 (`tests/builderops/ckm/test_design_cockpit.py`) matched via the directory prefix. No `UNMATCHED`.
- `tests/builderops/ckm` genuinely contains 19 test modules, so the comment's claim is accurate and the entry **broadens** the evidence rather than narrowing it.
- **No swallow-everything.** There is no bare `tests/builderops` entry, and one could not be introduced silently: I substituted `tests/builderops` for `tests/builderops/ckm` and the test **fails** at line 1296 on `all(relative.startswith("tests/builderops/") ...)` — `"tests/builderops"` does not start with `"tests/builderops/"`. The structural guard holds.
- **No spurious prefix matches.** `f"{candidate}/"` for the file candidates (e.g. `tests/builderops/test_design_run_cli.py/`) matches nothing, so file entries are only satisfied by exact membership.
- **Escape check**: added a stray `tests/builderops/zprobe/test_design_stray.py` — correctly rejected with `... is not covered by the named focused-regression set`.
- The `.exists()` / `.is_file()` split is correct: suites may now legitimately be directories; surfaces are all still files.

---

### New finding

**P3 — two of the three new determinism assertions are vacuous at the fixture's cardinality.**
`tests/builderops/ckm/test_design_cockpit.py:620-624`

`_RecordingRegistry.descriptors()` returns exactly **one** descriptor, and exactly **one** run is excluded. Both of these are therefore comparisons of a length-1 sequence against its own sorted form, which can never fail:

```python
assert list(design_hub.excluded_run_ids) == sorted(design_hub.excluded_run_ids)
assert [item.design_agent_id for item in design_hub.adapters] == sorted(
    item.design_agent_id for item in design_hub.adapters
)
```

Verified empirically: replacing `excluded_run_ids=tuple(sorted(excluded))` with `tuple(reversed(excluded))` in `design_hub.py` leaves **all 330 tests** in `tests/builderops/test_design_hub_acceptance.py tests/builderops/ckm` green. The `excluded_run_ids` line is also redundant with the exact-tuple pin two lines above it (`== ("run.cockpit.excluded",)`).

Mitigations that keep this at P3 rather than P2:
- The adapters ordering *is* enforced elsewhere — replacing the adapters `sorted()` with `reversed()` is caught by the pre-existing `test_selector_uses_registered_adapter_availability` (`test_design_cockpit.py:384`).
- Only `excluded_run_ids` ordering is genuinely unenforced anywhere, and no fixture in the tree produces more than one excluded run, so this is a latent gap rather than an escaped defect.
- AC3 / INV-CDH-9's substantive claim — deterministic run ordering — is now enforced and mutation-verified. This is not a failed acceptance criterion.

Suggested (deferrable): either drop the two decorative lines, or give the fixture a second excluded run and a second adapter so they carry weight.

### Informational

**P3 — the PR body's evidence tables still cite the previous head `6a3736d61`.** Both the "Acceptance criteria — `Verify:` results at head" table and the "Validation actually run at head `6a3736d61`" table predate this delta. The claims are truthful about the head they name, and I independently reproduced every one of them at `2f92cc4c9` (below), so this is a stale label, not false evidence. Worth refreshing before merge.

**P3 — `tests/builderops/ckm/test_overview_html.py` lost its individual existence check.** It was previously `.is_file()`-verified as a named suite and force-included in `discovered`; it is now covered only by the `tests/builderops/ckm` directory entry, and its filename matches neither the `design` nor `model_inquiry` discovery filter. Low impact: it is not an AC `Verify:` target, and CI runs the whole directory. Note that `test_design_cockpit.py` is *not* exposed this way — `ACCEPTANCE_VERIFY_TARGETS` reads it by path, so AC3's target stays bound.

---

### Regression clearance re-confirmed at this head

| Check | Result |
|---|---|
| No `app/` production change in the PR | Confirmed — `gh pr diff` touches exactly 4 files: 2 spec `State:` lines + 2 test modules |
| Parent #4131 not closed or mutated | Confirmed — issue is `OPEN`, untouched by the diff |
| No CKM / Builder System owner doc promoted | Confirmed — only the two local spec `State:` lines change |
| `README.md :: Capability acceptance` fully unticked | Confirmed — all 5 boxes `- [ ]` |
| No provider provisioned, no credential read | Confirmed — delta contains zero matches for secret/credential/token/api-key/grant/provision |

### Gate results at `2f92cc4c9`

| Gate | Result |
|---|---|
| `python3 -m pytest -q tests/builderops/test_design_hub_acceptance.py tests/builderops/ckm/test_design_cockpit.py` | **9 passed** |
| `python3 -m pytest -q tests/builderops` (the CI lane the head-binding claim relies on) | **1087 passed** in 175.8s |
| `python3 -m ruff check app tests companion-ui/companion-app` | All checks passed |
| `python3 -m mypy app` | Success, 848 source files |
| `python3 scripts/docs_guard.py` | OK |

All mutation probes were reverted; `git status --short` is empty.

tests/builderops/test_design_hub_acceptance.py
```diff
@@ -0,0 +1,1281 @@
...
+        expected_admission_id=mismatch_projection.admission.admission_id,
+        expected_admission_hash=mismatch_projection.admission.content_hash,
+        expected_approval_id=mismatch_projection.approval.approval_id,
+        expected_approval_hash=mismatch_projection.approval.content_hash,
+    )
+    assert governance.projection("run.acceptance.mismatch").state == "succeeded"
+    assert _total_calls(adapters) == 1
+
+    # -- Case: one governed success per headless-registered adapter ----------
+    for index, agent_id in enumerate(HEADLESS_CAPABLE_AGENT_IDS):
```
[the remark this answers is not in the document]
RasmusTho: Disposition — terminal parent-acceptance audit of #4131 (audit 6).

**Not a defect in the test; the acceptance contract it cited was the defect.** Requiring "one governed success per registered adapter" over all three registered agents was structurally unsatisfiable, because `claude-design-via-claude-code` is hard-refused as `interactive_subscription_only` inside `DesignAgentAdapterRegistry` before any adapter lookup (INV-CDH-5A). The criterion was repaired by **PR #4503** (`b7bc0f966b6599757b477d6fe1634c736763c55f`) to require one governed success per *headless-registered* adapter plus an exact zero-provider-call refusal for every interactive-only adapter.

Re-verified independently on merged `origin/main` `74a7460afece70fd032858a91acc5ea27bd9c6fa`: the interactive route is proven at `test_design_hub_acceptance.py:660` while `codex` and `fable` are simultaneously available in the same governance, and the two buckets are asserted exhaustive and disjoint over `DESIGN_AGENT_IDS` at `:1264-1267`. Every registered route still carries a proof obligation — success or exact refusal.

Resolved as part of parent #4131 closure.

tests/builderops/test_design_hub_acceptance.py
RasmusTho: Disposition — terminal parent-acceptance audit of #4131 (audit 6).

**Fixed in this PR, not deferred.** The PR body records this as one of three final-gate P2s fixed at `2f92cc4c9` rather than routed to the deferred registry.

Re-verified independently on merged `origin/main` `74a7460afece70fd032858a91acc5ea27bd9c6fa`: `test_design_hub_acceptance.py:1264-1267` asserts `set(HEADLESS_CAPABLE_AGENT_IDS) | set(INTERACTIVE_ONLY_AGENT_IDS) == set(DESIGN_AGENT_IDS)` and that the intersection is empty, inside a test that runs in the required `builder_system` lane. A fourth registered adapter cannot land in neither bucket, so the matrix can no longer silently shrink.

Resolved as part of parent #4131 closure.

tests/builderops/ckm/test_design_cockpit.py
RasmusTho: Disposition — terminal parent-acceptance audit of #4131 (audit 6).

**Fixed in this PR, not deferred.** Recorded in the PR body as the most important of the three final-gate P2 fixes, mutation-verified at `2f92cc4c9` under `PYTHONHASHSEED` 0-5.

Re-verified independently on merged `origin/main` `74a7460afece70fd032858a91acc5ea27bd9c6fa`: `app/builderops/ckm/design_hub.py` now pins total order at three mutually redundant points — `known_run_ids = sorted(...)` (`:188`), `runs=tuple(sorted(runs, key=...))` (`:238`), and `excluded_run_ids=tuple(sorted(excluded))` (`:239`). The independent mutation probe recorded in the [conditional acceptance receipt](https://github.com/RasmusTho/agentic-pkm-mvp/issues/4131#issuecomment-5149608088) §5 reproduces detection when the redundant layers are removed together.

Resolved as part of parent #4131 closure.

tests/builderops/test_design_hub_acceptance.py
RasmusTho: Disposition — terminal parent-acceptance audit of #4131 (audit 6).

**Fixed in this PR, not deferred.** Third of the three final-gate P2s fixed at `2f92cc4c9`.

Re-verified independently on merged `origin/main` `74a7460afece70fd032858a91acc5ea27bd9c6fa`: `FOCUSED_REGRESSION_SUITES` (`test_design_hub_acceptance.py:1204-1218`) now names the whole `tests/builderops/ckm` directory rather than two of its nineteen modules, which is exactly the surface AC4 and #4313 `Suggested Validation` demand. The full `tests/builderops` tree (1087 tests) passes on this SHA.

The separate observation about the AC4 `Verify:` target *name* is P3 and owned by **#4502** (open, `agent:ready`).

Resolved as part of parent #4131 closure.

PR #4572, opened 2026-08-02
fix: refuse divergent Keychain material across shared-domain host secrets

RasmusTho:
Governing-Issue: #4512

Fixes #4512

Final-Review-Rounds: 2

## Summary
`heimdal.raw-store-key` is declared for two independently bootstrapped consumers (`heimdal-capture-watch`, `heimdal-api-ingress`) that feed one AES-256-GCM cipher domain, but nothing previously verified their Keychain items held the same value. The host secret contract now declares which logical secrets are shared-domain (`heimdal.raw-store-key` marked, model-provider keys not), and the bootstrap entrypoint (`run_with_host_secrets` / `main`) cross-checks a non-reversible digest of the resolving consumer's value against every other declared consumer's value on the same channel, refusing with a value-free named error (`HostSecretSharedDomainDivergenceError`) rather than proceeding into a split cipher domain. A sibling whose own item is unresolvable (the typed `HostSecretBootstrapError` every in-repo `keychain_lookup` raises) is skipped, not treated as divergence; any other error still fails closed. Raw-store-key material is compared as decoded bytes so two hex-case spellings of the same key are not falsely flagged as diverged. Matching material bootstraps byte-for-byte unchanged; the account template, request-time api preflight posture, and non-shared secrets are untouched.

## BuilderOps Routing
- Reason: Tier 2 single-issue slice; no BuilderOps record/projection surface was in scope.

## Notes
Validation:
- `ruff check app tests` — clean
- `python3 -m mypy app/ops/host_secret_bootstrap.py app/ops/host_secret_contract.py` — clean
- `python3 scripts/public_seam_lint.py --mode gate` — `secret_hits: 0`
- `python3 scripts/docs_guard.py` — OK
- `pytest -q tests/ops/test_host_secret_contract.py tests/ops/test_host_secret_bootstrap.py tests/builderops tests/deploy/test_deploy_channel_script.py` — 1206 passed
- `pytest -q tests/heimdal/test_media_ingress.py` — 1 unrelated pre-existing failure (`test_preflight_logs_the_remedy_matching_the_failing_precondition`, needs `STORE_BACKEND`/`DATABASE_URL` not configured in this sandbox; reproduces identically on `origin/main` with these changes stashed)

Security-sensitive surface (Keychain/credential-durability), declared high-risk runtime TCD surface — two consecutive clean independent review rounds required:
- Ran the review-before-CI gate (`--risk-surface security --risk-surface credential-durability`) before pushing.
- **Round 1** (fresh independent reviewer, no implementation context): verdict clean (no P0/P1). Two P2s found: (a) raw-store-key digest compared raw text instead of decoded bytes, so two hex-case spellings of the same key would false-positive as diverged; (b) the sibling-lookup skip caught bare `Exception`, which could silently swallow an unexpected programming error instead of failing closed. Both were cheap, well-understood fixes, so they were applied in this PR rather than deferred to the Known Defects registry (reviewer's own recommendation for finding (a)): normalized raw-store-key material to decoded bytes before the digest, narrowed the skip to the typed `HostSecretBootstrapError`, and added four regression tests (hex-case equivalence, programming-error-still-fails-closed, typed-failure-still-skips, plus the pre-existing three). Full suite re-run after the fix: 1206 passed.
- **Round 2**: pending on the post-fix head SHA before merge, per the two-consecutive-clean-round requirement for this declared high-risk surface.

Deferred (non-blocking, informational P3s from round 1, no code change needed): `HostSecretContract.shared_key_domain_secrets` defaults to an empty frozenset on a directly-constructed contract (matches the pre-existing `optional_secrets` pattern; the production loader always requires the field explicitly); the shared-domain declaration's cross-consumer Keychain read is documented only in code comments and the operator runbook, not in the contract schema itself; one test patches `subprocess.run` at module scope for the duration of that test (restored by `monkeypatch`, blast radius wider than the assertion needs but sound).
---

### DIFF
--- a/app/ops/host_secret_bootstrap.py
+++ b/app/ops/host_secret_bootstrap.py
@@ -322,13 +322,35 @@ def _secret_failure(*, secret: str, kind: str) -> HostSecretBootstrapError:
     )
 
 
-def _shared_domain_digest(value: str) -> bytes:
+def _shared_domain_comparable_bytes(kind: str, value: str) -> bytes:
+    """Normalize *value* to the bytes the cipher domain actually uses.
+
+    `raw-store-key` values are hex text that the raw store decodes with
+    `bytes.fromhex` before use as key material (`app/heimdal/raw_store.py`), so
+    two hex strings that differ only in case are the *same* key. Comparing the
+    raw text would flag that pair as diverged even though the domain agrees;
+    decode first so the equivalence check matches the cipher's own reading of
+    the value. A value that fails to decode is left as UTF-8 text — it is
+    already malformed for its declared kind and will fail
+    ``_validate_secret`` on its own path; this function only has to make sure
+    it does not COMPARE EQUAL to a well-formed sibling by accident, and text
+    bytes never collide with decoded key bytes here.
+    """
+    if kind == "raw-store-key":
+        try:
+            return bytes.fromhex(value)
+        except ValueError:
+            return value.encode("utf-8")
+    return value.encode("utf-8")
+
+
+def _shared_domain_digest(kind: str, value: str) -> bytes:
     """Non-reversible fingerprint used to compare shared-domain material.
 
     Never return, log, or persist this alongside the value it was derived
     from; it exists only to be compared against another digest in-process.
     """
-    return hashlib.sha256(value.encode("utf-8")).digest()
+    return hashlib.sha256(_shared_domain_comparable_bytes(kind, value)).digest()
 
 
 def _check_shared_domain_agreement(
@@ -336,6 +358,7 @@ def _check_shared_domain_agreement(
     channel: str,
     consumer: str,
     secret: str,
+    kind: str,
     value: str,
     contract: HostSecretContract,
     keychain_lookup: KeychainLookup,
@@ -343,15 +366,18 @@ def _check_shared_domain_agreement(
     """Refuse rather than proceed when a sibling consumer's material differs.
 
     Only secrets declared `shared_key_domain` are checked. A sibling consumer
-    whose own material cannot be resolved (not yet provisioned, transient
-    lookup failure) is not itself evidence of divergence — that failure
-    belongs to the sibling's own bootstrap invocation — so it is skipped here
-    rather than failing this consumer closed. Only two resolvable values that
-    disagree count as divergence.
+    whose own material cannot be resolved (not yet provisioned, a lookup
+    failure the sibling's own bootstrap will surface on its own path) is not
+    itself evidence of divergence, so it is skipped here rather than failing
+    this consumer closed — narrowly, on the same typed failure every
+    `keychain_lookup` implementation in this module raises, so an unexpected
+    programming error (wrong callable signature, a decode bug) still fails
+    this consumer closed instead of silently disabling the check. Only two
+    resolvable values that disagree count as divergence.
     """
     if not contract.is_shared_key_domain(secret):
         return
-    own_digest = _shared_domain_digest(value)
+    own_digest = _shared_domain_digest(kind, value)
     for sibling_consumer in sorted(
         contract.consumers_declared_for(channel=channel, secret=secret)
     ):
@@ -362,9 +388,9 @@ def _check_shared_domain_agreement(
         )
         try:
             sibling_value = keychain_lookup(contract.keychain_service, sibling_account)
-        except Exception:
+        except HostSecretBootstrapError:
             continue
-        if not hmac.compare_digest(_shared_domain_digest(sibling_value), own_digest):
+        if not hmac.compare_digest(_shared_domain_digest(kind, sibling_value), own_digest):
             raise HostSecretSharedDomainDivergenceError(secret)
 
 
@@ -413,6 +439,7 @@ def _resolve_consumer_environment(
                 channel=channel,
                 consumer=consumer,
                 secret=secret,
+                kind=kind,
                 value=value,
                 contract=contract,
                 keychain_lookup=keychain_lookup,
--- a/app/ops/host_secret_bootstrap.py
+++ b/app/ops/host_secret_bootstrap.py
@@ -6,6 +6,8 @@ import argparse
 from collections.abc import Callable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
 import ctypes
+import hashlib
+import hmac
 import os
 from pathlib import Path
 import re
@@ -52,6 +54,22 @@ class HostSecretBootstrapTerminated(HostSecretBootstrapError):
     def __init__(self, signum: int) -> None:
         super().__init__("host secret bootstrap terminated for declared consumer")
         self.signum = signum
+
+
+class HostSecretSharedDomainDivergenceError(HostSecretBootstrapError):
+    """Two declared consumers of a shared-domain secret resolved to different material.
+
+    Raised without either value, or anything derived from which a value could
+    be recovered — only the declared logical secret identifier, which is what
+    an operator needs to locate and re-provision the diverged Keychain items.
+    """
+
+    def __init__(self, shared_domain_secret_ref: str) -> None:
+        super().__init__(
+            "host secret bootstrap failed for declared consumer: shared-domain "
+            f"secret diverges across consumers: {shared_domain_secret_ref}"
+        )
+        self.shared_domain_secret_ref = shared_domain_secret_ref
 
 
 SignalHandler = Callable[[int, FrameType | None], None]
@@ -304,6 +322,52 @@ def _secret_failure(*, secret: str, kind: str) -> HostSecretBootstrapError:
     )
 
 
+def _shared_domain_digest(value: str) -> bytes:
+    """Non-reversible fingerprint used to compare shared-domain material.
+
+    Never return, log, or persist this alongside the value it was derived
+    from; it exists only to be compared against another digest in-process.
+    """
+    return hashlib.sha256(value.encode("utf-8")).digest()
+
+
+def _check_shared_domain_agreement(
+    *,
+    channel: str,
+    consumer: str,
+    secret: str,
+    value: str,
+    contract: HostSecretContract,
+    keychain_lookup: KeychainLookup,
+) -> None:
+    """Refuse rather than proceed when a sibling consumer's material differs.
+
+    Only secrets declared `shared_key_domain` are checked. A sibling consumer
+    whose own material cannot be resolved (not yet provisioned, transient
+    lookup failure) is not itself evidence of divergence — that failure
+    belongs to the sibling's own bootstrap invocation — so it is skipped here
+    rather than failing this consumer closed. Only two resolvable values that
+    disagree count as divergence.
+    """
+    if not contract.is_shared_key_domain(secret):
+        return
+    own_digest = _shared_domain_digest(value)
+    for sibling_consumer in sorted(
+        contract.consumers_declared_for(channel=channel, secret=secret)
+    ):
+        if sibling_consumer == consumer:
+            continue
+        sibling_account = contract.keychain_account(
+            channel=channel, consumer=sibling_consumer, secret=secret
+        )
+        try:
+            sibling_value = keychain_lookup(contract.keychain_service, sibling_account)
+        except Exception:
+            continue
+        if not hmac.compare_digest(_shared_domain_digest(sibling_value), own_digest):
+            raise HostSecretSharedDomainDivergenceError(secret)
+
+
 def _resolve_consumer_environment(
     *,
     channel: str,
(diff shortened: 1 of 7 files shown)


RasmusTho (commented):
## Independent review — round 1 of 2 (high-risk surface: Keychain host-secret bootstrap)

Reviewed cold against #4512. Verdict: **clean** — no P0/P1. Two P2s and three P3s below.

### What holds up

Verified in the local checkout at the head SHA, not just read:

- **No material or digest escapes.** `HostSecretSharedDomainDivergenceError` carries only the logical id; no `from exc` chaining of a value-bearing error; `_shared_domain_digest` output is never returned, logged, or persisted. `main` prints `str(exc)` only.
- **Fail-closed posture preserved and not bypassable.** The new error is a `HostSecretBootstrapError` but *not* a `HostSecretCredentialUnavailableError`, so the `--run-on-credential-unavailable` handoff (`host_secret_bootstrap.py:625-641`) cannot run a child on a diverged domain. Confirmed by reading the handler, not assumed.
- **Refusal happens before materialization.** Divergence raises inside `_resolve_consumer_environment`, i.e. before `materialize_consumer_environment` creates any temp file and before `runner` is invoked. `test_divergent_shared_domain_secret_fails_loud` asserts both (`not launched`, `tmp_path` empty).
- **No behavior change for non-shared secrets.** Gated on `contract.is_shared_key_domain`, not on consumer count — and `openai.api-key` (two declared consumers, not shared-domain) is proven to issue no sibling lookup.
- **Sibling account derivation is guarded.** `contract.keychain_account` calls `require_declared` (`host_secret_contract.py:84`), so the sibling tuple must itself be contract-declared.
- **Closed schema enforced.** `set(item) != _SECRET_FIELDS` rejects an omitted field, and `type(x) is not bool` rejects `1`/`0`/`"true"`/`None` — parametrised. Every in-repo producer of a contract payload was updated (grep for `child_binding` across `.py`/`.json`: only `config/secrets/host_secret_contract.json`, `tests/builderops/inquiry_intent.py`, `tests/ops/test_host_secret_contract.py`, all covered).
- `hmac.compare_digest` for the comparison; `pytest -q tests/ops/test_host_secret_contract.py tests/ops/test_host_secret_bootstrap.py` → 75 passed locally.

---

### P2 — `app/ops/host_secret_bootstrap.py:331,367` — two hex-case variants of the *same* AES key are refused as divergent

`_validate_secret` accepts `[0-9a-fA-F]{64}` (`:28`) and `resolve_raw_store_key` decodes with `bytes.fromhex` (`app/heimdal/raw_store.py:129`), so `"a"*64` and `"A"*64` are byte-identical key material feeding one cipher domain. The digest is taken over the raw string, so they compare unequal. Reproduced against this branch:

```
lookup: capture-watch -> "a"*64, api-ingress -> "A"*64   (both pass _validate_secret)
=> REFUSED: HostSecretSharedDomainDivergenceError ... heimdal.raw-store-key
   bytes.fromhex("a"*64) == bytes.fromhex("A"*64)  ->  True
```

Impact is not theoretical: on such a host *both* consumers now permanently refuse to bootstrap, and the governed deploy path (`scripts/lib/deploy_channel_compose.sh:380-412`, which nests both bootstraps around `docker compose`) fails the entire compose invocation with exit 78 — on a correctly-provisioned, non-diverged domain. The PR body defers this as "fail-safe direction, not a security gap". The direction is fail-safe, agreed, and that is why this is P2 and not blocking — but a false positive that takes down a working deployment is still a defect, and the remedy is one token (`_shared_domain_digest(value.strip().lower())`, or normalising per declared `kind`) plus a test. Cheaper to fix here than to carry a follow-up issue. Recommend fixing in-PR; if deferred, it needs a logged Issue, not just a PR-body note.

### P2 — `app/ops/host_secret_bootstrap.py:363-366` — the sibling-skip branch is over-broad *and* has zero test coverage

```python
try:
    sibling_value = keychain_lookup(contract.keychain_service, sibling_account)
except Exception:
    continue
```

Two problems, one line:

1. **Over-broad.** The docstring justifies the skip for "not yet provisioned, transient lookup failure", but the handler swallows *everything* — a `TypeError` from a changed lookup signature, a `UnicodeDecodeError` out of `_security_framework_keychain_lookup`, a locked-keychain error, any future refactor bug. Each of those silently disables the divergence check for that sibling with no signal and a zero exit. A fail-closed control whose disable path is this wide is a control that can stop working without anyone noticing. Suggest narrowing to `except HostSecretBootstrapError` — that is exactly the failure type both production lookups raise for an unresolvable item, and anything else then reaches `_resolve_consumer_environment`'s outer handler and fails closed, which is the correct default for an unexpected error inside a security check.
2. **Untested.** The four new tests cover divergent, matching, non-shared-gating, and no-leak. None exercises the skip branch. I verified it works by hand (sibling lookup raising → `run_with_host_secrets` returns 0), but the documented behaviour of a security-relevant branch on a TCD high-risk surface should be pinned by a regression test, or a later narrowing/widening of that `except` will go unnoticed.

The PR body logs the "no operator-visible signal on skip" half as deferred; the over-broad-except and the coverage gap are not logged. Not blocking: a genuinely unresolvable sibling item also fails that sibling's own required-secret bootstrap, so no split domain materialises at runtime.

### P3 — `app/ops/host_secret_contract.py:76` — the new field is default-open on the dataclass

`shared_key_domain_secrets: frozenset[str] = frozenset()` means any directly-constructed `HostSecretContract` silently has the check switched off, while `load_host_secret_contract` correctly demands the field explicitly. Only test code constructs directly today (`tests/ops/test_host_secret_bootstrap.py:264`, `tests/ops/test_host_secret_contract.py:66`), so there is no production exposure — but it is a security control whose safe state is the non-default one. Same shape as the pre-existing `optional_secrets`, so this is consistency-preserving rather than a regression; noting it so the pattern is a conscious choice.

### P3 — sibling reads widen the bootstrap's effective Keychain authority, asserted only in a comment

`heimdal-capture-watch`'s bootstrap now reads `dev:heimdal-api-ingress:heimdal.raw-store-key`. It is guarded and the value never leaves `_check_shared_domain_agreement`, but the contract has no representation of "a consumer may read a sibling's account for this secret" — it is stated only in a docstring. The operator guidance added to `DEFINE_HOST_SECRET_CONTRACT.md` is the natural place to state that a shared-domain declaration also grants cross-consumer read at bootstrap time.

### P3 — `tests/ops/test_host_secret_bootstrap.py:1171` patches the real `subprocess` module

`monkeypatch.setattr(host_secret_bootstrap.subprocess, "run", fake_run)` targets the `subprocess` module object itself, so `run` is patched process-wide for the duration. It is restored by `monkeypatch` and `_subprocess_runner` uses `Popen`, so it is sound today; the comment explains why the module attribute cannot be patched instead. Flagging only because the blast radius is wider than the assertion needs.

---

**Round 1 verdict: clean** (no P0/P1). Two P2s to fix-or-log before the round-2 pass; the hex-normalisation one is a one-token fix I'd rather see in this PR than in a follow-up.

RasmusTho (commented):
## Independent review — round 2 of 2 (cold, no implementation context)

Head SHA reviewed: `567f8748c4343c44a5524815133d0f24f4610fb4`
Scope: full PR diff (7 files), plus surrounding context in `app/ops/host_secret_bootstrap.py`, `app/ops/host_secret_contract.py`, `app/heimdal/raw_store.py`, `app/builderops/model_access_resolver.py`.

### What I independently verified

**Contract surface (`app/ops/host_secret_contract.py`)**
- `shared_key_domain` is a closed, mandatory field: `set(item) != _SECRET_FIELDS` rejects both omission and extras, and `type(shared_key_domain) is not bool` blocks the `1`/`"true"` smuggling path that `optional` already guards. Confirmed by the new parametrized loader tests (`omitted` / `truthy-int` / `falsy-int` / `string` / `null`).
- Because the field is now mandatory, I swept for any other contract payload that would break: `grep -rl logical_id --include=*.json` returns only `config/secrets/host_secret_contract.json`, and the only in-repo synthetic secret-declaration builder is `tests/builderops/inquiry_intent.py:147`, which the diff updates. No docs carry a stale schema example (`docs/LOCAL_SECRET_PROVISIONING/`, `docs/MODEL_ACCESS_SUBSTRATE/` checked). No shell/jq consumer parses the `secrets` array (`scripts/lib/deploy_channel_compose.sh` goes through `load_host_secret_contract`). Clean.
- `is_shared_key_domain` mirrors `is_optional` exactly, including raising `UndeclaredSecretConsumerError` on an unknown logical id. `consumers_declared_for` is correctly channel-scoped.

**Bootstrap check (`app/ops/host_secret_bootstrap.py:325-394`, wired at `:438`)**
- Ordering is right: the check runs *after* `_validate_secret` on the resolving consumer's own value, so the own side of every comparison is always well-formed 64-char hex.
- `HostSecretSharedDomainDivergenceError` subclasses `HostSecretBootstrapError`, so it survives the `except HostSecretBootstrapError: raise` re-raise at `:448` and at `:475` unwrapped, and `main` maps it to exit 78 with only the logical id. It is *not* a `HostSecretCredentialUnavailableError`, so the `run_on_credential_unavailable` opt-in at `:652` cannot launch a child into a split cipher domain. Correct fail-closed posture.
- Value-freedom holds: the error message carries the logical id only; the digest is never returned, logged, or persisted; `hmac.compare_digest` is used on the digests.
- Gating is on `shared_key_domain`, not on consumer count — `openai.api-key` has two declared consumers and correctly performs no sibling lookup (`test_non_shared_domain_secret_with_two_consumers_skips_sibling_lookup`).
- No request-time regression: `app/builderops/model_access_resolver.py` only `load_host_secret_contract`s; nothing on the API/preflight path now performs an extra Keychain read.

**Round-1 fix (a) — hex decode.** Correct and load-bearing. `_RAW_STORE_KEY_PATTERN` is `^[0-9a-fA-F]{64}$` (accepts both cases) and `app/heimdal/raw_store.py:129` does `bytes.fromhex(raw.strip())`, so two case spellings genuinely are the same AES key. Comparing raw text would have bricked both consumers on a correctly provisioned host. The `except ValueError -> utf-8` fallback cannot produce a false *agreement* in any reachable state, since the own side is always validated hex.

**Round-1 fix (b) — narrowed `except`.** Correct. Every in-repo lookup (`_security_keychain_lookup` non-zero exit / `OSError`, and `_security_framework_keychain_lookup` non-zero `OSStatus` / decode failure) raises `HostSecretBootstrapError`, so a genuinely absent sibling still skips; anything else reaches `:450` and fails this consumer closed. Also worth recording: divergence is still caught even when the first consumer skipped an unprovisioned sibling, because the second consumer's own bootstrap runs the same check in the reverse direction.

**Test evidence is not false-green.** I re-ran everything at this SHA — `tests/ops/test_host_secret_contract.py` + `tests/ops/test_host_secret_bootstrap.py` (78 passed), `tests/builderops` + `tests/deploy/test_deploy_channel_script.py` (1128 passed) = 1206, matching the PR body; `ruff check app tests` clean; `mypy` on both changed modules clean. I then reverted each round-1 fix in isolation and confirmed the two new regression tests actually bite: `test_shared_domain_agreement_is_hex_case_insensitive` and `test_sibling_lookup_programming_error_still_fails_closed` both FAIL against the pre-fix implementation and pass at head. Working tree restored, verified clean at `567f874`.

### Findings

**P2 — a skipped divergence check is indistinguishable from a passed one** (`app/ops/host_secret_bootstrap.py:389-392`)
The `except HostSecretBootstrapError: continue` skip is silent: nothing on stdout/stderr, no receipt, no marker in the materialized env file. An operator therefore cannot tell "both items read and agreed" from "the sibling was never read". This matters beyond the intended not-yet-provisioned case: if the sibling item exists but is unreadable to the calling process (a `security find-generic-password` ACL denial returns non-zero non-interactively, exactly like a missing item), the denial is typically symmetric, so *both* consumers skip each other and the split cipher domain this PR exists to catch starts silently on both sides. Suggested bounded follow-up: emit one value-free stderr line naming the logical id and the sibling consumer whose material could not be compared.
Severity rationale (recorded deliberately, given the protected-findings rule): this is a coverage gap in a newly added detector, not a credential/durability failure introduced by the change. Pre-PR behaviour was zero detection, the skip is required for the legitimate provision-one-consumer-first flow, and no test evidence is falsified by it. Acceptable to defer; must be logged.

**P3 — the sibling value is never grammar-validated before comparison** (`app/ops/host_secret_bootstrap.py:339-343`)
`bytes.fromhex` ignores interior ASCII whitespace, so a sibling item spelled `"aa aa ..."` decodes to the same bytes and reads as agreeing even though it fails `_RAW_STORE_KEY_PATTERN`. Harmless in practice (that sibling fails closed on its own bootstrap path, so no domain is split), but the docstring's "text bytes never collide with decoded key bytes here" is a slightly stronger claim than what is enforced. No change required.

**P3 — only one direction of the divergence check is covered**
Every new divergence test bootstraps `heimdal-capture-watch`. Nothing asserts that bootstrapping `heimdal-api-ingress` against a diverged `heimdal-capture-watch` item also refuses. It is the same code path, so this is coverage completeness only.

**P3 — nothing pins the class hierarchy that keeps divergence out of the `run_on_credential_unavailable` handoff**
The fail-closed behaviour at `:652` depends on `HostSecretSharedDomainDivergenceError` *not* being a `HostSecretCredentialUnavailableError`. That is currently true and correct, but no test asserts it, so a future reparenting could quietly make a split cipher domain "runnable". A one-line `assert not issubclass(...)` would pin it.

**P3 — module-scope `subprocess.run` patch in `test_shared_domain_check_never_logs_key_material`** (`tests/ops/test_host_secret_bootstrap.py:1291`)
`monkeypatch.setattr(host_secret_bootstrap.subprocess, "run", ...)` patches the real `subprocess` module process-wide for the test's duration. Restored by `monkeypatch` and sound as written; blast radius is wider than the assertion needs. Already logged in round 1; repeated here only for completeness.

No P0 or P1 findings. The round-1 fixes are correct, independently verified to be load-bearing, and introduce no new problem.

**VERDICT: clean**

PR #4716, opened 2026-08-09
Harden model inquiry fallback and role focus

RasmusTho:
## Change Lane
- [ ] Docs authoring lane
- [ ] Governance lane

Final-Review-Rounds: 1

Governing-Issue: #4713

## Linked Issue
Closes #4713

## SBS Impact
- Primary subsystem: Builder System
- Secondary subsystem(s): None
- Write class: BuilderOps runtime and repository documentation
- Persistence impact: Durable provider-attempt receipts and degraded terminal outcome
- Derived/rebuildable impact: Human-readable inquiry reports remain rebuildable from canonical trace records
- New or changed contract: Role-specific prompt lineage and bounded subscription-adapter fallback
- Owner-doc impact: Updated model inquiry README and adapter contract
- Transition debt impact: Legacy shared-prompt hashes remain readable and resumable
- Boundary risk: High; fallback is restricted to local subscription command adapters and provider API fallback remains forbidden

## Owner-Doc Writeback

## Summary
- Make subscription-backed model inquiry resilient to one unavailable lane with bounded, durable fallback and role-focused prompts.
- Persist every eligible failed attempt before trying the bounded alternate, then expose the effective target and a truthful `degraded_consensus` outcome.
- Keep refusal, credential, suspicious-output, unexpected, persistence, and provider-API paths fail-closed.

## Validation
- Focused model-inquiry, adapter, CLI, promotion, governance, and skill-entrypoint suite — 106 passed
- `ruff check app tests` — passed
- `mypy app` — passed for 876 source files
- `python scripts/lint_skills_consistency.py` — passed
- `DOCS_GUARD_ALLOW_TEMPORAL_SKIP=1 python scripts/docs_guard.py` — passed after confirming the capability owner docs are updated and no global temporal owner-doc truth changed
- Initial final review on `32aa2aa48` found two P1s; both were repaired in `eeaa0acb6` with session-expiry and recovered-distinct-target regression tests
- Independent convergence review of `eeaa0acb6` — approved with no remaining P0/P1 findings
- Single final review found six P1 issues across its repair/re-verification loop; all are fixed, and the reviewer approved logic SHA `48b0cb82a`
- Final CI exposed a stale pinned launcher digest; the one-line digest-only repair in `acd95af08` is self-verified by the 96-test host-installer and inquiry suite
- `scripts/review_before_ci_gate.py` — satisfied for `external-api`, `credential-durability`, and `state-machine`

## BuilderOps Routing
- Reason: Issue #4713 and this PR diff are the durable authority; no separate operational BuilderOps material was produced.

## Notes
Claim receipt: `github-comment:5233736410`. Degraded consensus is visible and is not promotion-ready.


app/builderops/model_inquiry_runner.py
```diff
@@ -26,11 +27,21 @@
...
     parse_model_turn_response,
 )
 from app.builderops.models import BuilderOpsValidationError
 
 ROLES = ("fable", "gpt_codex")
+FALLBACK_ADAPTER_FAILURE_CLASSES = frozenset(
+    {
+        "command_exit_nonzero",
+        "command_timeout",
+        "session_expired",
```
RasmusTho: [P1] Keep expired subscription sessions terminal

`session_expired` is emitted when the subscription CLI reports authentication-required/unauthorized/invalid-credentials, but including it in this fallback allowlist makes the runner invoke the alternate adapter. That violates Issue #4713's explicit constraint that credential failures must not retry or fall back (and the updated adapter contract says the same). I reproduced this with a local-command primary raising `AdapterExecutionError(..., failure_class="session_expired")`: the alternate executed all four turns and the run returned `degraded_consensus` instead of failing closed. Remove this auth failure class from the eligible fallback set and cover that the alternate is not called.

app/builderops/model_inquiry_runner.py
```diff
@@ -628,6 +749,36 @@ def _error_classification(error: Exception, outcome: str) -> str:
...
+    ]
+    effective_targets = {
+        (
+            item.turn.get("adapter_id"),
+            item.turn.get("provider"),
+            item.turn.get("model"),
+        )
+        for item in reviews
+    }
+    return bool(attempts) or len(effective_targets) != len(ROLES)
```
RasmusTho: [P1] Derive degradation from the accepting targets

`bool(attempts)` makes any earlier eligible failure force `degraded_consensus`, even when the two review turns that establish consensus use distinct effective targets. I reproduced a run where each primary failed only its draft, the alternate produced that draft, and both primaries recovered for review; the accepting turns were `fable/anthropic` and `gpt/openai`, yet the terminal outcome was degraded solely because two draft-attempt receipts existed. This contradicts #4713's distinction between genuine two-target `consensus` and same-target `degraded_consensus` and can incorrectly refuse promotion. Base this decision on the effective target identities relevant to consensus (and keep the service validator in sync), rather than the mere presence of any attempt receipt.

app/builderops/model_inquiry_runner.py
RasmusTho: Fixed in eeaa0acb6043f7bdea84574552cb32366782ad03.  is now terminal and the new  proves that the alternate subscription adapter is never called.

app/builderops/model_inquiry_runner.py
RasmusTho: Fixed in eeaa0acb6043f7bdea84574552cb32366782ad03. Consensus/degradation now derives only from the effective accepting review targets;  covers a recovered earlier attempt with a genuine two-target accepting pair.

app/builderops/model_inquiry_runner.py
RasmusTho: Correction: Fixed in eeaa0acb6043f7bdea84574552cb32366782ad03. session_expired is now terminal, and test_session_expiry_does_not_fallback proves that the alternate subscription adapter is never called.

app/builderops/model_inquiry_runner.py
RasmusTho: Correction: Fixed in eeaa0acb6043f7bdea84574552cb32366782ad03. Consensus/degradation now derives only from the effective accepting review targets; test_recovered_distinct_targets_remain_genuine_consensus covers a recovered earlier attempt with a genuine two-target accepting pair.

app/builderops/model_inquiry_runner.py
```diff
@@ -491,7 +587,31 @@ def _attempt_failure(
...
+        return details
+
+    def _candidate_adapters(
+        self,
+        adapters: Mapping[str, ModelTurnAdapter],
+        *,
+        role: str,
+    ) -> tuple[ModelTurnAdapter, ...]:
+        primary = adapters[role]
+        if self._adapters is None:
```
RasmusTho: **P1 — The sanctioned run path cannot enter this fallback chain.** The only non-test construction in the repository is `app/builderops/cli.py::_inquiry_run`, which calls `ModelInquiryRunner(service)` with `self._adapters is None`; this guard therefore always returns `(primary,)`. The documented `yggdrasil-model-inquiry` launcher is supposed to own that same BuilderOps command, so an unavailable subscription lane still terminates instead of trying the configured alternate. Please wire the sanctioned host-local subscription adapters into the production runner/CLI path (while retaining the LocalCommand-only guard), and add an end-to-end test through that invocation path.

RasmusTho (commented):
Independent current-SHA review found one P1 inline finding.

app/builderops/model_inquiry_runner.py
```diff
@@ -628,6 +748,29 @@ def _error_classification(error: Exception, outcome: str) -> str:
...
+def _fallback_was_used(reviews: Sequence[TurnExecution]) -> bool:
+    effective_targets = {
+        (
+            item.turn.get("adapter_id"),
+            item.turn.get("provider"),
+            item.turn.get("model"),
+        )
+        for item in reviews
+    }
+    return len(effective_targets) != len(ROLES)
```
RasmusTho, line 779: **P1 — Adapter labels can launder same-model agreement into promotable `consensus`.** The new effective-target key includes `adapter_id`, so two logical lanes configured with distinct IDs but the same provider/model produce `len == 2` here and are recorded as genuine consensus; the persistence validator mirrors the same tuple, and promotion admits it. That violates #4713's explicit out-of-scope constraint against treating same-model calls as independent-model acceptance evidence. Define/attest an actual target fingerprint (or at minimum require distinct provider/model targets) and use it both here and in terminal validation; add a regression where distinct adapter IDs wrap the same provider/model and must yield `degraded_consensus`.

RasmusTho (commented):
Second independent current-SHA P1 finding.

app/builderops/model_inquiry_runner.py
```diff
@@ -237,168 +260,220 @@ def _execute_turn(
...
+                trace,
+                role=role,
+                phase=phase,
+                round_index=round_index,
+                input_refs=input_refs,
+                reviewed_refs=reviewed_refs,
+                adapter_identity=sanitized_adapter_identity(adapter),
+                context_hash=context_hash,
+            )
+            prior_attempt = next(
```
RasmusTho, line 342: **P1 — Pre-change failed candidates are retried after the prompt-lineage migration.** This lookup only recognizes the newly role-prompted `adapter_request_id`. A durable attempt recorded before this PR has the legacy shared-prompt hash/ID; on resume `_request_packet` produces a different ID and the runner calls that same candidate again. The PR promises legacy shared-prompt hashes remain resumable and that resume skips already-failed candidates, so this breaks the no-retry/durable-ordering contract across deployment. Check both current and legacy request IDs for the same candidate (with the existing safe fallback eligibility rules), and add a resume regression containing a legacy failed attempt receipt.

RasmusTho (commented):
Third independent current-SHA P1 finding.

app/builderops/model_inquiry_runner.py
RasmusTho: Final reachable fix: 48b0cb82aa863dc0d806af9d8c824cbc0033cca9. Expired sessions remain terminal and the alternate is not called.

app/builderops/model_inquiry_runner.py
RasmusTho: Final reachable fix: 48b0cb82aa863dc0d806af9d8c824cbc0033cca9. Consensus now derives from distinct accepting provider/model targets, including recovery after an earlier receipted attempt.

app/builderops/model_inquiry_runner.py
RasmusTho: Fixed in final reachable SHA 48b0cb82aa863dc0d806af9d8c824cbc0033cca9. The canonical CLI now constructs the fixed LocalCommand subscription pair when the owner-controlled operational mode is enabled, with an end-to-end CLI fallback regression.

app/builderops/model_inquiry_runner.py
RasmusTho: Fixed in final reachable SHA 48b0cb82aa863dc0d806af9d8c824cbc0033cca9. Distinct adapter IDs with the same provider/model remain degraded and non-promotable.

app/builderops/model_inquiry_runner.py
RasmusTho: Fixed in final reachable SHA 48b0cb82aa863dc0d806af9d8c824cbc0033cca9. Resume recognizes current and legacy request IDs for already-failed candidates and does not replay the legacy attempt.

Issue #4582, opened 2026-08-02
[Multi-Vault Runtime] cutover-fence-floor-and-closure: fence every DB producer, set the runtime floor, close MVR-05A (MVR-05A8)


RasmusTho:
## Context
Child of #2143, seventh and final bounded slice of MVR-05A under
`docs/MULTI_VAULT_RUNTIME/ROUTE_REQUESTS_THROUGH_ACTIVE_CONTEXT.md :: 05A child decomposition`. Every
durable projection and the outbox are binding-keyed by the time this starts, and the per-binding
effect lease exists. This slice installs the cutover machinery around them — the all-process fence,
the minimum-runtime floor, the compatibility ingress translator, the scalar worker's poll/ack gate,
and the sealed GOV-revocation seam — and writes the stage's owner-doc contracts. It is the slice that
lets the 05A hub post its stage receipt. 05A hub: #3859.

Parent: #2143

## Scope
**MVR-05A8 — mixed-version fence, runtime floor, worker gate, and MVR-05A closure.**

- Generate the fence inventory from production compose and runtime producer truth instead of a
  hard-coded list. The "every enabled process that can reach the DB or outbox seam" population is
  currently four uncoordinated copies with no cross-check between them:
  `scripts/instance_state_writer_inventory.py:33` (`COMPOSE_WRITER_SERVICES`),
  `app/instance/instance_state.py:26` (`_REQUIRED_CONSUMERS`),
  `scripts/lib/instance_state_deployment.sh:194` (the literal `stop api worker watcher
  heimdal-capture-watch`, which *is* the entire drain plan today), and
  `app/instance/principal_fence.py:66` (`COMPOSE_AUTH_SERVICES`, a superset and the only one with an
  anti-drift test against the real compose file). A fifth vocabulary classifies *native* writers by
  argv, so extending only the compose set leaves native writers unfenced.
- Expect a real finding rather than a green gate. The `migrate` service reaches the main Postgres
  (`env_file: ./config/runtime.defaults.env`, `command: run_migrations.sh`) but is absent from
  `COMPOSE_WRITER_SERVICES`; it is listed in `COMPOSE_NON_AUTH_SERVICES`, which is *correct* for the
  principal fence because it mints no instance principal. The two sets answer different questions, so
  the repair is to the DB/outbox writer inventory, not to the auth classification. `migrate` also
  needs explicit handling as the process that *runs* the binding-keyed migration: it cannot be a
[...]

Issue #4581, opened 2026-08-02
[Multi-Vault Runtime] outbox-dual-key-dedup: bind the outbox and dedup across the cutover without rotating keys (MVR-05A7)


RasmusTho:
## Context
Child of #2143, sixth bounded slice of MVR-05A under
`docs/MULTI_VAULT_RUNTIME/ROUTE_REQUESTS_THROUGH_ACTIVE_CONTEXT.md :: 05A child decomposition`. The
stage's original outbox criterion asked for legacy idempotency keys to be "scoped/coalesced" into
upgraded keys before the first binding-keyed producer starts. That is not implementable, and the
reason is structural rather than a matter of effort — see `Constraints`. The specification has been
amended to the dual-key design this slice delivers, which preserves the same invariant by a mechanism
a migration can actually execute. 05A hub: #3859.

Parent: #2143

## Scope
**MVR-05A7 — outbox binding column and dual-key compatibility dedup.**

- Add `vault_binding_id` and a nullable `legacy_key` column to `outbox`. It has carried exactly six
  columns — `id, topic, payload, created_at, delivered_at, attempts` — since revision `f3a1c9d2e4b7`,
  and no revision has altered it since.
- Backfill `legacy_key = id` for every existing row, pending and delivered. This is a copy, not a
  recomputation, and it is what makes the cutover possible at all.
- Classify `vault_binding_id` for existing rows from provable evidence; rows whose binding is not
  provable quarantine.
- Binding-keyed producers derive **both** keys — the unchanged legacy key and the binding-scoped one —
  insert under the binding-scoped `id`, and suppress the emission when a row already exists under
  either key.
- Do not rotate `OUTBOX_IDEMPOTENCY_NAMESPACE`. Its never-rotate promise is documented in
  `app/services/outbox.py` and stays literally intact under this design.
- Route the change through the single `write_outbox_event` choke point rather than the individual
  producer call sites. Five of the worker's re-emissions (retry, dead-letter, and the knowledge
  signal) must **inherit the source row's binding** rather than resolve ambient context, or a
  dead-letter of an A-bound row lands under B.

## Source Anchors
- `docs/MULTI_VAULT_RUNTIME/ROUTE_REQUESTS_THROUGH_ACTIVE_CONTEXT.md :: 05A child decomposition`
[...]

PR #4928, opened 2026-08-15
Add cross-process per-binding effect lease

### LINKED ISSUE #4580: [Multi-Vault Runtime] per-binding-effect-lease: build the cross-process shared/exclusive effect lease (MVR-05A6)
## Context
Child of #2143, fifth bounded slice of MVR-05A under
`docs/MULTI_VAULT_RUNTIME/ROUTE_REQUESTS_THROUGH_ACTIVE_CONTEXT.md :: 05A child decomposition`. The
per-binding shared/exclusive effect lease that the stage's worker-gate and revocation-seam criteria
depend on **does not exist in any form**: there are zero `effect_lease` hits under `app/**`. This
slice builds that concurrency primitive from scratch. It touches no durable projection, so it carries
none of the migration risk of the three table groups and can be proven on its own — which is why it
sits after them rather than before. 05A hub: #3859.

Parent: #2143

## Scope
**MVR-05A6 — cross-process per-binding shared/exclusive effect lease.**

- A cross-process lease keyed by binding, with two modes. A *shared* holder represents one in-flight
  authorized effect (a dispatch, acknowledgement, and receipt window). An *exclusive* holder
  represents a change to the binding's authorization, location, or existence. The behavioural
  contract is that an exclusive acquirer cannot complete while any shared holder is inside its effect
  window, and a shared acquirer cannot begin once an exclusive acquirer is pending.
- Preserve the lock ordering the lane already fixes: the MVR-01 host-global ownership fence is taken
  first, then the per-binding lease.
- Reuse the durability idiom of `app/instance/ownership_ledger.py` — its ledger-then-registry lock
  ordering, its storage-mutation capability gate, its atomic private JSON write, and its separate
  crash-recovery journal — without assuming its mechanism. That module is exclusive-only across all
  1,401 of its lines. Two shared/exclusive `flock` call sites already exist in this repository
  (`app/settings/compiler.py:123` and `app/agent_memory/provisional_write.py:244`) and must be
  evaluated as prior art before persisting shared-holder state under an exclusive lock. Choose with a
  stated reason; the contract is the behaviour above, not the mechanism.
- Give the lease its own validated producer on the registry store, following the MVR-02/MVR-04
  `set_dimension_state` precedent, rather than adding a fourth slot to a general extension-state
  setter.
- Crash safety: a holder that dies without releasing must not deadlock the binding, and recovery must
  not silently promote a stale shared holder into a completed effect.

## Source Anchors
- `docs/MULTI_VAULT_RUNTIME/ROUTE_REQUESTS_THROUGH_ACTIVE_CONTEXT.md :: 05A child decomposition`
- `docs/MULTI_VAULT_RUNTIME/README.md :: Cross-Task Invariants / Interaction Safety`
- `app/instance/ownership_ledger.py`
- `app/settings/compiler.py`
- `app/instance/principal_fence.py`

## SBS Impact
- Primary subsystem: WSP
- Secondary subsystem(s): GOV, OEF
- Write class: mechanical instance-local state; no human-artifact write authority
- Authority impact: none granted; the lease serialises effects against authority changes, it does not
  decide authority
- Persistence impact: durable instance-local lease state plus a crash-recovery journal; no content
  and no vault data
- Derived/rebuildable impact: lease state is instance-local mechanical state, rebuildable from a
  quiesced instance
- New or changed contract: a per-binding shared/exclusive effect lease, acquired after the host-global
  ownership fence
- Owner-doc impact: follow-up-issue — the spec assigns `docs/ARCHITECTURE.md` to MVR-05B and the
  deployment/ops description of the lease to MVR-05A8, so this slice adds no owner-doc writeback
- Transition debt impact: no effect
- Boundary risk: lock-order inversion between the host-global ownership fence and the per-binding
  lease would deadlock every enabled process, so ordering must be enforced by test rather than by
  convention

## Constraints
- Dependency: #4579 (MVR-05A5). Keep this Issue `agent:blocked` until that dependency is merged and
  reconciled on `origin/main`. Execution through MVR-06D is serial; no parallel dispatch.
- The lease is a concurrency primitive only. It must not read, write, or classify any durable
  projection or outbox row; those belong to MVR-05A3, MVR-05A4, MVR-05A5 and MVR-05A7.
- Lock ordering is host-global ownership fence first, per-binding lease second, enforced by a test
  that fails on inversion.
- Shared-ness must be observable and countable: the exclusive path has to be able to tell that a
  shared holder is still inside its effect window. A plain advisory shared lock that cannot answer
  that question does not satisfy the contract.
- If the lease module lives under `app/instance/**` and opens a database connection, it needs an
  entry in `tests/guard/test_no_direct_db_imports.py::ALLOW_FILES`; a new top-level `app/` package
  also needs registering in `importlinter.ini`.
- Old-image detection, where needed, must be file-existence based, following the
  `deployment-*-restart-fence.json` pattern the compose commands already grep for, not state an old
  image would have to parse.
- Keep #2143 as a blocked validation hub; this child never claims or closes the parent or #3859.

## Acceptance Criteria
- [ ] **MVR-05A6:** An exclusive acquirer for a binding cannot complete while any shared holder for
  that binding is inside its effect window; it either blocks or waits for the already-authorized
  effect to finish.
  - Verify: `tests/instance/test_binding_effect_lease.py::test_exclusive_acquirer_waits_for_an_in_flight_shared_holder`
- [ ] **MVR-05A6:** A shared acquirer cannot begin a new effect window once an exclusive acquirer is
  pending for the same binding, so exclusive work cannot be starved by a stream of shared holders.
  - Verify: `tests/instance/test_binding_effect_lease.py::test_pending_exclusive_blocks_new_shared_acquisition`
- [ ] **MVR-05A6:** Shared and exclusive holders for **different** bindings proceed concurrently
  without serialising against each other.
  - Verify: `tests/instance/test_binding_effect_lease.py::test_leases_for_distinct_bindings_do_not_serialise`
- [ ] **MVR-05A6:** The lease is cross-process, not merely cross-thread: two separate operating-system
  processes observe the same exclusion.
  - Verify: `tests/instance/test_binding_effect_lease.py::test_exclusion_holds_across_separate_processes`
- [ ] **MVR-05A6:** A holder that dies without releasing does not deadlock the binding, and recovery
  never reports a stale shared holder's effect as completed.
  - Verify: `tests/instance/test_binding_effect_lease_recovery.py::test_crashed_holder_recovers_without_deadlock_or_false_completion`
- [ ] **MVR-05A6:** Every production acquisition takes the host-global ownership fence before the
  per-binding lease, and the guard fails on an inverted order at a production call site.
  - Verify: `tests/architecture/test_multi_vault_lease_lock_order.py::test_production_callers_take_the_ownership_fence_before_the_binding_lease`
- [ ] **MVR-05A6:** The lease has its own validated registry-store producer rather than an additional
  slot on a general extension-state setter, and an invalid lease state is rejected at the producer.
  - Verify: `tests/instance/test_binding_effect_lease_store.py::test_lease_state_has_its_own_validated_producer`

## Out of Scope
- Wiring the lease into the scalar worker's dispatch, acknowledgement, and receipt path, and into the
  GOV-revocation seam. Those are MVR-05A8.
- Foreground read and write producers, which the lane assigns to MVR-05B and MVR-05C.
- Removal, relocation, and cross-channel transfer, which remain dormant until MVR-06B.
- Every durable projection and outbox change.

## Suggested Validation
- `pytest -q tests/instance/test_binding_effect_lease.py tests/instance/test_binding_effect_lease_recovery.py`
- `pytest -q tests/instance/test_binding_effect_lease_store.py tests/architecture/test_multi_vault_lease_lock_order.py`
- `pytest -q tests/guard/test_no_direct_db_imports.py`
- `pytest -q -m "not pg"`
- `mypy app`
- `ruff check app tests`
- `python3 scripts/validate_source_anchors.py` against this Issue body.

## Source Docs
- `docs/MULTI_VAULT_RUNTIME/ROUTE_REQUESTS_THROUGH_ACTIVE_CONTEXT.md`
- `docs/MULTI_VAULT_RUNTIME/README.md`
- `docs/MULTI_VAULT_RUNTIME/PARENT_FEATURE_ISSUE.md`

## Applies learning (optional)
Applies the MVR-01 ownership-ledger and MVR-03 principal-fence precedent for cross-process instance
state, and the repeated finding that `app/dispatcher/leases.py` is an unrelated builder-side
subsystem that is easy to mistake for existing coverage of this primitive.

RasmusTho:
Final-Review-Rounds: 1

Governing-Issue: #4580

## Linked Issue

Fixes #4580

## SBS Impact

- Primary subsystem: WSP
- Secondary subsystem(s): GOV, OEF
- Write class: mechanical instance-local state; no human-artifact write authority
- Persistence impact: durable instance-local per-binding lease state plus crash journal
- Derived/rebuildable impact: lease state is instance-local and rebuildable from a quiesced instance
- New or changed contract: cross-process per-binding shared/exclusive effect lease acquired after the host-global ownership fence
- Owner-doc impact: follow-up issue #4582 owns deployment/operations wiring; MVR-05B owns the architecture writeback
- Transition debt impact: none
- Boundary risk: lock-order inversion, stale process authority, or ambiguous registry recovery could deadlock an enabled binding; focused guards and fail-closed recovery cover these paths

## Owner-Doc Writeback

- [x] Owner-doc follow-up issue created and linked: #4582 (with MVR-05B retaining its assigned architecture writeback).

## Summary

- add a durable, countable cross-process shared/exclusive lease keyed by vault binding, with FIFO exclusive fairness and distinct-binding concurrency
- make per-holder activity locks authoritative across PID namespaces, enforce host-global ownership-fence-first ordering, and converge lease state with the validated registry producer through previous/next-only crash journals
- rebuild lease-local state from restored registry authority under the existing quiescence fence and reject linked private state paths without mutating their targets
- add deterministic process, crash, cancellation, fork, descriptor, fairness, restore, lock-order, registry-producer, private-path, and WriteGuard proofs without wiring the MVR-05A8 worker/revocation consumers

## Validation

- focused lease/import/WriteGuard/restore selector: 58 passed
- registry/active-context regression selector: 13 passed
- `pytest -q tests/guard/test_no_direct_db_imports.py`: passed
- `ruff check app tests`: passed
- `mypy app`: passed (893 source files)
- `lint-imports --config importlinter.ini`: passed (5 contracts kept, 0 broken)
- source-anchor validator: passed against issue #4580
- host-leased serial non-PG suite: 13,529 passed, 49 skipped, 405 deselected, 18 xfailed
- Sol/xhigh mechanism convergence at `b193b17240c33761366858615c856b60e7d68dfd`: P0/P1/P2 all zero; one historical severity-routing observation remains P3 `record_only`
- base integration: rebased onto `06fdf08012672b087af059818ea6398dcba8e319`; zero incoming path overlap, stable patch ID and all 12 delivery blobs/full binary patch preserved byte-identically

## BuilderOps Routing

- Reason: protected mechanism repair iterations were resolved and independently re-reviewed within this issue delivery; no reusable Builder-System learning or authority-class promotion was identified.

## Notes

- Residual platform risk is fail-closed: acquisition is refused without `O_NOFOLLOW` or trustworthy local process-incarnation metadata; cross-namespace liveness authority comes from per-holder locks, not observer-local PIDs.
- #4581 remains blocked; this PR does not claim or wire MVR-05A7/MVR-05A8 work.

### DIFF
--- a/app/instance/binding_effect_lease.py
+++ b/app/instance/binding_effect_lease.py
@@ -3,12 +3,14 @@
 from __future__ import annotations
 
 import copy
+import ctypes
 import fcntl
 import hashlib
 import json
 import os
 import stat
 import subprocess
+import sys
 import tempfile
 import time
 from contextlib import contextmanager
@@ -24,6 +26,55 @@ from app.instance.vault_registry import VaultRegistryStore
 
 LEASE_SCHEMA = "agentic-pkm.binding-effect-lease.v1"
 JOURNAL_SCHEMA = "agentic-pkm.binding-effect-lease-journal.v1"
+
+
+class _DarwinProcBsdInfo(ctypes.Structure):
+    """Native ``proc_bsdinfo`` layout from Darwin ``sys/proc_info.h``."""
+
+    _fields_ = [
+        ("pbi_flags", ctypes.c_uint32),
+        ("pbi_status", ctypes.c_uint32),
+        ("pbi_xstatus", ctypes.c_uint32),
+        ("pbi_pid", ctypes.c_uint32),
+        ("pbi_ppid", ctypes.c_uint32),
+        ("pbi_uid", ctypes.c_uint32),
+        ("pbi_gid", ctypes.c_uint32),
+        ("pbi_ruid", ctypes.c_uint32),
+        ("pbi_rgid", ctypes.c_uint32),
+        ("pbi_svuid", ctypes.c_uint32),
+        ("pbi_svgid", ctypes.c_uint32),
+        ("rfu_1", ctypes.c_uint32),
+        ("pbi_comm", ctypes.c_char * 16),
+        ("pbi_name", ctypes.c_char * 32),
+        ("pbi_nfiles", ctypes.c_uint32),
+        ("pbi_pgid", ctypes.c_uint32),
+        ("pbi_pjobc", ctypes.c_uint32),
+        ("e_tdev", ctypes.c_uint32),
+        ("e_tpgid", ctypes.c_uint32),
+        ("pbi_nice", ctypes.c_int32),
+        ("pbi_start_tvsec", ctypes.c_uint64),
+        ("pbi_start_tvusec", ctypes.c_uint64),
+    ]
+
+
+class _DarwinProcBsdShortInfo(ctypes.Structure):
+    """Native ``proc_bsdshortinfo`` used to identify an unreaped zombie."""
+
+    _fields_ = [
+        ("pbsi_pid", ctypes.c_uint32),
+        ("pbsi_ppid", ctypes.c_uint32),
+        ("pbsi_pgid", ctypes.c_uint32),
+        ("pbsi_status", ctypes.c_uint32),
+        ("pbsi_comm", ctypes.c_char * 16),
+        ("pbsi_flags", ctypes.c_uint32),
+        ("pbsi_uid", ctypes.c_uint32),
+        ("pbsi_gid", ctypes.c_uint32),
+        ("pbsi_ruid", ctypes.c_uint32),
+        ("pbsi_rgid", ctypes.c_uint32),
+        ("pbsi_svuid", ctypes.c_uint32),
+        ("pbsi_svgid", ctypes.c_uint32),
+        ("pbsi_rfu", ctypes.c_uint32),
+    ]
 
 
 class BindingEffectLeaseError(RuntimeError):
@@ -484,20 +535,86 @@ class BindingEffectLeaseManager:
             remainder = raw[raw.rindex(")") + 2 :].split()
             return f"proc:{remainder[19]}", remainder[0]
         except (OSError, IndexError, ValueError):
-            try:
-                result = subprocess.run(
-                    ["ps", "-o", "stat=", "-o", "lstart=", "-p", str(pid)],
-                    check=True,
-                    capture_output=True,
-                    text=True,
-                    timeout=1,
-                )
-                output = result.stdout.strip().split(maxsplit=1)
-                if len(output) == 2:
-                    return f"ps:{output[1]}", output[0][:1]
-            except (OSError, subprocess.SubprocessError):
-                pass
+            if sys.platform == "darwin":
+                return BindingEffectLeaseManager._darwin_process_identity(pid)
             raise BindingEffectLeaseError("a trustworthy process-incarnation token is unavailable")
+
+    @staticmethod
+    def _darwin_process_identity(pid: int) -> tuple[str, str]:
+        """Return a microsecond-resolution native process incarnation on Darwin."""
+
+        try:
+            library = ctypes.CDLL("/usr/lib/libproc.dylib", use_errno=True)
+            proc_pidinfo = library.proc_pidinfo
+        except (AttributeError, OSError) as exc:
+            raise BindingEffectLeaseError(
+                "a trustworthy process-incarnation token is unavailable"
+            ) from exc
+        proc_pidinfo.argtypes = [
+            ctypes.c_int,
+            ctypes.c_int,
+            ctypes.c_uint64,
+            ctypes.c_void_p,
+            ctypes.c_int,
+        ]
+        proc_pidinfo.restype = ctypes.c_int
+        info = _DarwinProcBsdInfo()
+        result = proc_pidinfo(
+            pid,
+            3,  # PROC_PIDTBSDINFO
+            0,
+            ctypes.byref(info),
+            ctypes.sizeof(info),
+        )
+        if (
+            result == ctypes.sizeof(info)
+            and info.pbi_pid == pid
+            and info.pbi_status in {1, 2, 3, 4, 5}
+            and info.pbi_start_tvsec > 0
+            and info.pbi_start_tvusec < 1_000_000
+        ):
+            state = "Z" if info.pbi_status == 5 else "?"
+            return (
+                f"darwin:{info.pbi_start_tvsec}:{info.pbi_start_tvusec}",
+                state,
+            )
+
+        # Full BSD info may disappear after process exit while an unreaped PID
+        # still answers `kill(0)`. The native short flavor retains `p_stat`, so
+        # it can prove a zombie is dead without supplying a weak live token.
+        short = _DarwinProcBsdShortInfo()
+        short_result = proc_pidinfo(
+            pid,
+            13,  # PROC_PIDT_SHORTBSDINFO
+            0,
+            ctypes.byref(short),
+            ctypes.sizeof(short),
+        )
+        if (
+            short_result == ctypes.sizeof(short)
+            and short.pbsi_pid == pid
+            and short.pbsi_status == 5
+        ):
+            return f"darwin:zombie:{pid}", "Z"
+        if BindingEffectLeaseManager._darwin_process_is_zombie(pid):
+            return f"darwin:zombie:{pid}", "Z"
+        raise BindingEffectLeaseError("a trustworthy process-incarnation token is unavailable")
+
+    @staticmethod
+    def _darwin_process_is_zombie(pid: int) -> bool:
+        """Use ``ps`` only as terminal-state proof, never as an identity token."""
+
+        try:
+            result = subprocess.run(
+                ["ps", "-o", "stat=", "-p", str(pid)],
+                check=True,
+                capture_output=True,
+                text=True,
+                timeout=1,
+            )
+        except (OSError, subprocess.SubprocessError):
+            return False
+        return result.stdout.strip().startswith("Z")
 
     @staticmethod
     def _empty_state(vault_binding_id: str) -> _LeaseState:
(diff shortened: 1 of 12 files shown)


RasmusTho (commented):
Independent high-risk review verdict: **BLOCK** at exact head `4ad2f33947809d832bf214e97a4c9dae4e9c4d57`.

Live `main` was `06fdf08012672b087af059818ea6398dcba8e319` at final base-drift readback. A synthetic merge tree was clean and the intervening files do not overlap this mechanism. The 17 issue-focused tests pass, but the inline adversarial reproductions expose three additional P1 failures.

Carry-forward findings on this same head also remain valid:
- [P1 registry/binding revision churn](https://github.com/RasmusTho/agentic-pkm-mvp/pull/4928#discussion_r3789630472) — locally reproduced as registry/binding revision `1 -> 2 -> 3` across one shared acquire/release.
- [P2 opaque binding-ID normalization](https://github.com/RasmusTho/agentic-pkm-mvp/pull/4928#discussion_r3789630475) — locally reproduced with a currently valid whitespace-bearing registration.

Convergence trigger: there are now multiple blocking findings in the same stateful lease mechanism. Stop point-fixing and run the repository's mechanism-level convergence gate over namespace-independent liveness, durable-state/registry revision separation, all exception and crash edges, and backup/restore. After repair, run the focused tests plus the new adversarial cases and obtain one fresh clean independent review on the repaired exact SHA.

app/instance/binding_effect_lease.py
```diff
@@ -0,0 +1,653 @@
...
+    def _holder(self, holder_id: str, *, ticket: int, mode: str) -> _HolderState:
+        return {
+            "holderId": holder_id,
+            "pid": os.getpid(),
+            "processStart": self._process_start(os.getpid()),
+            "ticket": ticket,
+            "mode": mode,
+        }
+
+    def _holder_alive(self, holder: Mapping[str, object]) -> bool:
```
RasmusTho: [P1] Make holder liveness valid across production PID namespaces

The persisted `pid` and `/proc/<pid>/stat` token are interpreted in the *observer's* PID namespace. The production API, worker, watcher, and Heimdal services share `instance-state` but do not share a PID namespace, so a sibling service cannot authoritatively evaluate another service's PID. `_load_reconciled_locked()` consequently removes a live foreign shared holder or exclusive waiter. I reproduced the resulting contract break: after an exclusive was pending, a foreign-identity reconciler removed both live records and entered a new shared window; the exclusive timed out, and the original shared holder's release raised `held binding effect lease state is missing`. This violates both countability and the no-barging AC. Use namespace-independent liveness evidence (for example, lock-backed holder/waiter tokens) and prove the case with two isolated PID namespaces/services, not only `multiprocessing` in one namespace.

app/instance/binding_effect_lease.py
```diff
@@ -0,0 +1,653 @@
...
+                                    updated["exclusiveHolder"] = holder
+                                self._commit_locked(vault_binding_id, current, updated)
+                                return _HeldLease(vault_binding_id, holder_id, mode, descriptor)
+                if deadline is not None and time.monotonic() >= deadline:
+                    raise BindingEffectLeaseTimeout(
+                        f"timed out waiting for {mode} binding effect lease"
+                    )
+                time.sleep(self.poll_interval)
+        except BaseException:
+            if pending is not None:
```
RasmusTho: [P1] Release the gate even when pending cleanup also fails

`_discard_pending()` runs before the `LOCK_UN`/`close` block. If an exclusive acquisition has already obtained `LOCK_EX` and its durable promotion fails, a correlated failure while discarding the pending row escapes here and skips descriptor cleanup. The raw integer FD then remains open and keeps the binding's exclusive gate locked until the long-lived process exits. I reproduced this by failing the promotion and cleanup commits: even after persisted pending state was repaired, a new shared acquisition still timed out on the leaked gate. Put unlock/close in an outer `finally` that cannot be bypassed by cleanup, preserve both errors, and add the dual-failure recovery test.

app/instance/binding_effect_lease.py
```diff
@@ -0,0 +1,653 @@
...
+        path = self._state_path(vault_binding_id)
+        if path.exists():
+            self._assert_private_file(path)
+            try:
+                value = json.loads(path.read_text(encoding="utf-8"))
+            except (OSError, json.JSONDecodeError) as exc:
+                raise BindingEffectLeaseError("binding effect lease state is corrupt") from exc
+            local_state = self._validate_state(vault_binding_id, value)
+            registry_state = self._registry_state(vault_binding_id)
+            if local_state != registry_state:
```
RasmusTho: [P1] Reconcile lease files with governed instance-state restore

A journal-less local/registry mismatch is permanently fatal here, but `InstanceStateBackup` on current `main` captures/restores only the registry and ownership artifacts; it neither includes nor clears this new state directory and its journals. A normal quiesced rollback therefore restores registry generation N while leaving later local generation M in place, and the first observe/acquire fails with `binding effect lease state diverges from registry without a journal`. I reproduced backup-at-revision-1, one acquire/release (local generation 2), registry restore, then this failure. Either include the lease generation atomically in backup/restore or deterministically clear/rebuild the derived files from the restored registry under the quiescence fence, with an advance-then-restore round-trip test.

